PR #45, opened 2025-06-13
Release OpenHands Cloud as source-available

rbren:


charts/openhands/templates/user-waitlist-configmap.yaml
```diff
@@ -0,0 +1,11 @@
+{{- if .Values.allowedUsers }}
```
malhotra5, line 1: Can we nuke this? Doesn't seem to be in use

rbren: Yup good idea, but I want to minimize drift for now. Currently all these files are duplicated between this and our internal deploy repo

Once this gets merged, my hope is to 💣 the deploy version and start using this

xingyaoww (approved):
LGTM!

charts/openhands/README.md
```diff
@@ -0,0 +1,378 @@
...
+   cd openhands-cloud/charts/openhands
+   ```
+
+2. Create the openhands namespace:
+   ```bash
+   # Create namespace (if you want to use a different namespace, you will need to change it in all the commands that follow)
+   kubectl create namespace openhands
+   ```
+
+3. Create a secret for your LLM
```
rbren, line 38:
```suggestion
3. Create a secret for your LLM. We'll assume Anthropic here, but you can set any env vars you'll need to connect to your LLM, including e.g. OpenAPI keys, or AWS keys for Bedrock models. You can use any env var names you want--we'll reference them again below in our LiteLLM setup
```

rbren:
Yeah this has been tested probably a dozen times, but it's changed every time, either with fixes, or convenience features 🙃 

I got it installed in a KIND cluster up to the "DNS" step today, so it's working _reasonably_ well.

Main TODOs are:
* Auto-create LiteLLM team (in progress)
* Enable email-based auth
* Maybe auto-create secrets?

PR #55, opened 2025-06-13
update some bits from upstream chart

rbren:


rbren:
@openhands fix this
```
Error:  templates/: template: openhands/templates/proactive-convo-clean-cronjob.yaml:1:14: executing "openhands/templates/proactive-convo-clean-cronjob.yaml" at <.Values.proactiveConvoClean.enabled>: nil pointer evaluating interface {}.enabled

Error: Process completed with exit code 1.
```

PR #125, opened 2025-07-31
Add securityContext and runAsUser options for all pods in Helm charts

rbren:
## Description
This PR adds securityContext and runAsUser options for all pods in the Helm charts to enhance security by ensuring pods run with a non-root user.

## Changes
- Added securityContext settings with runAsUser: 1000 and runAsNonRoot: true to values.yaml files in all three charts:
  - openhands
  - image-loader
  - runtime-api
- Updated all pod specifications in templates to include the securityContext settings:
  - Deployments
  - Jobs
  - CronJobs
  - DaemonSets

## Security Impact
This change improves the security posture of the application by:
- Ensuring pods run with a non-root user (UID 1000)
- Explicitly preventing pods from running as root with runAsNonRoot: true

## Testing
These changes should be tested in a non-production environment to ensure all pods start correctly with the new security context settings.

### DIFF
--- a/charts/image-loader/templates/daemonset.yaml
+++ b/charts/image-loader/templates/daemonset.yaml
@@ -17,6 +17,10 @@ spec:
       labels:
         {{- include "image-loader.selectorLabels" . | nindent 8 }}
     spec:
+      {{- with .Values.securityContext }}
+      securityContext:
+        {{- toYaml . | nindent 8 }}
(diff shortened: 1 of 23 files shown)


chuckbutkus:
Do we want to default to non-root?  Or should we default to root and allow override?

chuckbutkus (approved):

PR #129, opened 2025-08-05
Update values file and config job template

chuckbutkus:


chuckbutkus:
> Could you also update the version of the chart in the Chart.yaml

Done

sliceofpepperoni (approved):

PR #175, opened 2025-08-21
fix env for root vs openhands

rbren:
## Helm Chart Checklist

### DIFF
--- a/charts/openhands/values.yaml
+++ b/charts/openhands/values.yaml
@@ -231,6 +231,9 @@ runtime:
   image:
     repository: ghcr.io/all-hands-ai/runtime
     tag: 7a168b9b5f33e63746456779032c0caa623e23f4-nikolaik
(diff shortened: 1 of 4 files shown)


charts/openhands/values.yaml
```diff
@@ -231,6 +231,9 @@ runtime:
   image:
     repository: ghcr.io/all-hands-ai/runtime
     tag: 7a168b9b5f33e63746456779032c0caa623e23f4-nikolaik
+  # If true, sets SANDBOX_USER_ID=0 and RUN_AS_OPENHANDS=false
+  # If false, sets SANDBOX_USER_ID=1000 and RUN_AS_OPENHANDS=true
+  runAsRoot: false
```
rbren, line 236:
```suggestion
  runAsRoot: true
```

chuckbutkus (approved):

PR #180, opened 2025-08-22
feature - Add optional linear and jira env

raymyers:
## Description

Add environment variables supporting Jira and Linear integrations (enterprise features), matching the chart portion of this internal PR:

https://github.com/All-Hands-AI/deploy/pull/1409/files#diff-2c0894607a3bc76e85c3cc04cf2b1d0cc6b19e08e37d0557d20b516360b81ad5

## Helm Chart Checklist

- [X] I have updated the `version` field in `Chart.yaml` for each modified chart
- [X] I have verified backwards compatibility with existing values.yaml configurations
- [X] I have updated the chart's README.md if there are any breaking changes or new required values

### DIFF
--- a/charts/openhands/values.yaml
+++ b/charts/openhands/values.yaml
@@ -129,27 +129,23 @@ ingress:
   # host: "app.example.com"
   class: traefik
   # Legacy annotations field - kept for backward compatibility
-  annotations:
-    {}
+  annotations: {}
     # UPDATE: if you use cert-manager, enter your clusterIssuer may not match.
     # cert-manager.io/cluster-issuer: letsencrypt-production
-  
+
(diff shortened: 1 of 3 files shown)


charts/openhands/values.yaml
```diff
@@ -587,6 +582,16 @@ runtime-api:
     CPU_REQUEST: "500m"
     EPHEMERAL_STORAGE_SIZE: "10Gi"
 
+jira:
+  enabled: false
+
+jiraDc:
+  enabled: false
+  enableOAuth: false
```
malhotra5, line 590:
```suggestion
```

should always be enabled, i think this was done to work on the integration code locally without auth

there are probably better ways of handling local development imho

PR #460, opened 2026-03-19
PLTF-309: updated replicated with openhands fix for max budget none

aivong-openhands:
## Description

There is an updated fix for max budget none which is an improvement for the fix for the out of funds error https://github.com/OpenHands/OpenHands/pull/13482

## Helm Chart Checklist

## Additional Notes

Can continue to have a conversation without the out of funds error:
Screenshot 2026-03-19 at 12 45 00 PM

Can pull/push/create a PR on a github repo:
Screenshot 2026-03-19 at 12 48 56 PM
Screenshot 2026-03-19 at 12 49 02 PM


aivong-openhands:
If we get @hieptl 's fix in, then this PR is no longer needed. I simply wanted to verify the replicated openhands setup with the updated out of funds error fix

PR #478, opened 2026-03-24
PLTF-355: Add HTTP proxy configuration for Replicated deployments

jlav:
## Description

A customer requires all outbound traffic to route through a corporate HTTP proxy that performs SSL interception with a self-signed certificate. This PR adds proxy configuration to the KOTS admin console so customers can configure `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY`, and `SSL_VERIFY` for their deployment.

**Proxy settings** — When enabled, proxy environment variables are injected into the `openhands`, `keycloak`, and `litellm` workloads. `runtime-api` is intentionally excluded — it only communicates with the Kubernetes API and internal cluster services, so it has no outbound traffic that needs proxying. It also has a [known bug](https://github.com/kubernetes-client/python/issues/2460) in its Kubernetes Python client (v34.1.0+) that ignores `NO_PROXY`, which would cause it to hang on startup if proxy vars were set.

**Auto-computed `NO_PROXY`** — Rather than requiring the customer to manually list every internal service that should bypass the proxy, `NO_PROXY` is auto-computed to include internal k8s service hostnames, Kubernetes DNS suffixes (`cluster.local`, `svc`), and the customer's configured domain and all derived/custom hostnames. Customer-provided `NO_PROXY` entries are additive — they're appended to this base list, never replacing it. Bare domain forms (without leading dots) are used because they work correctly as suffix matches across all the HTTP clients in our stack (`httpx`, `curl`/`git`, Java).

The `NO_PROXY` computation uses Replicated template variables defined in the `when` field via a YAML folded block scalar (`>-`). Variable assignments produce no output, so the `when` value still resolves to just the `proxy_enabled` boolean. This avoids repeating the computation for each workload while keeping the logic readable and passing YAML validation.

**`SSL_VERIFY`** — Allows disabling certificate verification when the proxy uses a self-signed certificate for SSL interception. This is a stopgap — the proper fix is to support injecting custom CA certificates into the workloads, which will come in a follow-up.

**`OH_AGENT_SERVER_ENV`** — Proxy settings are also passed to the agent server runtime via the `OH_AGENT_SERVER_ENV` JSON variable, which includes `GIT_SSL_NO_VERIFY` (inverted from `SSL_VERIFY`) so that git operations inside the sandbox also respect the proxy's certificate configuration.

## Testing

To simulate a corporate proxy with SSL interception:

1. Run [mitmproxy](https://mitmproxy.org/) locally to act as an intercepting proxy
2. Expose it with [ngrok](https://ngrok.com/) to get a publicly reachable proxy URL: `ngrok tcp 8080`
3. In the KOTS admin console, enable proxy and set:
   - `HTTP_PROXY` / `HTTPS_PROXY` to the ngrok TCP URL (e.g., `http://0.tcp.ngrok.io:12345`)
   - `SSL_VERIFY` to disabled (mitmproxy uses its own CA)
4. Deploy and verify:
   - `openhands`, `keycloak`, and `litellm` pods have the proxy env vars set
   - `NO_PROXY` includes internal services and configured domain hostnames
   - `runtime-api` does **not** have proxy env vars
   - Watch the mitmproxy flow logs to confirm traffic is routing through the proxy — in testing, traffic to GitHub was visible flowing through mitmproxy
   - Internal service-to-service traffic (e.g., openhands → keycloak) bypasses the proxy

## Helm Chart Checklist

## Additional Notes

No Helm charts were modified — changes are limited to the Replicated manifests (`replicated/config.yaml` and `replicated/openhands.yaml`). The proxy configuration is injected via `optionalValues` with `recursiveMerge: true`, so existing deployments without proxy enabled are unaffected.

### DIFF
--- a/replicated/openhands.yaml
+++ b/replicated/openhands.yaml
@@ -409,7 +409,7 @@ spec:
         {{repl $llmDomain := $derive | ternary (printf "llm-proxy.%s" $bd) (ConfigOption "llm_proxy_hostname") }}
         {{repl $runtimeApiDomain := $derive | ternary (printf "runtime-api.%s" $bd) (ConfigOption "runtime_api_hostname") }}
         {{repl $runtimeBaseDomain := $derive | ternary (printf "runtime.%s" $bd) (ConfigOption "runtime_base_hostname") }}
-        {{repl $computedNoProxy := "localhost,127.0.0.1,10.0.0.0/8,svc,svc.cluster.local,cluster.local,kubernetes,keycloak,openhands-service,openhands-litellm,openhands-minio,oh-main-runtime-api,oh-main-lite-llm,oh-main-clickhouse,oh-main-langfuse" }}
+        {{repl $computedNoProxy := "127.0.0.1,cluster.local,keycloak,keycloak-headless,kubernetes,localhost,oh-main-clickhouse,oh-main-langfuse,oh-main-lite-llm,oh-main-runtime-api,openhands-integrations-service,openhands-litellm,openhands-mcp-service,openhands-minio,openhands-runtime-api,openhands-service,svc" }}
         {{repl if ne $bd "" }}{{repl $computedNoProxy = printf "%s,%s" $computedNoProxy $bd }}{{repl end }}
         {{repl if ne $appDomain "" }}{{repl $computedNoProxy = printf "%s,%s" $computedNoProxy $appDomain }}{{repl end }}
         {{repl if ne $authDomain "" }}{{repl $computedNoProxy = printf "%s,%s" $computedNoProxy $authDomain }}{{repl end }}


aivong-openhands:
> @aivong-openhands Ah, yeah, ngrok requires a subscription for tcp tunnels. I had the proxy running on my local machine and I am using ngrok to route traffic to it. I just did that since it was the lowest friction.
> 
> You'll need to expose `mitmproxy` _somewhere_ where the workloads can connect to it. Could always stand up another VM and expose the proxy, but it'd probably be cheaper to just expense an ngrok account. I find uses for it all the time.

Ok thanks for the info, I'll give that ngrok and mitmproxy setup a shot!

PR #537, opened 2026-04-08
Remove automation strip-prefix middleware

jlav:
## Description

The automations service now serves under `/api/automation` directly (see OpenHands/automation#37), so the Traefik `StripPrefix` middleware in this chart is no longer needed. This removes the middleware and its ingress annotation, letting the ingress route traffic straight through.

## Helm Chart Checklist

## Additional Notes

Depends on OpenHands/automation#37 being deployed first — without the base path change in the service, removing the middleware would break routing.


malhotra5 (approved):
LGTM

PR #559, opened 2026-04-18
Port runtime-api rollout hardening into cloud chart

aivong-openhands:
## Summary
- port the runtime-api chart hardening from OpenHands/runtime-api#512 into `charts/runtime-api`
- add `minReadySeconds` and a conservative rolling update strategy to the runtime-api Deployment
- make runtime-api probes configurable in values and add a startup probe alongside the existing readiness/liveness behavior

## Source
- OpenHands/runtime-api#512

## Incident context
A recent production Datadog investigation for `Runtime API HTTP 5xx errors` pointed to a short burst of runtime-api failures during scale-up/rollout:
- monitor: `Runtime API HTTP 5xx errors > {{threshold}} in 15 minutes` (Datadog monitor `2248011`)
- incident window: `2026-04-18 18:49:41Z` to `2026-04-18 18:49:46Z`
- symptom: Traefik `502` responses on `GET /sessions/<id>` across multiple session IDs
- validated separately: Service, EndpointSlice, and Ingress wiring were correct; the `:80` backends seen in Traefik logs were the error-page middleware, not the original runtime-api upstream
- most likely cause: a newly added runtime-api pod began receiving traffic before it was fully ready

This PR reduces that rollout-time risk by ensuring new runtime-api pods must pass startup/readiness checks before becoming routable, while also keeping pods ready for a minimum dwell period and avoiding unavailable capacity during rolling updates.

## Validation
- `helm lint ./charts/runtime-api`
- `helm dependency build ./charts/runtime-api`
- `helm template runtime-api ./charts/runtime-api`

## Notes
This keeps the OpenHands Cloud runtime-api chart aligned with the rollout/readiness safeguards proposed in the upstream runtime-api chart PR.

_This draft PR was created and updated by an AI assistant (OpenHands) on behalf of the user._

### DIFF
--- a/charts/openhands/Chart.yaml
+++ b/charts/openhands/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 description: OpenHands is an AI-driven autonomous software engineer
 name: openhands
 appVersion: cloud-1.22.1
-version: 0.4.8
+version: 0.4.9
 maintainers:
   - name: rbren
   - name: xingyao
@@ -38,7 +38,7 @@ dependencies:
     condition: replicated.enabled
   - name: runtime-api
     repository: oci://ghcr.io/all-hands-ai/helm-charts
-    version: 0.2.10
+    version: 0.2.11
     condition: runtime-api.enabled
   - name: automation
     repository: oci://ghcr.io/all-hands-ai/helm-charts
--- a/charts/runtime-api/Chart.yaml
+++ b/charts/runtime-api/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: runtime-api
 description: A Helm chart for the FastAPI application
-version: 0.2.10 # Change this to trigger a new helm chart version being published
+version: 0.2.11 # Change this to trigger a new helm chart version being published
 appVersion: "1.0.0"
 dependencies:
   - name: postgresql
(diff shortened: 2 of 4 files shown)


neubig:
@openhands is /health, /ready, or /alive the appropriate endpoint to probe here? You can take a look at the endpoints and the differences between them here: https://github.com/OpenHands/software-agent-sdk/blob/b7fb93274d9974d06de9a89a8d8ca3159a022b2d/openhands-agent-server/openhands/agent_server/server_details_router.py#L61

neubig:
Good catch — for **this chart specifically, `/health` is the right endpoint today**.

The distinction in the linked `server_details_router.py` is:
- `/alive`: process is up
- `/health`: basic health/process-up check
- `/ready`: only returns 200 after initialization is complete

But those endpoints are from the **sandbox agent server**, not from the `runtime-api` control-plane service that `charts/runtime-api` deploys.

I checked the `OpenHands/runtime-api` source as well:
- `main.py` exposes `/health`
- it does **not** expose `/ready` or `/alive` for the runtime-api service itself
- the `/ready` probe usage in `k8s_resources.py` is for the **per-runtime sandbox pods**, not for the runtime-api API deployment

So in this PR, probing `/health` is the correct choice for the deployment as it exists today. If we want stricter readiness semantics for the runtime-api service itself, the right follow-up would be to add a real `/ready` endpoint there and then point **startup/readiness** at `/ready` while keeping **liveness** on `/health` or `/alive`.

_This comment was created by an AI assistant (OpenHands) on behalf of the user._

hieptl (approved):
Thank you! 🙏

PR #561, opened 2026-04-20
Add support for LMNR (Laminar) configuration

juanmichelini:
## Summary

Add support for Laminar monitoring/analytics configuration to both the `openhands` and `automation` charts.

### openhands chart (`charts/openhands/`)
- Add `laminar.apiKeyFromSecret` config option to inject `LMNR_PROJECT_API_KEY` from a Kubernetes secret
- Add `laminar.apiUrl` config option to set `LMNR_API_URL` environment variable
- Only injects env vars when `laminar.enabled: true`

### automation chart (`charts/automation/`)
- Add `lmnrApiKeyFromSecret` config option to inject `LMNR_PROJECT_API_KEY` from a Kubernetes secret
- Add `lmnrApiUrl` config option to set `LMNR_API_URL` environment variable

### 1. Create the Kubernetes secret

```bash
kubectl create secret generic lmnr-api-key \\
  --from-literal=LMNR_PROJECT_API_KEY="your-actual-api-key" \\
  -n your-namespace
```

### 2. Configure via Helm values

**For openhands chart:**
```yaml
laminar:
  enabled: true
  apiKeyFromSecret:
    name: lmnr-api-key
    key: LMNR_PROJECT_API_KEY
  apiUrl: "https://app.lmnr.ai"
```

**For automation chart:**
```yaml
lmnrApiKeyFromSecret:
  name: lmnr-api-key
  key: LMNR_PROJECT_API_KEY

lmnrApiUrl: "https://app.lmnr.ai"
```

## Environment Variables Added

| Variable | Chart | Source |
|----------|-------|--------|
| `LMNR_PROJECT_API_KEY` | openhands, automation | From Kubernetes secret |
| `LMNR_API_URL` | openhands, automation | From Helm value |


charts/openhands/templates/_env.yaml
```diff
@@ -337,6 +337,19 @@
 - name: DD_TRACE_SAMPLING_RULES
   value: '[{"service":"deploy","name":"fastapi.request","resource":"/integration/*","sample_rate":0.05}, {"service":"deploy","name":"fastapi.request","resource":"/slack/on-*","sample_rate":0.05}]'
 {{- end }}
+{{- if .Values.laminar.enabled }}
+{{- if .Values.laminar.apiKeyFromSecret }}
+- name: LMNR_PROJECT_API_KEY
```
aivong-openhands, line 351: This will enable the API key to be set in openhands. Does the SDK already look for this environment variable and set it for the agent server runtimes?

Otherwise we could look into setting `OH_AGENT_SERVER_ENV`. It is a json string. Not sure if there is a way to use a templated property to account for other non laminar values. We currently set this OH_AGENT_SERVER_ENV in two places in the replicated install (for HTTP_PROXY and VERTEXAI_CREDENTIALS):
https://github.com/All-Hands-AI/OpenHands-Cloud/blob/main/replicated/openhands.yaml#L408 
https://github.com/All-Hands-AI/OpenHands-Cloud/blob/main/replicated/openhands.yaml#L438
So i'd like to make sure those configurations do not break 🤔

aivong-openhands:
@juanmichelini you can test this preview chart of openhands in a deploy repo feature environment like https://github.com/OpenHands/deploy/pull/3519

charts/openhands/templates/_env.yaml
juanmichelini: @aivong-openhands yes, software-agent-sdk looks for the LMNR_PROJECT_API_KEY env var.
See https://github.com/search?q=repo%3AOpenHands%2Fsoftware-agent-sdk%20LMNR_PROJECT_API_KEY&type=code

It's how we trace laminar withe PR review bot, we just set the env var for LMNR_PROJECT_API_KEY  and the software-agent-sdk does the rest https://github.com/OpenHands/extensions/blob/main/plugins/pr-review/action.yml

juanmichelini:
> @juanmichelini you can test this preview chart of openhands in a deploy repo feature environment like [OpenHands/deploy#3519](https://github.com/OpenHands/deploy/pull/3519)

I would need to set the secret somewhere right? how do I do that?

charts/openhands/templates/_env.yaml
aivong-openhands: Cool it sounds like the LMNR_PROJECT_API_KEY environment variable could be set directly because the agent runs in a script. I think OH_AGENT_SERVER_ENV will still need to be set so the agent server runtimes can inherit the laminar environment variable from openhands. We can see how your preview chart testing in a deploy PR goes to verify this

aivong-openhands:
> > @juanmichelini you can test this preview chart of openhands in a deploy repo feature environment like [OpenHands/deploy#3519](https://github.com/OpenHands/deploy/pull/3519)
> 
> I would need to set the secret somewhere right? how do I do that?

You will get your own namespace for your feature environment in the k8s cluster. I think you can create the secret in that namespace manually and then reference it in your PR for testing

charts/openhands/templates/_env.yaml
aivong-openhands: @jlav the `OH_AGENT_SERVER_ENV` is set in a couple of places in the replicated install for the proxy and the vertex AI configs. Is it possible to have some helm or k8s helper object to construct the full json string value for `OH_AGENT_SERVER_ENV` with all the pieces of info it needs without overwriting values?

juanmichelini:
UPDATE: 

propagating LMNR_ env vars with https://github.com/OpenHands/OpenHands/pull/14123

and using default empty LMNR_BASE_URL (api.laminar.ai), it works.

evidence:

image

https://laminar.sh/project/96bcf449-9fb8-4747-bbcf-d4e45a55b8d0/traces?pastHours=24

juanmichelini:
launching a second test with LMNR_BASE_URL=laminar-api.test-lmnr-pr-561.staging.all-hands.dev

juanmichelini:
**Update**: Added gRPC ingress support to enable external trace export.

The Laminar app-server exposes:
- HTTPS API: Port 443 (external) → 8080 (container)
- **gRPC: Port 8443 (external) → 8001 (container)**

This change adds a dedicated gRPC ingress (`laminar-app-server-grpc-ingress-gcp.yaml`) that routes gRPC traffic on port 8001, resolving the `StatusCode.UNAVAILABLE` errors when external services try to export traces.

Changes:
- New template for gRPC ingress routing
- Added `appServer.service.port` and `appServer.grpcPort` values
- Added `global.cloudProvider` setting

This PR was updated by OpenHands AI assistant.

juanmichelini:
**Update 2**: Simplified gRPC ingress to use the **same hostname** as HTTP.

Changes:
- Both ingresses now use `laminar-api.test-lmnr-pr-561.staging.all-hands.dev`
- Added `traefik.ingress.kubernetes.io/backend-protocol: h2c` annotation for HTTP/2
- Reuses same TLS certificate (`laminar-app-server-tls`)


After merging, redeploy PR #3860 and test with:
```python
Laminar.initialize(
    base_url="https://laminar-api.test-lmnr-pr-561.staging.all-hands.dev",
    grpc_port=8443,
    project_api_key=os.environ["LMNR_PROJECT_API_KEY"]
)
```

This PR was updated by OpenHands AI assistant.

juanmichelini:
**Update 3**: Fixed service port issue.

Problem: The laminar subchart only exposes port 8000 (HTTP) in its service, but the gRPC ingress needs port 8001.

Solution: Created `laminar-app-server-service.yaml` template that overrides the subchart service to expose both ports:
- Port 8000 (HTTP)
- Port 8001 (gRPC)

Redeploy PR #3860 after this merges to get the full fix.

This PR was updated by OpenHands AI assistant.

juanmichelini:
## 🔧 Fixed gRPC Ingress Routing Issue

### Problem
The gRPC ingress was conflicting with the HTTP ingress because both used:
- Same hostname: `laminar-api.test-lmnr-pr-561.staging.all-hands.dev`
- Same path: `/` (root)

This caused all traffic (including gRPC) to route to the HTTP server on port 8000, resulting in **404 errors** for gRPC requests like:
```
POST /opentelemetry.proto.collector.trace.v1.TraceService/Export HTTP/1.1" 404
```

### Solution
Updated the gRPC ingress to match only **OpenTelemetry gRPC paths** (`/opentelemetry.*`) which routes them to port 8001 where the actual gRPC server is listening.

### Routing After Fix
- ✅ **gRPC traffic** (`/opentelemetry.*`) → Port **8001** (gRPC server)
- ✅ **HTTP/REST traffic** (`/api/v1/...`, etc.) → Port **8000** (HTTP server)

### Testing
After redeployment, traces should be successfully ingested to the custom Laminar instance. You can verify by:
1. Sending a message in the OpenHands UI
2. Checking the Laminar UI for new traces
3. Monitoring pod logs for successful trace POSTs (should see HTTP 200 instead of 404)

Co-authored-by: openhands <openhands@all-hands.dev>

juanmichelini:
## 🔧 Updated Fix: Using gRPC Subdomain

### Issue with Previous Fix
The SDK was trying to connect to `laminar-api.test-lmnr-pr-561.staging.all-hands.dev:8443`, but Kubernetes Ingress **only handles ports 80 and 443**. The LMNR/OpenTelemetry SDK automatically adds `:8443` for gRPC connections, which wasn't accessible through the ingress.

### New Solution: Separate gRPC Subdomain
Changed the gRPC ingress to use a dedicated subdomain: `grpc.laminar-api.test-lmnr-pr-561.staging.all-hands.dev`

**Benefits:**
- Uses standard HTTPS port **443** (compatible with Kubernetes Ingress)
- Dedicated hostname avoids path-based routing conflicts
- Separate TLS certificate for isolation
- Standard pattern used by many gRPC services

### Configuration Required

After redeployment, update the secret to use the **gRPC subdomain** with explicit port:

```bash
kubectl delete secret lmnr-base-url -n openhands-test-lmnr-pr-561
kubectl create secret generic lmnr-base-url \
  --from-literal=LMNR_BASE_URL="grpc.laminar-api.test-lmnr-pr-561.staging.all-hands.dev:443" \
  -n openhands-test-lmnr-pr-561
```

**Note:** Including `:443` explicitly prevents the SDK from adding `:8443`.

### Architecture After Fix

```
HTTP/REST API:
  laminar-api.test-lmnr-pr-561.staging.all-hands.dev:443 
  → Port 8000 (HTTP server)

gRPC Traces:
  grpc.laminar-api.test-lmnr-pr-561.staging.all-hands.dev:443 
  → Port 8001 (gRPC server)
```

### Testing
1. Redeploy with the updated Helm chart
2. Update the `lmnr-base-url` secret as shown above
3. Restart OpenHands pods
4. Send a message and check for traces in the Laminar UI
5. Verify logs show successful exports (HTTP 200, not UNAVAILABLE)

Co-authored-by: openhands <openhands@all-hands.dev>

charts/openhands/templates/laminar-app-server-grpc-ingress-gcp.yaml
```diff
@@ -0,0 +1,37 @@
...
+    - host: {{ $grpcHostname }}
+      http:
+        paths:
+          - path: /
+            pathType: Prefix
+            backend:
+              service:
+                name: laminar-app-server-service
+                port:
+                  number: {{ .Values.laminar.appServer.grpcPort | default 8001 }}
```
aivong-openhands: Could we just add another rule for port 8001 for grpc instead of creating a new ingress object? https://github.com/All-Hands-AI/OpenHands-Cloud/blob/853f7b0c25dd388e41f0f241249ae546abfa1289/charts/openhands/templates/laminar-app-server-ingress-gcp.yaml#L25

charts/openhands/templates/laminar-app-server-service.yaml
```diff
@@ -0,0 +1,26 @@
+{{/*
+  Service for exposing Laminar app-server with both HTTP and gRPC ports.
+  This service aggregates both ports since the laminar subchart only exposes HTTP.
+*/}}
+{{- if and .Values.laminar.enabled (eq .Values.laminar.global.cloudProvider "gcp") }}
+apiVersion: v1
+kind: Service
+metadata:
+  name: laminar-app-server-service
```
aivong-openhands: I believe this service already exists in the staging cluster 🤔 I think we should modify that with laminar helm values if possible instead of creating a new service object.

charts/openhands/templates/laminar-app-server-grpc-ingress-gcp.yaml
aivong-openhands: Interesting there is a laminar helm chart configuration doc note that mentions the ingress object is not for grpc traffic https://github.com/lmnr-ai/lmnr-helm/blob/main/CONFIGURATION.md#app-server-ingress-gcp-and-other-providers. It would probably be a good idea to reach out to laminar is ask them how they recommend we enable grpc support

aivong-openhands: I also see some traefik configuration for the grpc route https://github.com/lmnr-ai/lmnr-helm/blob/main/examples/networking/traefik-app-server.yaml#L45. We use traefik so maybe some missing configuration for exposing this grpc port 8001 🤔

aivong-openhands: https://github.com/lmnr-ai/lmnr-helm/pull/19/changes#diff-4457586151b357a84bcebbf3a4964fa3d81e07e0b51c53615dfd62fee6292489. There were changes to the laminar chart to support gRPC on traefik with TLS. We don't have this updated laminar chart in staging though so I can coordinate with you on that next week. Then you won't have to deal with gRPC/ingress in your PR

charts/openhands/templates/_env.yaml
jlav: I think we can pull `OH_AGENT_SERVER_ENV` up into the values file as a map, so you'd get a values entry that looks like this:
```yaml
agentServerEnv:
   KEY: value
```

Then build a helper that merges that with other values gathered from elsewhere in the chart, like `LMNR_PROJECT_API_KEY`, and prints out the JSON. You can look at how we construct user-provided env vars as an example.

With that in place, the replicated installer can add to `agentServerEnv` just using a recursive merge.

aivong-openhands: Thanks @jlav! I created https://github.com/All-Hands-AI/OpenHands-Cloud/issues/578 to add that values entry and helper.

Looks like @juanmichelini found a good alternative by auto forwarding the LMNR environment variables from openhands to the agent server runtimes so we are good to go here and can avoid the use of that environment variable 👏

juanmichelini:
@openhands let's drop gRPC for now, please revert back to f3162de1a0a587b26a43b9d1e5ba320e08ed32b7

charts/openhands/templates/_env.yaml
jlav: By the way is `LMNR_PROJECT_API_KEY` something that we're using in production and intend to expose to every user?

And one more concern, runtime API matches requests to warm runtimes only when the requested environment matches the warm runtime exactly. If the `openhands` server is passing `LMNR` variables through in requests, but the warm runtimes don't have that same variable configured, requests never route to the warm runtimes.

juanmichelini:
@aivong-openhands @neubig 

it works! deployed successfully on https://github.com/OpenHands/deploy/pull/3860/

created a conversation at 

https://test-lmnr-pr-561.staging.all-hands.dev/conversations/610d501454a34857afddc4e7d58af3c8

and got traces at 

https://laminar.test-lmnr-pr-561.staging.all-hands.dev/project/8c0b9c02-b2dd-4d9f-91aa-f5d6f42f7b0e/traces?pastHours=4&traceId=edbd3910-6ad4-71e9-6748-a347c399b6e7&chat=true

See screenshots with evidence

image

image

charts/automation/values.yaml
```diff
@@ -152,6 +167,22 @@ automationWebhookSecretFromSecret:
   name: automation-webhook-secret
   key: webhook-secret
 
+# Laminar project API key from K8s secret
```
aivong-openhands: is this a duplicate section?

charts/openhands/values.yaml
```diff
@@ -352,7 +353,24 @@ keycloak:
 
 laminar:
   # Enable this if you want to use laminar
-  enabled: false
+  enabled: true
```
aivong-openhands: Should stay defaulted to false and enabled in specific environments like SaaS staging for now and replicated when users opt into analytics

charts/openhands/templates/_env.yaml
jlav: Hey @aivong-openhands and @juanmichelini, mind looking at my question from above?

One concern is a security issue and the other is a breaking bug, so I don't want them to slip through.

aivong-openhands: Laminar is not in production yet but when it is, I think the LMNR_PROJECT_API_KEY will be auto forwarded when it is available (when laminar is configured for an environment and the k8s secrets are set/created).

LMNR_PROJECT_API_KEY will only be set when laminar is enabled. Would it be a runtime API change to propagate the environment variable when it is set to the warm runtimes?

charts/automation/values.yaml
juanmichelini: it's twice, one for automation and one for openhands, you mean I can delete one?

charts/openhands/values.yaml
juanmichelini: Set to false!

charts/automation/values.yaml
juanmichelini: actually it was thrice, I removed completely from automation since the automation file does not have the field 

laminar:
enable: bool

aivong-openhands: yeah it was showing up twice for automation

charts/openhands/templates/_env.yaml
jlav: Is `LMNR_PROJECT_API_KEY` something that we can safely expose, though? I'm concerned about adding an API key directly to the agent sandboxes, because that means anyone can use the key.

I don't think there's anything in place now to dynamically propagate environment variables to the warm runtimes, unfortunately. So there would have to be some engineering work to support that. The quickest way to ensure requests will land on a warm runtime is to make sure that the warm runtime config map will always match sandbox requests, exactly. This is a bit of an architectural challenge.

aivong-openhands: @juanmichelini maybe we can ask in a conversation in your feature environment if it can tell us the laminar api key?

It sounds like we will break warm runtimes with this change too so that is a blocker.

jlav: Unfortunately anything added to the agent sandbox environment variables is accessible to the user. It's an untrusted sandbox so the user can easily inspect it and steal this key.

I would look into whether or not the Laminar key provides read access to our data, and if there's a way to limit the scopes to only allow sending traces. If not, we'd probably need to ask Laminar to make that a feature, or build a proxy to hide the key.

aivong-openhands: Turns out the default key does provide read access. However, we can switch to an ingest only key that can only send traces.

juanmichelini: agreed, we can create a new one with ingest only 
image

jlav: Awesome! Thanks for looking into it.

charts/openhands/values.yaml
```diff
@@ -353,6 +353,23 @@ keycloak:
...
+
+  # Laminar base URL from K8s secret
+  # Used for Laminar monitoring/analytics
+  baseUrlFromSecret:
+    name: lmnr-base-url
+    key: LMNR_BASE_URL
+
+  # Force HTTP instead of HTTPS for Laminar connections
+  # Useful for development/test environments without TLS
+  forceHttp: true
```
aivong-openhands, line 373: @juanmichelini will we want this default for staging/prod SaaS laminar?

juanmichelini: @aivong-openhands yes, it won't work with gRPC so forcing htttp by default is correct

aivong-openhands (approved):

juanmichelini:
@aivong-openhands next step would be to set staging with laminar enabled: true and the env vars.
would that be a PR in the deploy repo?

aivong-openhands:
> @aivong-openhands next step would be to set staging with laminar enabled: true and the env vars. would that be a PR in the deploy repo?

yes it would be

PR #567, opened 2026-04-22
feat: Automation frontend path

raymyers:
## Description

Add `/automations` frontend path, in addition to the existing backend path`/api/automation`. These will both point to the same pod and docker image.

## Helm Chart Checklist

- [X] I have updated the `version` field in `Chart.yaml` for each modified chart
- [X] I have tested the chart upgrade path from the previous version
- [X] I have verified backwards compatibility with existing values.yaml configurations
- [X] I have updated the chart's README.md if there are any breaking changes or new required values

## Additional Notes

Ticket: PLTF-1240

### DIFF
--- a/charts/openhands/Chart.yaml
+++ b/charts/openhands/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 description: OpenHands is an AI-driven autonomous software engineer
 name: openhands
 appVersion: cloud-1.22.1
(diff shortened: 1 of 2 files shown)


raymyers:
Evidence:

Screenshot 2026-04-22 at 2 33 41 PM

malhotra5 (approved):
LGTM

PR #571, opened 2026-04-23
APP-1325: enable the v1 gitlab resolver in replicated

aivong-openhands:
## Description

When the user chooses gitlab auth, enable the v1 resolver.

## Helm Chart Checklist

## Additional Notes

Screenshot 2026-04-23 at 6 09 12 PM
Screenshot 2026-04-23 at 6 14 14 PM


charts/openhands/values.yaml
```diff
@@ -114,7 +114,7 @@ gitlabWebhookInstallation:
 
 image:
   repository: ghcr.io/openhands/enterprise-server
-  tag: cloud-1.22.1
+  tag: sha-169d937
```
aivong-openhands: This fix in openhands was needed to register the gitlab resolver callback early enough to avoid a race condition https://github.com/OpenHands/OpenHands/pull/14110

aivong-openhands: - [x] use a cloud release tag when it is available

aivong-openhands: https://github.com/OpenHands/OpenHands/commit/9a743ff51ace13e6afd83329334c624224f60fd5 was merged with hash `9a743ff `

aivong-openhands: https://github.com/OpenHands/OpenHands/commit/137bede1f5e867a7eb17ae36e78a05ffef6ec09a would be even better because then we can surface the gitlab integration settings

jlav (approved):

charts/openhands/values.yaml
aivong-openhands: @mamoodi does openhands cloud-1.23.0 include https://github.com/OpenHands/OpenHands/commit/137bede1f5e867a7eb17ae36e78a05ffef6ec09a?

aivong-openhands: @mamoodi confirmed it does not. Also the commits should have the cloud release tag if they are included

aivong-openhands: Waiting for cloud-1.24.0 then

aivong-openhands: I see https://github.com/OpenHands/OpenHands/commit/137bede1f5e867a7eb17ae36e78a05ffef6ec09a has been tagged with cloud-1.24.0

PR #592, opened 2026-04-29
FDE-49: add Azure OpenAI support to Replicated installer

ak684:
Wire azure_endpoint, azure_api_version, and azure_deployment_name through config -> secrets -> litellm-env-secrets + openhands-env-secrets. Adds an optionalValues block in replicated/openhands.yaml mirroring the vertex AI pattern: sets LITELLM_DEFAULT_MODEL, populates OH_AGENT_SERVER_ENV with Azure credentials for direct agent-server access, and registers an azure-<deployment> entry in LiteLLM's model_list using AZURE_API_KEY / AZURE_API_BASE / AZURE_API_VERSION env vars.

Also fixes a pre-existing bug where azure_endpoint was collected by the installer config form but never plumbed into any secret or env var, and bumps charts/openhands-secrets Chart.yaml from 0.1.6 to 0.1.7 to reflect the values schema change.

Screenshot 2026-04-30 at 12 02 04 PM

## Helm Chart Checklist

## Additional Notes

---
Linear: FDE-49


replicated/openhands.yaml
```diff
@@ -521,30 +428,37 @@ spec:
             HTTPS_PROXY: '{{repl ConfigOption "https_proxy"}}'
             NO_PROXY: '{{repl $computedNoProxy}}'
             SSL_VERIFY: '{{repl if ConfigOptionEquals "ssl_verify" "1"}}True{{repl else}}False{{repl end}}'
+    - when: '{{repl ConfigOptionEquals "llm_provider" "azure" }}'
+      recursiveMerge: true
+      values:
+        env:
+          LITELLM_DEFAULT_MODEL: 'litellm_proxy/azure-{{repl ConfigOption "azure_deployment_name"}}'
+          OH_AGENT_SERVER_ENV: '{{repl printf "{\"AZURE_API_KEY\": \"%s\", \"AZURE_API_BASE\": \"%s\", \"AZURE_API_VERSION\": \"%s\"}" (ConfigOption "azure_api_key") (ConfigOption "azure_endpoint") (ConfigOption "azure_api_version") }}'
```
aivong-openhands, line 536: I think it will be good to add the self signed cert environment variables here too https://github.com/All-Hands-AI/OpenHands-Cloud/pull/587/changes#diff-1f22149bf096f4f567331cb9db29c199ebc21656ce17dc4f17b562c472d86edc. This was merged in yesterday

jlav: Why do we set these env vars on the agent at all? Shouldn't these just be wired directly into only LiteLLM configs, then proxied via LiteLLM?

aivong-openhands: I added support for both the LiteLLM proxy models and the direct models for Vertex AI. If we want to only support LiteLLM proxy models then we should document that decision and update the codebase

jlav: What do you mean by direct models? I wasn't aware there was an option besides proxying via LiteLLM.

aivongtest342423: It's when the configured model does not have llm_proxy/ in front of the name (not using one of the proxy config models). I think the agent server runtimes do use LiteLLM still via the SDK but not via the LiteLLM proxy instance that we stand up

aivongtest342423: ^ this is still me btw just on a test account

jlav (approved):

replicated/openhands.yaml
jlav: Okay, I don't think that's something we want to allow. Am I interpreting this correctly that if those environment variables are set that requests will only be directed directly to the LLM provider via the SDK, but never pass through the proxy? Which would then completely supersede the LLM settings in OpenHands?

To fully support the feature set for OHE, we need LLM requests to always be proxied via the LiteLLM proxy. I believe setting LLM configuration settings directly on the agent server is really only intended for single-user, OSS usage.

aivong-openhands: Ok we can remove the configurations for the direct approaching of using the LiteLLM SDK to access models then. There was no real ask for it. I just saw that it was another way to access models when testing Vertex AI integration so I added it

jlav: For whenever we can get around to it:
[PLTF-2043: Remove Direct Model Access configuration from Replicated Installation](https://linear.app/all-hands-ai/issue/PLTF-2043/remove-direct-model-access-configuration-from-replicated-installation)

PR #605, opened 2026-05-04
PLTF-2297: add LMNR_HTTP_PORT support for self-hosted Replicated deployments

aivong-openhands:
## Summary
- Add `laminar.httpPort` value to the openhands chart (default `null`)
- Render `LMNR_HTTP_PORT` env var when `laminar.httpPort` is set
- Set `httpPort: 8000` in the Replicated analytics when block

The lmnr SDK defaults to port 443 for HTTP transport. Self-hosted Replicated environments expose the laminar app server on port 8000 (plain HTTP via internal ClusterIP), so without this override traces are sent to the wrong port and silently dropped.

Verified working on `replicated-01.aws.aivong.platform-team.all-hands.dev`.


aivong-openhands:
@openhands look at failing validate chart version PR check and bump only the necessary chart versions

PR #606, opened 2026-05-04
[PLTF-2869] Add Microsoft Entra ID (service principal) auth for Azure OpenAI

ak684:
## Summary

Adds an Azure OpenAI Authentication Method select to the Replicated installer config with two options:

- **API Key** (existing behavior, unchanged)
- **Microsoft Entra ID — Service Principal** (new) — customer pastes tenant_id, client_id, client_secret. LiteLLM authenticates to Microsoft Entra and uses the resulting bearer token to call Azure OpenAI. No long-lived API keys on the cluster.

## Testing

End-to-end validated against a real Azure OpenAI resource (`alona-openai`, eastus2, `gpt-4.1` deployment) with Microsoft Entra ID service principal auth:

- Service principal created in tenant; granted `Cognitive Services OpenAI User` role on the OpenAI resource
- Replicated headless install on AWS embedded-cluster VM with `azure_auth_method=service_principal`
- All app pods Ready post-install
- Chat completion through the LiteLLM proxy returned a valid response from gpt-4.1, confirming the OAuth handshake to `login.microsoftonline.com` and bearer-token call to Azure both succeed
- Conversation through the OpenHands GUI returned a real LLM response
- Cluster inspection confirms only SP credentials exist in `litellm-env-secrets` (no `AZURE_API_KEY` env var or secret value anywhere) — the response could only have come through Entra ID auth

Backwards compatibility: default `azure_auth_method=api_key` preserves existing customer behavior.

## Helm Chart Checklist

- [x] I have updated the `version` field in `Chart.yaml` for each modified chart *(openhands-secrets: 0.1.10 → 0.1.11)*
- [x] I have updated the chart's README.md if there are any breaking changes or new required values *(no breaking changes; chart README does not currently document LLM-specific values)*

Screenshot 2026-05-04 at 1 59 08 PM

Screenshot 2026-05-04 at 12 46 03 PM

---
Linear: PLTF-2869


replicated/config.yaml
```diff
@@ -176,15 +176,45 @@ spec:
           type: password
           when: 'repl{{ ConfigOptionEquals "llm_provider" "mistral" }}'
           required: true
+        - name: azure_auth_method
+          title: Azure Authentication Method
+          help_text: Authentication method for Azure OpenAI. Service Principal (Microsoft Entra ID) is recommended for production.
+          type: select_one
+          default: "api_key"
```
aivong-openhands, line 183: nit - it feels like we should default to the recommended service principal method

aivong-openhands (approved):

replicated/config.yaml
aivong-openhands: I see service principal is recommended for production but these replicated installs are for POCs

PR #651, opened 2026-05-21
[PLTF-2701] Restart openhands on secret config changes (checksum + CI guard)

ak684:
## What

Make a change to any secret in the KOTS config actually take effect on Deploy, instead of silently waiting for a manual pod restart.

- Add a `checksum/config-secrets` annotation to every long-running deployment that reads the secret env (openhands, openhands-integrations, openhands-mcp), set to a KOTS-rendered sha256 of every secret (password) config value (`secretsChecksum` in `replicated/openhands.yaml`). When a secret changes, the hash changes, the pod templates change, and the pods roll so the new value is loaded. CronJobs are not annotated since each run is a fresh pod that re-reads the secret. Only the hash crosses into the chart, never the secret values.
- Add a CI check (`scripts/check_secret_checksum.py` + workflow) that fails if a password config field is not in `secretsChecksum`, so a new secret can't be added without wiring it into the restart.

## Background: the situation that drove this

While wiring up an optional Bitbucket Data Center bot token (OpenHands-Cloud #649, code in OpenHands #14517), I set the token in the admin console, clicked Deploy, and nothing happened. The webhook event reached the app, but the bot-token env var on the running pod was still empty, so OpenHands kept posting as the previous identity.

Root cause: secret-backed env vars (sourced via `secretKeyRef`) are read only at pod start, and Kubernetes does not restart a pod when a referenced Secret changes. The openhands deployment already carries `checksum/*` annotations for keycloak, litellm, and the CA bundle, but none for the provider secrets, so a token change did not roll the pod. The only fix at the time was a manual `kubectl rollout restart`.

This is not specific to BBDC. The same class of bug is already filed for LiteLLM (PLTF-42) and for the image pull secret (PLTF-153), both still in Triage. The BBDC bot token is a third instance of one root cause, which is why a general fix is worth considering rather than another one-off.

## Validation

- KOTS `sha256sum` renders a real hash, and a config change rolls the pod: verified on a live Replicated install (replicated-01) using the BBDC bot token. Changing the token rolled the openhands pod and the new token loaded; restoring it rolled back. Same mechanism, more inputs, here.
- CI guard: passes when all 33 password fields are covered, and fails (naming the missing field) when one is dropped from the checksum.
- `helm template` renders the `checksum/config-secrets` annotation.

## Tradeoff: this is one option, not necessarily the option

The checksum is coarse. It restarts the openhands pod on any secret change, including secrets that only affect other components (for example an LLM key used only by LiteLLM). It also relies on the explicit field list staying in sync, which the CI guard enforces but does not remove.

Stakater Reloader watches the actual Secret objects and rolls only the workloads that reference them, with no list to maintain, at the cost of a new in-cluster controller dependency. The tracking issue compares the two approaches. This PR is the concrete checksum option referenced there.

## Merge order note

`bitbucket_data_center_bot_token` lands in #649, not main. Once #649 merges, that field must be added to `secretsChecksum` here, and the CI guard will flag it if it is missing.

## Enforcement: this guard only blocks if it is made a required check

The workflow makes the guard run, but running is not blocking. Main's rulesets are `deletion`, `non_fast_forward`, and `pull_request`, with no `required_status_checks` rule, so CI checks here are advisory: a red guard is visible but does not block merge on its own. (Classic branch protection could not be read with my token, so an admin should confirm in repo settings.)

To actually prevent a bad merge, an admin needs to add `check-secret-checksum` to the branch's required status checks, ideally strict (require up to date) so it re-evaluates against latest main before merge.

This extra setup (a workflow plus a required, strict check) is part of the cost of the checksum approach, and a point in favor of Reloader, which needs none of it. See PLTF-2701.

---
Linear: PLTF-2701


jlav (approved):
Can you just double check that the checksum is applied anywhere that references these secrets? There may be some other deployments besides `openhands` that use them.

PR #665, opened 2026-05-27
Add top-level `enabled` flag to openhands chart

jlav:
## Description

Changing the cloud release process so all dependency updates flow through the openhands chart, giving us a single source of truth for which dependency combinations have been vetted. Some dependencies will still be managed as separate Helm releases — this flag lets the umbrella chart install just those dependencies without also installing the openhands app.

`enabled: true` is the default, so existing installs are unaffected.

## Helm Chart Checklist

## Additional Notes

Verified `helm template` with defaults renders byte-identical to main (no spurious config-checksum change, so no surprise pod rollout). With `enabled: false`, no resources from `charts/openhands/templates/` render; subcharts continue to respect their own `enabled` flags so any subset can be installed alone.

minio still gates on `filestore.ephemeral` (not its own `enabled`) — left as-is to preserve existing semantics. crd-check is a library chart with no resources, so no condition is needed there.

### DIFF
--- a/charts/openhands/Chart.yaml
+++ b/charts/openhands/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 description: OpenHands is an AI-driven autonomous software engineer
 name: openhands
 appVersion: cloud-1.34.0
-version: 0.7.30
+version: 0.7.31
 maintainers:
   - name: rbren
   - name: xingyao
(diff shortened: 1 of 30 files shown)


raymyers (approved):
Looks good. I worry a bit that it's going to be easy for some future resource to miss the conditional. Wonder if there's any standard way of testing that. This is good to go though.

jlav:
Helm unit tests would catch that pretty easily. If `enabled` is false, make sure no templates are produced.

Not something to introduce as part of this PR, but one of the many, many things we could validate cheaply if we had a test suite.

PR #691, opened 2026-06-08
Allow full custom LiteLLM model strings

ak684:
## Summary

- Preserve full `provider/model` LiteLLM strings for the Replicated custom LLM provider instead of always prefixing `openai/`.
- Keep the existing OpenAI-compatible behavior for bare custom model names by continuing to render `openai/<model>`.
- Update custom model help text and add a Helm render regression test for both bare and full provider/model custom models.

## Root Cause

A customer custom LLM endpoint can be non-OpenAI-shaped, for example an Anthropic-shaped gateway. The chart previously rendered every custom model as `openai/<custom_model>`, so a full LiteLLM model string like `anthropic/claude-opus-4-7` became `openai/anthropic/claude-opus-4-7`, causing LiteLLM to call the endpoint with the wrong provider semantics.

## Validation

- `uv run --with pytest python -m pytest scripts/test_keycloak_realm_template.py scripts/test_openhands_secrets_template.py -q`
- `make lint`
- Published Replicated release sequence `937` to channel `alona/custom-llm-raw-model-prefix`.
- Installed on `instance-2` and verified app/admin HTTP 200.
- Verified live `openhands-litellm-config` renders `model: anthropic/claude-sonnet-4-5-20250929`, not `openai/anthropic/...`.

Test report: `/Users/alonaking/replicated-tests/alona/custom-llm-raw-model-prefix/2026-06-08/custom-llm-raw-model-instance-2-133352/REPORT.md`

### DIFF
--- a/replicated/config.yaml
+++ b/replicated/config.yaml
@@ -304,7 +304,7 @@ spec:
           required: true
         - name: custom_base_url
           title: Custom LLM Base URL
-          help_text: Base URL of your OpenAI-compatible LLM endpoint (e.g., https://openrouter.ai/api/v1, http://my-vllm:8000/v1)
+          help_text: Base URL of your custom LLM endpoint (e.g., https://openrouter.ai/api/v1, http://my-vllm:8000/v1)
           type: text
           when: 'repl{{ ConfigOptionEquals "llm_provider" "custom" }}'
           required: true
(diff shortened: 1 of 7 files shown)


replicated/config.yaml
```diff
@@ -316,7 +316,7 @@ spec:
           required: false
         - name: custom_model
           title: Custom LLM Model Name
-          help_text: Model identifier as accepted by the endpoint (e.g., anthropic/claude-sonnet-4-5 for OpenRouter, llama-3.1-70b-instruct for vLLM)
+          help_text: Model identifier as accepted by the endpoint. Bare names are routed as OpenAI-compatible models (e.g., llama-3.1-70b-instruct becomes openai/llama-3.1-70b-instruct). To use a non-OpenAI-shaped LiteLLM provider, enter the full provider/model string (e.g., anthropic/claude-opus-4-7).
```
jlav: If no one else is using this, can we just update the help text to say "for OpenAI compatible models, prefix your model name with `openai/`", then template in whatever the user provides, and avoid the conditional and the prefxing altogether?

LiteLLM has a bunch of bare model names, so that approach is more flexible.

ak684: Yea I thought about that! I hesitated because I genuinely was not sure what other customers are using custom llms and I didnt want to break existing set ups.. but I can ask internally and confirm what the situation actually is so that if there's an opportunity to simply on that fact, we can.

jlav (approved):

replicated/config.yaml
jlav: Okay, approved so you're unblocked. Either outcome is fine, but obviously prefer just removing the prefix if we can.

jpshackelford:
Thanks for this, @ak684!

PR #693, opened 2026-06-08
PLTF-2882: Decouple Keycloak realm provisioning from the bundled subchart

aivong-openhands:
## Description
The `keycloak-config` init container (and its `keycloak-config-script` /
`keycloak-realm-template` ConfigMaps) that creates/updates the OpenHands realm,
its OIDC client, and identity providers was gated solely on `keycloak.enabled`.
That same flag controls the bundled Keycloak subchart, so split deployments that
run Keycloak as a separate release (`keycloak.enabled: false`, `keycloak.url`
pointing at an external Keycloak) could never run realm provisioning without
standing up a second Keycloak.

This adds a `keycloak.provisionRealm` flag and an `openhands.keycloakProvisionRealm`
helper that gates these resources on `keycloak.enabled OR keycloak.provisionRealm`.

## Helm Chart Checklist
- [x] I have updated the `version` field in `Chart.yaml` (0.7.41 → 0.7.42)
- [x] No new *required* values or breaking changes (new flag defaults to `false`)

## Additional Notes
Backwards compatible: when `keycloak.enabled` is true, behavior is unchanged
(bundled deployments still provision the realm). The new flag only adds the
ability to provision against an external Keycloak. `provisionRealm` lives under
`keycloak:` for grouping; it is read only by the parent chart and ignored by the
Keycloak subchart (which has no values schema).

Verified with `helm template` across three cases: (a) `keycloak.enabled=true`
renders the realm-config init container as before; (b) `keycloak.enabled=false`
+ `keycloak.provisionRealm=true` renders the init container with no bundled
Keycloak server; (c) both false renders nothing.

### DIFF
--- a/charts/openhands/Chart.yaml
+++ b/charts/openhands/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 description: OpenHands is an AI-driven autonomous software engineer
 name: openhands
 appVersion: sha-d765b0b
-version: 0.7.42
+version: 0.7.43
 maintainers:
   - name: rbren
   - name: xingyao
--- a/charts/openhands/Chart.yaml
+++ b/charts/openhands/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 description: OpenHands is an AI-driven autonomous software engineer
 name: openhands
 appVersion: sha-d765b0b
-version: 0.7.41
+version: 0.7.42
 maintainers:
   - name: rbren
   - name: xingyao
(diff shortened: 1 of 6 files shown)


jlav (commented):
~~Needs to default to `true` in the replicated config~~
Edit: misread. All good.

jlav (approved):

charts/openhands/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 description: OpenHands is an AI-driven autonomous software engineer
 name: openhands
 appVersion: sha-d765b0b
-version: 0.7.41
+version: 0.7.42
```
aivong-openhands:
```suggestion
version: 0.7.43
```

PR #695, opened 2026-06-09
refactor(charts): embed service charts as openhands subcharts

jlav:
## Description

Follow-up to #665, which made the openhands chart the single source of truth for vetted dependency combinations by adding `enabled` flags everywhere.

This repo still vendored runtime-api, automation, plugin-directory, and integrations-hub as top-level charts, each versioned and published to GHCR on its own. But nothing consumes them standalone - they only ever ship as part of openhands. Keeping them independently versioned meant a change to any one of them needed its own version bump, its own publish, and a matching dependency-version bump in the umbrella, plus a pile of CI choreography to rewrite OCI refs to alpha previews on PRs.

This folds those four into `charts/openhands/charts/` as standard embedded subcharts. Their versions are now frozen and changes ride along with an openhands version bump. That collapses the release path to match the model #665 set up.

crd-check stays published independently, since `charts/infra` depends on it by OCI reference.

## Helm Chart Checklist

## Additional Notes

The subchart dependency entries in the umbrella `Chart.yaml` are now repository-less with `version: "*"`, kept only for their `condition` flags. The `enabled` gating from #665 carries over unchanged.

A couple of things worth flagging for review:

- helm doesn't recurse into embedded subcharts when resolving dependencies, so the embedded charts' own deps (postgresql, minio, helm-release-pruner) would silently drop out of the package. CI and the Makefile now run `helm dependency update` on each embedded subchart before packaging the umbrella.
- `.gitignore` changed from `charts/**/charts` to `charts/**/charts/*.tgz` so the embedded sources stay tracked while vendored archives stay ignored.

I verified `helm template` of the new layout renders byte-identical to the current OCI-based umbrella, `make lint` passes end-to-end (it previously couldn't even package the umbrella locally because of the unpublished automation OCI pin), and the update-script test suite is green.

I've deployed the alpha chart version published by this PR's CI pipeline to the dev environment: https://github.com/OpenHands/saas-deploy/pull/72/changes

I've also performed a replicated upgrade from a release built with this chart.


dylan-openhands:
This seems super conceptually reasonable. However, will this be the first instance of migrating into embedded subcharts?

One thought/concern is just around that upgrade path and what we can do to sanity check the upgrade in general. Specifically the upgrade from non subchart -> subchart is the most likely one to trip it up. I'm definitely not an expert (yet) on helm, so bear with me 😄

jlav:
@dylan-openhands I've deployed the alpha chart version published by this PR's CI pipeline to the dev environment: https://github.com/OpenHands/saas-deploy/pull/72/changes

I've also exercised the upgrade using one of my replicated testing environments, so we're pretty well covered!

Updated the PR description to note that.

dylan-openhands (approved):

PR #706, opened 2026-06-12
Bump Embedded Cluster to 2.13.5+k8s-1.33

ak684:
## What

1. Bumps the Embedded Cluster version pin in `replicated/embedded-cluster.yaml` from `2.13.0+k8s-1.33` to `2.13.5+k8s-1.33`.
2. Removes the Helm-4-only `--force-conflicts` flag from the `openhands` and `openhands-secrets` HelmChart `helmUpgradeFlags` — **required** for deploys to work on EC ≥2.13.3 (see validation comment). These must ship together.

## Why

A customer needs to rotate the Admin Console TLS certificate. EC 2.13.2 introduced the supported one-liner for this:

```bash
sudo ./openhands admin-console update-tls --tls-cert <cert> --tls-key <key>
```

Without it, the options are the anonymous-upload Admin Console flow (Replicated docs flag it as a temporary security hole) or manually patching the `kotsadm-tls` secret via kubectl.

## Why 2.13.5 specifically

- Latest patch on the 2.13.x line — stays on Kubernetes 1.33 (1.33.6 → 1.33.8 node patch only), no k8s minor jump, no Helm 4 migration (that lands in EC 2.15.0).
- **Avoids 2.13.2 exactly**: 2.13.1/2.13.2 shipped Helm v4 with default server-side apply, which could fail application upgrades; 2.13.3 fixed it by pinning Helm v3.
- Small KOTS bump 1.129.1 → 1.129.4; no breaking changes.
- Fixes the `embedded-cluster-operator` CrashLoopBackOff seen on long-lived installs (operator pod limits removed — 2.13.2 release note; confirmed fixed on R02).
- Picks up Go 1.25.7 / crypto/tls CVE-2025-68121 fixes.

Release notes: https://docs.replicated.com/release-notes/rn-embedded-cluster

## Why the --force-conflicts removal is coupled

EC 2.13.3+ pins the KOTS-bundled Helm back to v3, which has no `--force-conflicts` flag — with it, every chart deploy on the upgraded cluster fails with `Error: unknown flag: --force-conflicts` (reproduced on R02). The flag was added in #648 for Helm-4 server-side-apply field-manager conflicts, which don't occur under Helm v3's client-side apply. Cluster upgrades always complete before the app deploys, so this release's charts always deploy under Helm v3.

## Rollout

Existing installs pick up the EC infrastructure upgrade automatically on their next admin-console app update. Expect a longer deploy than chart-only upgrades (k0s + kotsadm restart; ~15 min observed on m8i.2xlarge) and a brief API-server blip. **Known hang risk on single-node installs:** warm runtime/sandbox pod PDBs (`maxUnavailable: 0`) block the node drain — see validation comment.

## Validation

In-place 2.13.0 → 2.13.5 upgrade on R02 — PASS. Details in the PR comment; full report in `~/replicated-tests/alona-ec-2.13.5/2026-06-11/REPORT.md`. `update-tls` verified working post-upgrade.

### DIFF
--- a/replicated/embedded-cluster.yaml
+++ b/replicated/embedded-cluster.yaml
@@ -3,7 +3,7 @@ kind: Config
 metadata:
   name: embedded-cluster-config
 spec:
-  version: "2.13.0+k8s-1.33"
+  version: "2.13.5+k8s-1.33"
   domains:
     proxyRegistryDomain: "images.r9.all-hands.dev"
     replicatedAppDomain: "updates.r9.all-hands.dev"
--- a/replicated/openhands.yaml
+++ b/replicated/openhands.yaml
@@ -8,7 +8,7 @@ spec:
   releaseName: openhands
   namespace: openhands
   weight: 10
-  helmUpgradeFlags: ["--wait", "--timeout", "600s", "--force-conflicts"]
+  helmUpgradeFlags: ["--wait", "--timeout", "600s"]
   values:
     global:
       imageRegistry: 'images.r9.all-hands.dev'
--- a/replicated/secrets.yaml
+++ b/replicated/secrets.yaml
@@ -8,7 +8,7 @@ spec:
   namespace: openhands
   releaseName: openhands-secrets
   weight: 1
-  helmUpgradeFlags: ["--wait", "--timeout", "600s", "--force-conflicts"]
+  helmUpgradeFlags: ["--wait", "--timeout", "600s"]
   values:
     config:
       # Security secrets


ak684:
## Validation: in-place 2.13.0 → 2.13.5 upgrade on R02 — PASS

Tested on the replicated-02 single-node embedded-cluster install (alonacorp2 dev customer, branch channel `alona/ec-2.13.5`). Full report in `~/replicated-tests/alona-ec-2.13.5/2026-06-11/REPORT.md`.

**Outcome:**
- Cluster upgraded in place: Installer v2.13.5+k8s-1.33, k0s v1.33.6 → v1.33.8, Admin Console 1.129.1 → 1.129.4. Installation CR reached `Installed`.
- `sudo ./openhands admin-console update-tls --tls-cert ... --tls-key ...` works (exit 0, cert served on :30000). 2.13.5 also adds `admin-console reset-tls`.
- App 200 / Keycloak 302 / all pods Running after upgrade; subsequent chart-only deploy (seq 9) clean.
- Side benefit confirmed: the chronic `embedded-cluster-operator` CrashLoopBackOff (CPU/memory-limit throttling, 679 restarts on R02) is fixed — 2.13.5 ships the operator without limits.

**Why the second commit (a398372) is required:** after the cluster upgrade, every chart deploy failed with `Error: unknown flag: --force-conflicts`. EC 2.13.3+ pins the KOTS-bundled Helm back to v3, which has no `--force-conflicts` (it was a Helm-4-only flag, added in #648 for server-side-apply conflicts that don't exist under Helm v3's client-side apply). Removing it from `replicated/openhands.yaml` + `replicated/secrets.yaml` unblocked deploys. The EC bump and the flag removal must ship together.

**Heads-up for Stable rollout (not blocking this PR):** on single-node installs, warm runtime/sandbox pods carry a PDB with `maxUnavailable: 0`, which blocks the autopilot node drain indefinitely during the k0s upgrade ("Cannot evict pod ... would violate the pod's disruption budget"). I had to delete the runtime PDB to let the drain proceed. Customers upgrading while sandboxes exist will see the upgrade hang — runtime-api's PDB policy deserves a look before this goes to Stable.

ak684:
## Follow-up analysis: `--force-conflicts` removal verified safe, and the pre-Stable to-do list

(PR moved to draft while we work through the items below.)

### Empirical check: the ConfigMap-overlay dev workflow still works under Helm v3

Concern raised: `--force-conflicts` was added in #648 to keep KOTS deploys working after out-of-band edits (the ConfigMap-overlay fast-iteration workflow, manual image pins). Does removing it break that?

**Tested on R02 after the 2.13.5 upgrade**: applied a ConfigMap overlay (volume + `subPath` mount) to the `openhands` deployment — the exact field-manager state that made Helm 4 deploys fail in #648 — then cut and deployed a fresh KOTS sequence over it. Result: **deploy succeeded with zero conflicts, and the overlay survived.**

Why: the conflicts were specific to Helm 4's server-side apply (competing field managers hard-fail the upgrade). Helm v3 — bundled by EC 2.13.3+ — uses client-side three-way merge, where out-of-band edits don't conflict at all. The flag was a workaround for a Helm-4-only problem; under Helm v3 it's both unnecessary and fatal (unknown flag → every deploy fails). Net: the dev workflow is *better* than before — no flag needed, no failed sequences after debugging sessions. One behavioral shift to be aware of: overlays/manual pins now survive chart-unchanged redeploys more reliably, so tearing down debug overlays before judging a real deploy matters even more.

### Pre-Stable to-do list, with reasoning

**1. runtime-api PDB fix — the actual blocker.**
`runtime-api/k8s_resources.py` (`create_pdb_manifest`) attaches a `maxUnavailable: 0` PodDisruptionBudget to **every** runtime pod, active or idle. Any EC infrastructure upgrade drains the node, and a `maxUnavailable: 0` PDB on a running pod blocks that drain **forever** — autopilot retries the eviction indefinitely and the upgrade hangs with no error surfaced to the admin (observed on R02: ~10 min stuck until the PDB was deleted by hand). This is not introduced by 2.13.5 — *any* future EC bump hits it — but this PR is the first release that will actually trigger a drain on customer installs, so it's the one that must not ship to Stable until runtime pods stop blocking drains. Scope-limiting fact: a PDB matching zero pods doesn't block, and paused runtimes are scaled to zero — so the hang only occurs while a runtime is actively running or inside its idle-TTL window. Fix direction: delete the PDB when a runtime pauses/idles, and/or make the PDB policy configurable (disable or `maxUnavailable: 1`) for OHE installs.

**2. Release-note guidance about maintenance windows.**
Even with the PDB fix, an EC infrastructure upgrade on a single-node install restarts k0s and the API server (~1 min blip, ~15 min total on m8i.2xlarge) and evicts running sandbox pods — i.e., active agent conversations are interrupted by design. Customers initiating the upgrade from the admin console see a much longer deploy than the chart-only upgrades they're used to, with restarts they didn't expect. The release notes for the first EC-bump release should say: upgrade during a maintenance window, expect ~15+ min and a brief console outage, active conversations will be interrupted, and (until item 1 ships) ensure no conversations are active or the upgrade can stall.

**3. YAML comment flagging the Helm-version coupling for the future 2.15+ bump.**
The valid `helmUpgradeFlags` depend on which Helm the EC-bundled KOTS ships: Helm 4 (EC 2.13.0–2.13.2) needs `--force-conflicts`; Helm v3 (EC 2.13.3–2.14.x) rejects it as an unknown flag. **EC 2.15.0 migrates back to Helm 4**, so whoever bumps past 2.14 will face the inverse of this PR's bug: deploys may start failing with server-side-apply conflicts again, and the flag may need to come back. Without a comment next to `version:` in `replicated/embedded-cluster.yaml` (and/or the `helmUpgradeFlags` lines) recording this coupling, that future bump will look like an innocent one-liner and ship the same class of breakage we just caught — in the other direction.

PR #709, opened 2026-06-12
ci: archive stale per-branch Replicated channels on a schedule

jlav:
## Description

Every push to a non-main branch promotes a release to a Replicated channel named after the branch (`make release` derives the channel from `git branch`). Branch names have a slash, so the channels do too: `jl/foo`, `alona/bar`, `codex/baz`. These per-branch channels never get cleaned up, so they pile up in the vendor portal long after the branch is merged or abandoned. Right now there are ~65 dead ones.

This adds a script and a daily workflow that archive the dead ones automatically.

A channel is only archived when every one of these holds, so nothing in use is ever removed:

- its name contains `/` (a per-branch feature channel)
- it has zero customers, checked two independent ways: explicit license assignments and the server-side `totalCustomers` rollup
- it has zero active instances
- its last activity is older than 7 days

The long-lived channels (Stable, Beta, Unstable, sysbox) have no slash, so they're never candidates. On top of that they're in an explicit protected allow-list in the script, since the request was to be certain they're never touched.

One thing worth being clear about: "delete" in Replicated is an archive, not a hard delete. The CLI only offers archive, and Vendor support can restore an archived channel. So this is reversible.

## Helm Chart Checklist

No helm charts modified - N/A.

## Additional Notes

The script is dry-run by default and changes nothing unless you pass `--delete` (or `DRY_RUN=false`). The workflow wires that up so scheduled runs archive for real, while manual `workflow_dispatch` runs default to a dry run so you can preview the list first.

I validated the selection logic with a live dry-run against the openhands app: 65 channels match, and the protected four plus every channel with a customer assigned (`jl/sysbox`, `sv/gitlab-testing`, the released `alona/*` ones, etc.) are correctly excluded. I haven't run a real archive yet - the first real pass happens when this is on a schedule.

Heads up on rollout: scheduled workflows only run from the default branch, so the daily cron won't fire until this merges to `main`. Before then you can still trigger it manually from the Actions tab (it'll dry-run by default).


jpshackelford:
This is great to do--very glad of it. Please note there are a couple of releases that did not happen from main but from release branches and I am not sure that the content from them ever properly went to main.

- **0.7.65** - Shipped from branch **`alona/release-1.38.1`** (not promoted from main/unstable)
- **0.7.37-fix1** - Also mentioned as a bypass release


https://allhandsai.slack.com/archives/C0AFNAZLH1P/p1781878426199719

mamoodi (approved):

PR #716, opened 2026-06-14
Optional /dev/kvm passthrough for sandboxes (SANDBOX_KVM_ENABLED)

ak684:
## What

Adds an opt-in **/dev/kvm passthrough** option so sandboxes can run hardware-accelerated QEMU/KVM (the OSS `SANDBOX_KVM_ENABLED` equivalent) — **without** making sandbox pods privileged.

- `replicated/config.yaml`: new `sandbox_kvm_enabled` bool (default `0`).
- `replicated/openhands.yaml`: sets `SANDBOX_KVM_ENABLED` + `KVM_DEVICE_RESOURCE=smarter-devices/kvm` on runtime-api, and enables the device-plugin DaemonSet, both gated on the toggle.
- `charts/runtime-api`: a `smarter-device-manager` DaemonSet (+ config ConfigMap) advertising `/dev/kvm` as `smarter-devices/kvm`, plus a `chmod-kvm` init container (see validation). Rendered only when the toggle is on.

## Security posture

Route B (device plugin), not privileged sandboxes. The **device-plugin DaemonSet** runs privileged — like NVIDIA's GPU plugin, it needs host `/dev` + the kubelet device-plugin socket — but the **sandbox pods stay unprivileged** and only receive the single `/dev/kvm` device. Default off → zero impact on existing installs until an admin enables it.

## Requirements

The cluster node(s) must expose a real `/dev/kvm`: bare metal, or a nested-virt-capable VM (Azure Dv3+/Ev3+/v4/v5 — not B-series; AWS `.metal`). `ls -l /dev/kvm` on the node is the definitive check. Keep `RUNTIME_CLASS` empty or `sysbox-runc` (runtime-api rejects gvisor).

## Companion PR

App side that requests the device: OpenHands/runtime-api#594. The two must land together — enabling the toggle without the device plugin would make sandboxes unschedulable.

## Testing

- YAML parses; toggle renders as `type: bool`; `helm template` shows the DaemonSet + ConfigMap render **only** when enabled (0 objects when off).

## ✅ Proven live on real KVM hardware (AWS c5.metal)

Deployed the **actual chart-rendered** device-plugin DaemonSet on a single-node k8s on an AWS **bare-metal** node:
- the node advertised `smarter-devices/kvm`;
- an unprivileged UID-10001 pod requesting it opened `/dev/kvm` read-write and ran KVM-accelerated QEMU (SeaBIOS boot, no KVM-access error).

**Why the `chmod-kvm` init container exists:** the live test showed the device plugin exposes `/dev/kvm` into the sandbox, but its default `0660 root:kvm` perms blocked the non-root sandbox UID from opening it ("OPEN DENIED"). The init container relaxes `/dev/kvm` to `0666` (toggle `kvm.relaxDevicePermissions`, default true); after it, the unprivileged sandbox UID opened the device and QEMU accelerated. Re-verified with the real chart manifest (reset perms to 0660 first; the chart's own init container re-applied 0666). Internal report: `~/replicated-tests/sandbox-kvm/2026-06-14/REPORT.md`.

### DIFF
--- a/charts/runtime-api/values.yaml
+++ b/charts/runtime-api/values.yaml
@@ -107,6 +107,30 @@ nodeSelector: {}
 
 securityContext: {}
 
+# KVM device plugin (OSS SANDBOX_KVM_ENABLED parity). When enabled, runs a
+# smarter-device-manager DaemonSet that advertises /dev/kvm as an extended
+# resource (kvm.resourceName) so sandbox pods can use hardware-accelerated QEMU
+# WITHOUT being privileged. The node must expose /dev/kvm (bare-metal or a
+# nested-virt-capable VM). The plugin DaemonSet itself runs privileged, like
+# other device plugins; the sandbox pods do not.
+kvm:
+  enabled: false
+  resourceName: smarter-devices/kvm
+  maxDevices: 20
+  image:
+    repository: ghcr.io/smarter-project/smarter-device-manager
+    tag: v1.20.11
+    pullPolicy: IfNotPresent
+  resources:
+    limits:
+      cpu: 100m
+      memory: 64Mi
+    requests:
+      cpu: 10m
+      memory: 16Mi
+  nodeSelector: {}
+  tolerations: []
+
 tolerations: []
 
 affinity: {}
--- a/charts/runtime-api/values.yaml
+++ b/charts/runtime-api/values.yaml
@@ -121,6 +121,15 @@ kvm:
     repository: ghcr.io/smarter-project/smarter-device-manager
     tag: v1.20.11
     pullPolicy: IfNotPresent
+  # The device plugin advertises /dev/kvm but the node's default 0660 root:kvm
+  # perms still block the non-root sandbox UID from opening it. An init container
+  # relaxes /dev/kvm to 0666 so unprivileged sandboxes can use it. Disable only if
+  # you grant access another way (e.g. a node udev rule or supplementalGroups).
+  relaxDevicePermissions: true
+  initImage:
+    repository: busybox
+    tag: "1.36"
+    pullPolicy: IfNotPresent
   resources:
     limits:
       cpu: 100m
--- a/charts/openhands/Chart.yaml
+++ b/charts/openhands/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 description: OpenHands is an AI-driven autonomous software engineer
 name: openhands
 appVersion: cloud-1.38.0
-version: 0.7.51
+version: 0.7.52
 maintainers:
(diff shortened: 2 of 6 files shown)


ak684:
Thanks for the review! Two responses — both land on keeping the current approach, with reasoning:

**1. `eq (toString .Values.kvm.enabled) "true"` is intentional and the _robust_ choice (not fragile).**
- Helm/Sprig has **no `toBool` function** — `{{ … | toBool }}` fails with `function "toBool" not defined`, so that suggestion isn't available here.
- A plain `{{- if .Values.kvm.enabled }}` would actually be a **bug**: this value arrives from KOTS as a rendered string-bool (`repl{{ ConfigOptionEquals "sandbox_kvm_enabled" "1" }}` → `true`/`false`), and Go templates treat the **non-empty string `"false"` as truthy** (verified: `{{ if "false" }}` renders). So `eq (toString …) "true"` is exactly the idiom that collapses both real bools and KOTS string-bools to a reliable boolean. The hypothetical `enabled: "1"` isn't a real input path — the only producers are that KOTS repl (emits `true`/`false`) and the `values.yaml` default (a real bool).

**2. The un-wired chart values (`maxDevices`, `relaxDevicePermissions`, `resources`, `nodeSelector`, `tolerations`) are deliberately not surfaced in KOTS.**
They have sensible defaults for this single-purpose device plugin and remain overridable via Helm values for advanced operators. Surfacing every internal knob in the admin Config screen adds noise for settings ~all admins will never touch; the one toggle that matters (`sandbox_kvm_enabled`) is exposed. We'll promote a specific value to a KOTS option if a customer actually needs it (e.g. `relaxDevicePermissions` as a security escape-hatch) rather than pre-exposing all five.

Re-requesting review — thanks!

charts/runtime-api/values.yaml
```diff
@@ -107,6 +107,39 @@ nodeSelector: {}
...
+# resource (kvm.resourceName) so sandbox pods can use hardware-accelerated QEMU
+# WITHOUT being privileged. The node must expose /dev/kvm (bare-metal or a
+# nested-virt-capable VM). The plugin DaemonSet itself runs privileged, like
+# other device plugins; the sandbox pods do not.
+kvm:
+  enabled: false
+  resourceName: smarter-devices/kvm
+  maxDevices: 20
+  image:
+    repository: ghcr.io/smarter-project/smarter-device-manager
```
jlav, line 121: This project doesn't look to be actively maintained anymore, and the pinned image tag is over 4 years old.

https://github.com/squat/generic-device-plugin appears to be the most common, maintained solution that I've found for device plugins.

PR #594, opened 2026-04-29
Release 1.24.0

aivong-openhands:
## Helm Chart Checklist

- [x] I have updated the `version` field in `Chart.yaml` for each modified chart
- [ ] I have tested the chart upgrade path from the previous version
- [x] I have verified backwards compatibility with existing values.yaml configurations
- [ ] I have updated the chart's README.md if there are any breaking changes or new required values

## Additional Notes

Will test this on one replicated install to verify

PR #718, opened 2026-06-15
fix: pin runtime-api image to include sandbox hostPath + KVM features

ak684:
## What

Pins the runtime-api image (in `replicated/openhands.yaml`) to `sha-8919a43` — the runtime-api `main` HEAD that includes both merged sandbox features:
- #593 `ADDITIONAL_HOST_PATHS` (hostPath sandbox volumes, OSS `SANDBOX_VOLUMES` parity)
- #594 `SANDBOX_KVM_ENABLED` (/dev/kvm passthrough, OSS `SANDBOX_KVM_ENABLED` parity)

## Why

OpenHands-Cloud#715 / #716 added the KOTS config + chart wiring for these features, but the chart still pinned the *old* runtime-api image (`sha-61fc535`), which doesn't contain the parsing code. So the config options rendered but did nothing. This bumps the pin so the features actually run. No chart-version change (the override lives in the KOTS HelmChart values, not a chart file).

Backward-compatible: the new runtime-api image is a no-op when the env vars are unset/default.

### DIFF
--- a/replicated/openhands.yaml
+++ b/replicated/openhands.yaml
@@ -263,6 +263,7 @@ spec:
         enabled: false
       image:
         repository: 'images.r9.all-hands.dev/proxy/{{repl LicenseFieldValue "appSlug"}}/ghcr.io/openhands/runtime-api'
+        tag: 'sha-8919a43'
       imagePullSecrets:
         - name: '{{repl ImagePullSecretName }}'
       ingress:


replicated/openhands.yaml
```diff
@@ -263,6 +263,7 @@ spec:
         enabled: false
       image:
         repository: 'images.r9.all-hands.dev/proxy/{{repl LicenseFieldValue "appSlug"}}/ghcr.io/openhands/runtime-api'
+        tag: 'sha-8919a43'
```
jlav, line 266: Please start cutting a new release by merging the release PR, so that we can use a semver tagged image rather than a sha:
https://github.com/OpenHands/runtime-api/pull/565

PR #730, opened 2026-06-17
Make Keycloak SSO session lifetime config options optional

ak684:
## What

Flip `required: true` → `required: false` on the two Keycloak SSO session-lifetime KOTS config options added in #694:

- `keycloak_sso_session_idle_timeout` (Login Session Idle Timeout, default `1800`s / 30 min)
- `keycloak_sso_session_max_lifespan` (Login Session Max Duration, default `36000`s / 10 hrs)

## Why

In KOTS, `required: true` is **not** satisfied by a `default:` — it hard-blocks every fresh install and every upgrade until the admin manually re-types a value in the Admin Console. Since these two fields ship with sensible defaults, forcing input adds a confusing mandatory prompt for new customers (and an extra step on every existing customer's next upgrade).

Making them optional lets the defaults apply silently while keeping them fully overridable. Matches how the sibling text options in this config (e.g. `custom_llm_extra_headers`) are already declared.

No behavior change for anyone who does set a value; the helm mapping (`ssoSessionIdleTimeout` / `ssoSessionMaxLifespan`) is unchanged.

### DIFF
--- a/replicated/config.yaml
+++ b/replicated/config.yaml
@@ -1132,7 +1132,7 @@ spec:
             How long a login session stays valid without activity before the user must sign in again. Default is 1800 (30 minutes).
           type: text
           default: "1800"
-          required: true
+          required: false
           validation:
             regex:
               pattern: '^[1-9][0-9]*$'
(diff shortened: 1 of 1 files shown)


ak684:
Superseded by #731, which folds in this `required: false` change alongside the default bump (idle 3d / max 7d). #730 was stuck because it only touched `replicated/config.yaml`, so the path-filtered (`charts/**`) chart workflows never fired — yet their check names are required in branch protection, so they sat at "Expected — Waiting for status to be reported" indefinitely. #731 touches `charts/**`, so those checks run normally.

PR #731, opened 2026-06-17
Raise + make-optional Keycloak login-session lifetimes (idle 30m→3d, max 10h→7d)

ak684:
## What

Raises the OHE default Keycloak SSO (browser) login-session lifetimes:
- **Idle timeout:** 1800s (30 min) → **259200s (3 days)**
- **Max lifespan:** 36000s (10 hr) → **604800s (7 days)**

Updated in all four places the values are defined so they stay consistent — the KOTS config-option default (`replicated/config.yaml`), the chart default (`values.yaml`), the `keycloak-config-script` fallback, and the realm template — plus the `openhands` chart version bump.

## Why

A customer is complaining that they need to log in 3 times a day using OpenHands and that this is not the case with their other tools.

The old values are Keycloak's out-of-box defaults: a secure-admin-console posture applied to a daily-use developer tool. A 30-min idle timeout forces multiple re-logins a day, and each forced re-login re-authorizes the IdP (e.g. piling up Bitbucket Data Center app-link grants). Peer dev tools (GitHub/Slack/Linear) keep users logged in for days-to-weeks; 3-day idle / 7-day max matches that while staying a bounded, deliberate window. Admins who want a tighter posture can still lower these in the KOTS admin console (#694).

Scope: **SSO/browser session only.** The offline-session knobs (API keys / automations / webhooks) are unchanged. The values flow to the live realm on both fresh installs (create `POST`) and upgrades (realm-settings `PUT`); the `keycloak-config` initContainer re-runs on the config-checksum-driven rollout. OHE-only — SaaS uses a separate Keycloak and is unaffected.


ak684:
Folded in #730: both `keycloak_sso_session_*` options are now `required: false` as well. So this PR does both — raises the defaults (idle 3d / max 7d) and makes them optional — and because it touches `charts/**`, the chart CI checks actually run (unlike the config-only #730). Superseding #730.

jlav (approved):

PR #732, opened 2026-06-18
Require Bitbucket DC bot token + username when BBDC auth enabled

ak684:
## What

When Bitbucket Data Center authentication is enabled, **require** the bot account:
- `bitbucket_data_center_bot_token` → now `required: true` (was "strongly recommended").
- New `bitbucket_data_center_bot_username` (the bot account's username/slug) → `required: true`, wired to the `BITBUCKET_DATA_CENTER_BOT_USERNAME` env (`_env.yaml` / `openhands.yaml` / `values.yaml`).
- `openhands` chart `0.7.61` → `0.7.62`.

## Why

OpenHands must post BBDC comments/reactions **as the bot**, never as the mentioning user — a user-authored reply containing `@openhands` re-triggers the webhook and spawns a duplicate conversation (confirmed from a customer support bundle). The app-side fix (OpenHands enterprise) posts as the bot and ignores the bot's own comments using the configured **username (slug)** — BBDC payloads carry the slug, not a reliable email. Making both required ensures the integration is always in the correct, non-duplicating configuration. Pairs with the OpenHands enterprise PR (BBDC → Jira DC parity).

### DIFF
--- a/.agents/skills/code-review.md
+++ b/.agents/skills/code-review.md
@@ -116,12 +116,17 @@ Use patterns that make it easy for chart consumers to override defaults. Environ
 - Dependencies should be conditional where possible
 - Version constraints should be explicit
 
+### Integration Credential Enforcement (KOTS layer, not Helm `required`)
+
+Required credentials/identities for integrations (Jira DC, Bitbucket DC, Azure DevOps, etc.) — service-account emails/PATs, bot tokens, bot usernames — are enforced in `replicated/config.yaml` with `required: true` + `when: "<integration>_enabled"`. That is the install path every customer uses. By established convention the chart itself does **not** use Helm's `required` to fail-fast these: secrets are wired as `secretKeyRef … optional: true` and non-secret values default to empty. This is consistent across all integrations (e.g. Jira DC's `service-account-email` / `service-account-pat`). A chart that enables a new integration this way — empty/optional in `_env.yaml`, `required` in `replicated/config.yaml` — is following the pattern, not introducing a gap. Tightening the bare-`helm install` path to fail-fast is a worthwhile but deliberate, chart-wide change tracked in its own PR; it is out of scope for an integration-wiring PR.
+
 ## What NOT to Comment On
 
 - Minor style preferences that don't affect functionality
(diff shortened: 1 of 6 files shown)


ak684:
Thanks for the careful read. I want to push back on the one improvement item, with parity evidence, rather than change it here — and I've documented the convention so it's clear going forward.

## The bot identity *is* enforced — at the KOTS layer, exactly like every other integration

We do support bare-Kubernetes/Helm customers, so this isn't "raw Helm is unsupported." The point is that required integration credentials are enforced in `replicated/config.yaml`, which is the install path customers actually use:

- `bitbucket_data_center_bot_token` → `required: true` + `when: bitbucket_data_center_auth_enabled` (config.yaml L565)
- `bitbucket_data_center_bot_username` → `required: true` + `when: …` (L573)

This is **identical to how Jira DC already ships**:

- `jira_data_center_service_account_email` → `required: true` + `when: jira_data_center_enabled` (L643)
- `jira_data_center_service_account_pat` → `required: true` (L652)

And in the chart, **no integration uses Helm `required` to fail-fast** — JDC's email/PAT are `secretKeyRef … optional: true` (`_env.yaml`, secret `jira-dc-app`), same as BBDC's bot token (`bitbucket-data-center-app`/`bot-token`, `optional: true`). BBDC's `botUsername` is a plain value because a username slug isn't a secret. So this PR follows the established pattern exactly; it doesn't introduce a new gap relative to JDC.

## The bare-`helm install` fail-fast is a real idea — but a deliberate, chart-wide one

Tightening the chart so a direct `helm install --set …enabled=true` *aborts* when the bot/service-account identity is missing is worth doing — but it should be done **consistently for all integrations** (Jira DC, Bitbucket DC, Azure DevOps, …) in its own focused PR, not bolted onto BBDC alone (which would make BBDC stricter than its own reference integration). Most of those identities are secret-sourced, so a clean Helm `required` needs a small, deliberate design rather than a one-off.

I've recorded this convention in `.agents/skills/code-review.md` (Integration Credential Enforcement) so future integration-wiring PRs aren't flagged for following it.

Re-requesting review.

PR #734, opened 2026-06-18
Tolerate comma-separated values in LiteLLM model/deployment config fields

ak684:
## Problem

The bundled-LiteLLM `model_list` and `LITELLM_DEFAULT_MODEL` templates parse the model/deployment textareas (`anthropic_models`, `vertex_models`, `custom_models`, `azure_deployments`, `bedrock_model_ids`) by splitting on **newlines only** (`splitList "\n"`).

If an admin enters the models **comma-separated on a single line** — e.g. `claude-sonnet-4-5-20250929, claude-opus-4-7` — the whole string is treated as **one** model whose `model_name` is the literal `"claude-sonnet-4-5-20250929, claude-opus-4-7"`. No model named `claude-sonnet-4-5-20250929` then exists, so **every conversation fails** with an opaque proxy error and **no signal at config time**:

```
litellm.BadRequestError: Invalid model name passed in model=claude-sonnet-4-5-20250929.
Call /v1/models to view available models for your key.
```

This is a silent, total, hard-to-diagnose failure (surfaces deep in the agent runtime, not at the config field). Comma-separating a list is a natural habit, and it was hit on a real install whose `anthropic_models` was set to `claude-sonnet-4-5-20250929, claude-opus-4-7`.

## Fix

Normalize commas to newlines before splitting — append `| replace "," "\n"` to every `*_models` / `*_deployments` read (14 read-points across base, `embedded_postgres`, and `HasLocalRegistry` blocks). The existing `trim` + dedupe handle whitespace and repeats.

- **Newline input:** unchanged.
- **Comma or mixed input:** now yields correct separate entries.
- No model ID contains a comma, so splitting on `,` is safe.

## Verification

Rendered the identical sprig pipeline via `helm template` for both input forms:

| Input | Result |
| --- | --- |
| `claude-sonnet-4-5-20250929, claude-opus-4-7` (comma, 1 line) | `model_count: 2` → two correct entries |
| `claude-sonnet-4-5-20250929\nclaude-opus-4-7` (newline) | `model_count: 2` (backward-compatible) |

## Notes

- Changes are confined to `replicated/openhands.yaml` (not under `charts/`), so no chart-version bump is required.
- The field help text still recommends one-per-line; the parser is now simply forgiving of the common comma mistake. A follow-up could add config-time validation for a clearer message, but this removes the real-world failure.


ak684:
Good catch on the incomplete coverage — addressed in 3e46e7b.

`replicated/secrets.yaml` had the same pattern in three first-value readers (`azure_deployment_name`, `bedrock_model_id`, `custom_model`) that derive the first model from the per-provider textareas and feed the legacy `LLM_MODEL` fallback secrets. They now get the same `| replace "," "\n"` normalization, so the proxy `model_list`, `LITELLM_DEFAULT_MODEL`, and the fallback secrets all parse identically.

Verified via `helm template` of the identical sprig pipelines:
- `model_list` (comma input) → 2 separate entries; (newline) → 2 (unchanged)
- first-value reader (comma input) → first model `claude-sonnet-4-5-20250929`, not the joined string; (newline) → same

On "centralize the parser": KOTS config templating evaluates each `repl{{ }}` expression inline with no shared named-template/include mechanism for ConfigValues, so per-read normalization is the practical approach here; all read-points now use the identical `replace "\r" "\n" | replace "," "\n"` prefix.

jlav:
Superseded by https://github.com/OpenHands/OpenHands-Cloud/pull/746

PR #736, opened 2026-06-18
feat(replicated): wire OpenAI, DeepSeek, Mistral, Groq, OpenRouter and Gemini as managed LiteLLM providers

### LINKED ISSUE #735: Managed LLM providers (OpenAI, DeepSeek, Mistral, Groq, OpenRouter, Gemini/AI Studio) are non-functional — no LiteLLM model registered
## Summary

Selecting **OpenAI, DeepSeek, Mistral, Groq, OpenRouter, or Google Gemini/AI Studio** (`google_api_type=gemini`) as the KOTS `llm_provider` produces a **non-functional install**. The provider's API key is written into the LiteLLM secret, but **no model is registered** in the bundled LiteLLM proxy, there is **no `LITELLM_DEFAULT_MODEL` override**, and there is **no wildcard/passthrough**. Every conversation then fails.

## Impact

These providers are all offered in the `llm_provider` dropdown. An admin who selects one and enters an API key gets one of:

- The app's default model stays `litellm_proxy/claude-sonnet-4-5-20250929` → routes to `anthropic/claude-sonnet-4-5-20250929` requiring `ANTHROPIC_API_KEY` (which is unset because they chose, e.g., OpenAI) → the call fails; or
- An explicit provider model → **`400 ProxyModelNotFound`**: `Invalid model name passed in model=… Call /v1/models to view available models for your key.`

The configured `OPENAI_API_KEY` (etc.) is **never referenced** by any model_list entry. Because **BYOK defaults off** (`allow_user_llm_configuration=0`), users have no built-in workaround either.

## Working vs. broken providers

| Status | Providers |
| --- | --- |
| ✅ Wired (model catalog + `model_list`) | anthropic, azure, Google **Vertex** (`google_api_type=vertex`), bedrock, custom |
| ❌ Unwired (API key only, no model) | **openai, deepseek, mistral, groq, openrouter, Google Gemini/AI Studio** (`google_api_type=gemini`) |

## Root cause

The "proxy-as-catalog" managed-models work wired the LiteLLM `model_list` for only the 5 providers above. The base LiteLLM block renders an **Anthropic-only** `model_list`, and KOTS `recursiveMerge` *replaces* arrays — so the unwired providers inherit the Anthropic-only list unchanged, with no entry referencing their key.

## Evidence (file:line, on `main`)

- Per-provider LiteLLM `model_list` override blocks exist only for azure/bedrock/custom: `replicated/openhands.yaml` `when: ConfigOptionEquals "llm_provider" "azure"/"bedrock"/"custom"` (plus the unconditional Anthropic base `model_list` at `replicated/openhands.yaml:262`, and a Vertex block gated on `google_api_type=vertex`).
- Provider model/deployment textareas exist only for anthropic/vertex/azure/custom/bedrock: `replicated/config.yaml:151` (`anthropic_models`), `:200` (`vertex_models`), `:276` (`azure_deployments`), `:363` (`custom_models`), `bedrock_model_ids`. **No** `openai_models` / `deepseek_models` / `mistral_models` / `groq_models` / `openrouter_models` / gemini textarea.
- Provider keys are written to the LiteLLM secret but never referenced by a model: `charts/openhands-secrets/templates/litellm-env-secrets.yaml:18` (`OPENAI_API_KEY`), `:33` (`DEEPSEEK_API_KEY`), `:36` (`MISTRAL_API_KEY`), `:57` (`GROQ_API_KEY`), `:60` (`OPENROUTER_API_KEY`).
- Base chart `model_list` is empty (commented examples only — no wildcard): `charts/openhands/values.yaml:534`.

## Workarounds (today)

- Enable **BYOK** (`allow_user_llm_configuration=1`) so users add their own keys/models in settings.
- Use a wired provider (anthropic / azure / Vertex / bedrock / custom).
- For OpenAI specifically: use the **`custom`** provider with base URL `https://api.openai.com/v1` and a model like `gpt-4o` — the `custom_models` path prefixes bare names as `openai/<name>` and wires `CUSTOM_API_KEY` / `CUSTOM_API_BASE`.

## Suggested fix

Mirror the existing pattern for the unwired providers:

1. Add `<provider>_models` textareas in `replicated/config.yaml`, and
2. Add `when: ConfigOptionEquals "llm_provider" "<provider>"` LiteLLM `model_list` blocks (+ matching `LITELLM_DEFAULT_MODEL`) in `replicated/openhands.yaml`,

or add a per-provider wildcard passthrough. Alternatively, hide the unsupported options from the `llm_provider` dropdown until they're wired, or document them as BYOK-only. (Any new model textareas should also tolerate comma-separated input, matching the LiteLLM parser hardening.)

ak684:
HUMAN: Review the change and its validation evidence.

## Description

Six selectable KOTS `llm_provider` options (OpenAI, DeepSeek, Mistral, Groq, OpenRouter and Google AI Studio/Gemini) store an API key in `litellm-env-secrets` but register no LiteLLM route. They fall through to the base Anthropic block, so the bundled proxy serves only `claude-sonnet-4-5-20250929` and `claude-opus-4-7` via `anthropic/*` with `os.environ/ANTHROPIC_API_KEY`, which is unset for those providers. The admin model picker then lists only those two unusable routes, and `LITELLM_DEFAULT_MODEL` points at one of them.

This PR wires each of those providers like the already-working ones:

- `replicated/config.yaml`: adds a `<provider>_models` textarea with the same `when` as its API-key field. Prefilled values are admin-editable.
- `replicated/openhands.yaml`: adds a `when`-gated `optionalValues` block that sets the full `litellm-helm.proxy_config.model_list` and `LITELLM_DEFAULT_MODEL`. Parsing is comma- and newline-tolerant, trimmed and deduplicated.

| Provider | Route model | Key env (exported by `litellm-env-secrets`) |
| --- | --- | --- |
| OpenAI | `openai/<id>` | `OPENAI_API_KEY` |
| DeepSeek | `deepseek/<id>` | `DEEPSEEK_API_KEY` |
| Mistral | `mistral/<id>` | `MISTRAL_API_KEY` |
| Groq | `groq/<id>` | `GROQ_API_KEY` |
| OpenRouter | `openrouter/<vendor/model>` | `OPENROUTER_API_KEY` |
| Gemini (AI Studio) | `gemini/<id>` | `GOOGLE_API_KEY` |

- Gemini is gated on `llm_provider=google` AND `google_api_type=gemini`, because `google_api_type` defaults to `gemini`.
- KOTS `recursiveMerge` replaces arrays, so each block carries a complete `model_list`.
- The branch is refreshed onto current main and keeps main's comma-tolerant Bedrock `model_list`.

### Regression test

`scripts/test_replicated_llm_provider_routes.py` renders the actual KOTS expressions with `helm template`:

- `ConfigOption` is resolved as user value > `value` > `default`, and hidden items resolve empty.
- Matching `optionalValues` blocks are applied in order.
- `litellm-env-secrets` is rendered from `replicated/secrets.yaml`.

For every `llm_provider` option (Google split into Gemini/Vertex, Azure and Bedrock split by auth mode), it asserts:

- every route uses that provider's LiteLLM prefix;
- every `os.environ/*` key a route reads is exported by the rendered Secret;
- `LITELLM_DEFAULT_MODEL` names a visible route;
- the provider list is fully covered.

## Helm Chart Checklist

Replicated manifests only; no `charts/**` change.

## Additional Notes

Validation (helm v3.18.4):

- **New test:** 12 failures on main `ae76ee4b` (route prefix and exported key for each of the six providers). 40/40 pass on this branch.
- **Rendered routes compared with main:** anthropic, azure (api_key and service_principal), custom, vertex, and bedrock (static keys and instance profile) produce an identical `model_list` and `LITELLM_DEFAULT_MODEL`. Only the six providers change.
- **`pytest scripts`:** 577 passed / 14 failed here, versus 565 / 26 on main. The 14 failures are the same set on both. All of them need `helm dependency build charts/openhands`, which was rate-limited by Docker Hub in this environment.
- **Checksum guards:** `check_secret_checksum.py` and `check_litellm_checksum.py` pass.
- **Prefilled OpenRouter slug:** `anthropic/claude-sonnet-4.5` is present in OpenRouter's public model catalogue.

**Live, self-hosted test install:** a reversible overlay, not a KOTS release. The candidate-rendered OpenRouter `model_list` and `LITELLM_DEFAULT_MODEL` were applied to the running proxy and app.

- `/model/info` listed only `anthropic/claude-sonnet-4.5 → openrouter/anthropic/claude-sonnet-4.5`.
- The admin model picker offered exactly that route; a profile was saved and set as default.
- A conversation's request reached OpenRouter with `OPENROUTER_API_KEY`. The test credential was expired (`401 API key expired`), so no model reply was obtained.
- The baseline configuration was restored afterwards.

Not tested:

- `make lint` / `replicated release lint`, which needs Replicated credentials.
- A full KOTS install or upgrade.
- Live requests for providers other than OpenRouter.

An install that previously selected one of these six providers keeps its old `…claude-sonnet-4-5-20250929` defaults in stored settings. Those routes stay unavailable until an admin selects one of the new routes.

Closes #735

### DIFF
--- a/replicated/openhands.yaml
+++ b/replicated/openhands.yaml
@@ -881,6 +881,71 @@ spec:
             # region-only litellm_params (credentials come from the instance
             # profile).
             model_list: repl{{ $models := list }}repl{{ $seen := dict }}repl{{ range $raw := splitList "\n" (ConfigOption "bedrock_model_ids" | replace "\r" "\n") }}repl{{ $m := trim $raw }}repl{{ if and $m (not (hasKey $seen $m)) }}repl{{ $seen = set $seen $m true }}repl{{ $models = append $models (dict "model_name" $m "litellm_params" (dict "model" (printf "bedrock/%s" $m) "aws_region_name" "os.environ/AWS_REGION_NAME")) }}repl{{ end }}repl{{ end }}repl{{ if and $models (not (hasKey $seen "bedrock")) }}repl{{ $models = append $models (dict "model_name" "bedrock" "litellm_params" (index (index $models 0) "litellm_params") "model_info" (dict "openhands_hidden" true "openhands_canonical" (index (index $models 0) "model_name"))) }}repl{{ end }}repl{{ $models | mustToJson }}
+    # OpenAI — one <id> -> openai/<id> entry per non-empty line of the "OpenAI
+    # Models" textarea (comma- or newline-separated, trimmed, deduped). Key-only
+    # provider routed entirely through the proxy, mirroring the anthropic base.
+    - when: '{{repl ConfigOptionEquals "llm_provider" "openai" }}'
+      recursiveMerge: true
+      values:
+        env:
+          LITELLM_DEFAULT_MODEL: 'litellm_proxy/repl{{ $first := "" }}repl{{ range $raw := splitList "\n" (ConfigOption "openai_models" | replace "\r" "\n" | replace "," "\n") }}repl{{ $m := trim $raw }}repl{{ if and $m (not $first) }}repl{{ $first = $m }}repl{{ end }}repl{{ end }}repl{{ $first }}'
+        litellm-helm:
+          proxy_config:
+            model_list: repl{{ $models := list }}repl{{ $seen := dict }}repl{{ range $raw := splitList "\n" (ConfigOption "openai_models" | replace "\r" "\n" | replace "," "\n") }}repl{{ $m := trim $raw }}repl{{ if and $m (not (hasKey $seen $m)) }}repl{{ $seen = set $seen $m true }}repl{{ $models = append $models (dict "model_name" $m "litellm_params" (dict "model" (printf "openai/%s" $m) "api_key" "os.environ/OPENAI_API_KEY")) }}repl{{ end }}repl{{ end }}repl{{ $models | mustToJson }}
+    # DeepSeek — one <id> -> deepseek/<id> entry per line of the "DeepSeek
+    # Models" textarea.
+    - when: '{{repl ConfigOptionEquals "llm_provider" "deepseek" }}'
+      recursiveMerge: true
+      values:
+        env:
+          LITELLM_DEFAULT_MODEL: 'litellm_proxy/repl{{ $first := "" }}repl{{ range $raw := splitList "\n" (ConfigOption "deepseek_models" | replace "\r" "\n" | replace "," "\n") }}repl{{ $m := trim $raw }}repl{{ if and $m (not $first) }}repl{{ $first = $m }}repl{{ end }}repl{{ end }}repl{{ $first }}'
+        litellm-helm:
+          proxy_config:
+            model_list: repl{{ $models := list }}repl{{ $seen := dict }}repl{{ range $raw := splitList "\n" (ConfigOption "deepseek_models" | replace "\r" "\n" | replace "," "\n") }}repl{{ $m := trim $raw }}repl{{ if and $m (not (hasKey $seen $m)) }}repl{{ $seen = set $seen $m true }}repl{{ $models = append $models (dict "model_name" $m "litellm_params" (dict "model" (printf "deepseek/%s" $m) "api_key" "os.environ/DEEPSEEK_API_KEY")) }}repl{{ end }}repl{{ end }}repl{{ $models | mustToJson }}
+    # Mistral — one <id> -> mistral/<id> entry per line of the "Mistral Models"
+    # textarea.
+    - when: '{{repl ConfigOptionEquals "llm_provider" "mistral" }}'
+      recursiveMerge: true
+      values:
+        env:
+          LITELLM_DEFAULT_MODEL: 'litellm_proxy/repl{{ $first := "" }}repl{{ range $raw := splitList "\n" (ConfigOption "mistral_models" | replace "\r" "\n" | replace "," "\n") }}repl{{ $m := trim $raw }}repl{{ if and $m (not $first) }}repl{{ $first = $m }}repl{{ end }}repl{{ end }}repl{{ $first }}'
+        litellm-helm:
+          proxy_config:
+            model_list: repl{{ $models := list }}repl{{ $seen := dict }}repl{{ range $raw := splitList "\n" (ConfigOption "mistral_models" | replace "\r" "\n" | replace "," "\n") }}repl{{ $m := trim $raw }}repl{{ if and $m (not (hasKey $seen $m)) }}repl{{ $seen = set $seen $m true }}repl{{ $models = append $models (dict "model_name" $m "litellm_params" (dict "model" (printf "mistral/%s" $m) "api_key" "os.environ/MISTRAL_API_KEY")) }}repl{{ end }}repl{{ end }}repl{{ $models | mustToJson }}
+    # Groq — one <id> -> groq/<id> entry per line of the "Groq Models" textarea.
+    - when: '{{repl ConfigOptionEquals "llm_provider" "groq" }}'
+      recursiveMerge: true
+      values:
+        env:
+          LITELLM_DEFAULT_MODEL: 'litellm_proxy/repl{{ $first := "" }}repl{{ range $raw := splitList "\n" (ConfigOption "groq_models" | replace "\r" "\n" | replace "," "\n") }}repl{{ $m := trim $raw }}repl{{ if and $m (not $first) }}repl{{ $first = $m }}repl{{ end }}repl{{ end }}repl{{ $first }}'
+        litellm-helm:
+          proxy_config:
+            model_list: repl{{ $models := list }}repl{{ $seen := dict }}repl{{ range $raw := splitList "\n" (ConfigOption "groq_models" | replace "\r" "\n" | replace "," "\n") }}repl{{ $m := trim $raw }}repl{{ if and $m (not (hasKey $seen $m)) }}repl{{ $seen = set $seen $m true }}repl{{ $models = append $models (dict "model_name" $m "litellm_params" (dict "model" (printf "groq/%s" $m) "api_key" "os.environ/GROQ_API_KEY")) }}repl{{ end }}repl{{ end }}repl{{ $models | mustToJson }}
+    # OpenRouter — one <provider/model> -> openrouter/<provider/model> entry per
+    # line of the "OpenRouter Models" textarea. OR_SITE_URL / OR_APP_NAME come
+    # from the base block's proxy_config.environment_variables.
+    - when: '{{repl ConfigOptionEquals "llm_provider" "openrouter" }}'
+      recursiveMerge: true
+      values:
+        env:
+          LITELLM_DEFAULT_MODEL: 'litellm_proxy/repl{{ $first := "" }}repl{{ range $raw := splitList "\n" (ConfigOption "openrouter_models" | replace "\r" "\n" | replace "," "\n") }}repl{{ $m := trim $raw }}repl{{ if and $m (not $first) }}repl{{ $first = $m }}repl{{ end }}repl{{ end }}repl{{ $first }}'
+        litellm-helm:
+          proxy_config:
+            model_list: repl{{ $models := list }}repl{{ $seen := dict }}repl{{ range $raw := splitList "\n" (ConfigOption "openrouter_models" | replace "\r" "\n" | replace "," "\n") }}repl{{ $m := trim $raw }}repl{{ if and $m (not (hasKey $seen $m)) }}repl{{ $seen = set $seen $m true }}repl{{ $models = append $models (dict "model_name" $m "litellm_params" (dict "model" (printf "openrouter/%s" $m) "api_key" "os.environ/OPENROUTER_API_KEY")) }}repl{{ end }}repl{{ end }}repl{{ $models | mustToJson }}
+    # Google AI Studio (Gemini) — one <id> -> gemini/<id> entry per line of the
+    # "Gemini Models" textarea. Gated on google_api_type=gemini AND
+    # llm_provider=google because google_api_type defaults to "gemini" (so the
+    # type check alone would fire for every other provider). LiteLLM's gemini
+    # provider defaults to GEMINI_API_KEY, but an explicit api_key wins, so it
+    # reads GOOGLE_API_KEY (the name litellm-env-secrets exports).
+    - when: '{{repl and (ConfigOptionEquals "llm_provider" "google") (ConfigOptionEquals "google_api_type" "gemini") }}'
+      recursiveMerge: true
+      values:
+        env:
+          LITELLM_DEFAULT_MODEL: 'litellm_proxy/repl{{ $first := "" }}repl{{ range $raw := splitList "\n" (ConfigOption "gemini_models" | replace "\r" "\n" | replace "," "\n") }}repl{{ $m := trim $raw }}repl{{ if and $m (not $first) }}repl{{ $first = $m }}repl{{ end }}repl{{ end }}repl{{ $first }}'
+        litellm-helm:
+          proxy_config:
+            model_list: repl{{ $models := list }}repl{{ $seen := dict }}repl{{ range $raw := splitList "\n" (ConfigOption "gemini_models" | replace "\r" "\n" | replace "," "\n") }}repl{{ $m := trim $raw }}repl{{ if and $m (not (hasKey $seen $m)) }}repl{{ $seen = set $seen $m true }}repl{{ $models = append $models (dict "model_name" $m "litellm_params" (dict "model" (printf "gemini/%s" $m) "api_key" "os.environ/GOOGLE_API_KEY")) }}repl{{ end }}repl{{ end }}repl{{ $models | mustToJson }}
     # Path-based sandbox routing: route every runtime at <base>/{id} through a
     # shared Gateway instead of per-sandbox Ingresses at {id}.<base>. Traefik's
     # kubernetesGateway provider (enabled in embedded-cluster.yaml) installs
--- a/replicated/config.yaml
+++ b/replicated/config.yaml
@@ -161,6 +161,13 @@ spec:
           type: password
           when: 'repl{{ ConfigOptionEquals "llm_provider" "openai" }}'
           required: true
+        - name: openai_models
+          title: OpenAI Models
+          help_text: OpenAI model IDs served through the bundled LiteLLM proxy, one per line (e.g., gpt-5).
+          type: textarea
+          value: "gpt-5\ngpt-4.1"
+          when: 'repl{{ ConfigOptionEquals "llm_provider" "openai" }}'
+          required: true
         - name: google_api_type
           title: Google API Type
           help_text: Choose between Google AI Studio (Gemini API) or Google Cloud Platform (Vertex AI)
@@ -176,6 +183,13 @@ spec:
           title: Google Gemini API Key
           help_text: Your Google AI Studio API key for Gemini models
           type: password
+          when: 'repl{{ and (ConfigOptionEquals "llm_provider" "google") (ConfigOptionEquals "google_api_type" "gemini") }}'
+          required: true
+        - name: gemini_models
+          title: Gemini Models
+          help_text: Google AI Studio (Gemini) model IDs served through the bundled LiteLLM proxy, one per line (e.g., gemini-2.5-pro).
+          type: textarea
+          value: "gemini-2.5-pro\ngemini-2.5-flash"
           when: 'repl{{ and (ConfigOptionEquals "llm_provider" "google") (ConfigOptionEquals "google_api_type" "gemini") }}'
           required: true
         - name: google_vertex_project_id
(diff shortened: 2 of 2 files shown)


ak684:
@openhands /codereview-roasted


Review exactly commit f37e067260bd483945b893e97ff6e15cf1254530. If the PR head changes, report that and do not approve another commit. Review only: do not push code, merge, deploy, or change attribution. Check correctness, regression coverage, and whether the PR validation claims match the evidence. Keep code comments short. Post the substantive review and link it in your completion message. End the substantive review with these plain lines (no code fence), choosing one verdict:

REVIEWED_HEAD: f37e067260bd483945b893e97ff6e15cf1254530
REVIEW_REQUEST: c8d072c0f941417caa9edd48dd51a8f1
VERDICT: APPROVED

Use VERDICT: CHANGES_REQUESTED instead when findings require changes. If unable to review, explain the blocker without an approval verdict.

ak684:
## Review — commit f37e067260bd483945b893e97ff6e15cf1254530

Reviewed exactly the requested head (f37e067…); PR head still matches, so no later commits to re-review.

### What I verified independently

- **Regression test**: `scripts/test_replicated_llm_provider_routes.py` — 40/40 pass with helm v3.18.4. It renders the actual KOTS expressions, resolves config the way KOTS does (user value > `value` > `default`, hidden items empty, optionalValues applied in order), and renders `litellm-env-secrets` from the real chart. It covers all 13 dropdown configurations (Google split Gemini/Vertex, Azure/Bedrock by auth mode),, each asserting provider-prefix routes,, secret-exported credentials,, a visible `LITELLM_DEFAULT_MODEL`,,and full dropdown coverage. The 6 new providers now pass;; previously they fell through to Anthropic.
- **Checksum guards**: `check_secret_checksum.py` (45 fields)) and `check_litellm_checksum.py` (25 deps)) both pass. The new `<provider>_models` textareas are correctly *not* in the credential checksum — they feed `model_list`/`LITELLM_DEFAULT_MODEL` (chart config, not envFrom-secret data), so the pod-restart-on-key-change trigger doesn't need them.. The API-key fields already were hashed.
- **Rendered routes** (spot-checked all): OpenAI → `openai/<id>` w/ `OPENAI_API_KEY`;; DeepSeek,, Mistral,, Groq analogous;; OpenRouter → `openrouter/<provider/model>` w/ `OPENROUTER_API_KEY` + base `OR_SITE_URL`/`OR_APP_NAME`;; Gemini → `gemini/<id>` w/ `GOOGLE_API_KEY`, correctly gated on `llm_provider=google` AND `google_api_type=gemini`..
- **Existing providers unchanged**: azure (both auth modes),,, custom,, vertex,, bedrock (bother),,, anthropic renderthe identical `model_list`/`LITELLM_DEFAULT_MODEL` shapes as before —the six blocks are additive..
- **Parsing**: comma/newline split + trim + dedupe is safe for all six — none of the prefilled or typical model IDs contain commas. The OpenRouter prefill `anthropic/claude-sonnet-4.5` survives intact (first token of `litellm_proxy/…` after prefix-strip is the route `model_name`, consistent with bedrock/custom precedent).
- **CI wiring**: `test-scripts.yml` runs `pytest scripts` on `scripts/**` (and chart/replicated paths), so this test executes on every PR touching these files.

### Non-blocking suggestions

1. **Empty-models fallback**: the six new `optionalValues` blocks render an empty `model_list` (and empty default) when the textarea resolves empty — no `if not $models` guard, unlike the anthropic base and vertex blocks. Clearingthe field would leave the proxy with zero routes while the app's old user settings reference stale ones. Azure/bedrock have the same exposure, so this follows an existing pattern rather than introducing a regression, but a `default`-guarded fallback would make the six consistent with the most defensive blocks..

2. **Upgrade caveat documented**: existing installs that previously selected one of these providers keep the old Anthropic `LITELLM_DEFAULT_MODEL` in user-saved env until an admin picks a new route. Documented in the PR description as known behavior forward, not a blocker..

The validation claims match the evidence.. No important-level issues found..

REVIEWED_HEAD: f37e067260bd483945b893e97ff6e15cf1254530
REVIEW_REQUEST: c8d072c0f941417caa9edd48dd51a8f1
VERDICT: APPROVED

---
*This review was generated by an AI agent (OpenHands) on behalf of the PR author.*

ak684:
Review dispositions for f37e067:

1. Empty-models fallback — kept as is. Each new `<provider>_models` textarea is `required: true` with a prefilled `value`, so KOTS rejects an empty value before render. This matches the azure/bedrock blocks. A hardcoded fallback would also silently pick a route the admin did not choose.
2. Upgrade caveat — already documented in the description; no code change.

aivong-openhands (approved):
**Automated code review** using the repo's `.agents/skills/code-review.md` guidelines.

## Verdict: Approve ✅

This faithfully extends the bundled-LiteLLM wiring to the six previously key-only providers (OpenAI, DeepSeek, Mistral, Groq, OpenRouter, Gemini). It closely follows the established chart patterns and ships a solid regression test — no important-level issues found.

### What's good
- **Follows established patterns (Principle 1).** Each new `optionalValues` block is a structural mirror of the existing anthropic-base / azure / vertex / custom / bedrock blocks: same `splitList`/`trim`/dedupe/`mustToJson` idiom, same `os.environ/*` key wiring, same `LITELLM_DEFAULT_MODEL` construction. No new approaches invented.
- **Consumer-friendly config (Principles 2–3).** New `*_models` textareas have descriptive titles + `help_text` with concrete examples; consumers set values directly with no hidden logic.
- **Correct `when` gating.** Provider blocks are mutually exclusive on `llm_provider`, and Gemini is c- **Correct `when` gating.** Provider blocks are mutually exclusive on `llm_provider`oog- **Correct `when` gatto - **Correct `when` gating.** Provider blocks are mutually exclusive on `llm_provider`, and Gemini is c- **Correct `when` gating.** Provider blocks are mutually exclusive on `llm_provider`oog- **Correct `when` gatto - **Correct `when` gating.** Provider blocks are muht prefix, every `o- **Correct `when` gating.** Provider blocks are mutually exclusive on `llm_provider`, and Gemini is c- **Correct `when` gating.** Provider blocks are mutually exclusive on `llm_provider`oog- **Correct `when` gatto - **Correct `when` gating.** Provider blocks are mutually exclusive on `llm_provider`, and Gemini is c- **Correct `wh 0- **Correct `when` gating.** Provider blocks are mutually exclusive on `llm_provider`, and Gemini is c- **Correct `when` gating.** Provider blocks are mutually exclusive on `llm_provider`y because each `*_models` item is `required: true` and prefilled. Worth being aware that if an admin clears a field to just commas/whitespace, `required: true` is still satisfied but `model_list` renders `[]` and `LITELLM_DEFAULT_MODEL` becomes a bare `litellm_proxy/`. This footgun already exists in the current azure/openai-style blocks, so it's consistent — flagging only as a future hardening idea (e.g., validate at least one token), not a change request.
- Prefilled model ids (`gpt-5`, `gemini-2.5-pro`, `openrouter anthropic/claude-sonnet-4.5`) will age over time, but they're admin-editable, so this is fine.

---

aivong-openhands:
## Mutation review of the new provider-route tests

I hand-wrote a small set of mutants against the KOTS templates this PR adds and ran them past `scripts/test_replicated_llm_provider_routes.py`. Baseline on `alona/wire-llm-providers`: **40 passed in ~3.4s**. Scope here is test *strength* only — whether the suite pins the behaviour, not whether the behaviour is right.

### Controls — the suite's core claims are pinned ✅

I reverted the essence of the fix three ways on the OpenAI block (all six new blocks are structurally identical, so this stands in for the set):

| Control | Result |
|---|---|
| Disable the block (`when` gate → impossible value) so `openai` falls back to the Anthropic base | ❌ caught |
| Route prefix `openai/` → `anthropic/` | ❌ caught |
| Credential `os.environ/OPENAI_API_KEY` → `os.environ/ANTHROPIC_API_KEY` | ❌ caught |

What makes these land: the suite doesn't just check a route exists — `test_route_credentials_are_exported_by_the_secret` re-renders the actual `litellm-env-secrets` Secret and asserts every `os.environ/*` a route reads is a key that Secret exports, and `test_routes_use_only_the_selected_provider` asserts *every* route carries the selected provider's prefix. That pair is what catches "fell through to Anthropic". Nice.

### Survivors — the parsing robustness is unasserted ✅ (passed)

Every provider block advertises comma-tolerant, CRLF-tolerant, trimmed, deduplicated parsing (and the PR description calls this out explicitly). But each parametrised case only ever renders the provider's **default** `*_models` value — newline-separated, whitespace-free, all-distinct. So nothing drives the parsing branches:

| Mutant (OpenAI block) | Result |
|---|---|
| Drop comma split (`\| replace "," "\n"`) | ✅ passed |
| Neuter dedup guard (`hasKey $seen $m` → always-false) | ✅ passed |
| Drop `trim` (`$m := trim $raw` → `$m := $raw`) | ✅ passed |
| Drop CRLF split (`\| replace "\r" "\n"`) | ✅ passed |

#### Why it matters

The help text says "one per line", but the template also splits on commas and the PR promises it. An admin who types `gpt-5, gpt-4.1` on one line, indents a model, or pastes the same id twice hits exactly these branches. With comma-splitting broken they'd get a single bogus model named `gpt-5, gpt-4.1` → route `openai/gpt-5, gpt-4.1`, which only fails later at inference. Untrimmed tokens misfire the same way; a duplicate silently produces two identical `model_list` entries. None of that is a rendering error the current suite would catch.

#### Fix — one adversarial-input case per provider (verified)

The four survivors share one root cause and one fix: feed a messy value and assert the parsed, cleaned, deduped result. Added to `scripts/test_replicated_llm_provider_routes.py`:

```python
def test_openai_models_parsing_is_comma_trim_dedup_tolerant() -> None:
    # comma-joined line, an indented line with surrounding spaces, a duplicate
    _, routes = _render_with(
        {
            "llm_provider": "openai",
            "openai_api_key": "k",
            "openai_models": "gpt-a, gpt-b\n  gpt-c  \ngpt-a",
        }
    )
    names = [r["model_name"] for r in routes]
    models = [r["litellm_params"]["model"] for r in routes]
    assert names == ["gpt-a", "gpt-b", "gpt-c"], names
    assert models == ["openai/gpt-a", "openai/gpt-b", "openai/gpt-c"], models
```

`_render_with` is `render_routes` with the config overridable (three lines, reusing the existing `_config_options` / `_optional_litellm_blocks` / `_helm_render` helpers). I verified both halves: it **passes** on the PR branch unmodified, and it **fails** with the comma, `trim`, and dedup mutants applied. Worth parametrising over all six providers, since they share the code.

#### Equivalent mutant — CRLF

The "drop `\r` split" mutant survives even against the messy-input test above, and I think it's genuinely equivalent: `trim` already strips the trailing CR from a `\r\n` pair, so `replace "\r" "\n"` only changes anything for lone-CR (classic-Mac) line endings, which no realistic paste produces. I'd leave it untested rather than add a case for it — flagging only so it's a decision, not an oversight.

### Not a test gap

- The gating asymmetry is intentional and already covered: the Gemini block gates on `llm_provider=google` **and** `google_api_type=gemini` (because `google_api_type` defaults to `gemini`), while the Vertex block gates on `google_api_type=vertex` alone. The `vertex` parametrised case would fail if the Gemini block leaked onto it, so that distinction is pinned.
- `LITELLM_DEFAULT_MODEL` resolution (first non-empty line → `litellm_proxy/<id>`) is well covered by `test_default_model_is_a_visible_route`.

---
This comment was generated by an AI assistant on behalf of the user.

replicated/openhands.yaml
```diff
@@ -1069,6 +1069,71 @@ spec:
...
+    # line of the "OpenRouter Models" textarea. OR_SITE_URL / OR_APP_NAME come
+    # from the base block's proxy_config.environment_variables.
+    - when: '{{repl ConfigOptionEquals "llm_provider" "openrouter" }}'
+      recursiveMerge: true
+      values:
+        env:
+          LITELLM_DEFAULT_MODEL: 'litellm_proxy/repl{{ $first := "" }}repl{{ range $raw := splitList "\n" (ConfigOption "openrouter_models" | replace "\r" "\n" | replace "," "\n") }}repl{{ $m := trim $raw }}repl{{ if and $m (not $first) }}repl{{ $first = $m }}repl{{ end }}repl{{ end }}repl{{ $first }}'
+        litellm-helm:
+          proxy_config:
+            model_list: repl{{ $models := list }}repl{{ $seen := dict }}repl{{ range $raw := splitList "\n" (ConfigOption "openrouter_models" | replace "\r" "\n" | replace "," "\n") }}repl{{ $m := trim $raw }}repl{{ if and $m (not (hasKey $seen $m)) }}repl{{ $seen = set $seen $m true }}repl{{ $models = append $models (dict "model_name" $m "litellm_params" (dict "model" (printf "openrouter/%s" $m) "api_key" "os.environ/OPENROUTER_API_KEY")) }}repl{{ end }}repl{{ end }}repl{{ $models | mustToJson }}
```
hieptl: Hello Alona,

Installs that already selected one of these six providers still have `litellm_proxy/claude-sonnet-4-5-20250929` stored in org defaults and member settings, and that route disappears in this release. The custom and Bedrock blocks already handle the same upgrade problem with a hidden alias (`openhands_hidden` plus `openhands_canonical`), so existing settings keep resolving without every admin and member having to re-select a model. Could we add the same kind of hidden `claude-sonnet-4-5-20250929` alias pointing at the first configured route here (and in the other five blocks)? If the name mismatch in LiteLLM spend logs is a concern, an upgrade note telling operators to re-select defaults would also work.

Thank you very much :pray:

hieptl (approved):

replicated/openhands.yaml
ak684: Fixed in 98739934fe82361c8b4148ab6c49584f17a82fa8. All six added providers now render the same hidden legacy-model alias convention used by existing provider blocks, forwarding to the first configured route. Explicit legacy-name routes remain visible and are not duplicated; empty lists produce no alias. The six alias regressions fail on the previous source. Added parameterized comma/CRLF/whitespace/duplicate and collision coverage as well: all 64 render tests and both checksum guards pass.

ak684:
@openhands /codereview-roasted


Review exactly commit 98739934fe82361c8b4148ab6c49584f17a82fa8. If the PR head changes, report that and do not approve another commit. Review only: do not push code, merge, deploy, or change attribution. Check correctness, regression coverage, and whether the PR validation claims match the evidence. Keep code comments short. Post the substantive review and link it in your completion message. End the substantive review with these plain lines (no code fence), choosing one verdict:

REVIEWED_HEAD: 98739934fe82361c8b4148ab6c49584f17a82fa8
REVIEW_REQUEST: 1a0ded8dc4e74577901c28504e95ea2f
VERDICT: APPROVED

Use VERDICT: CHANGES_REQUESTED instead when findings require changes. If unable to review, explain the blocker without an approval verdict.

ak684:
## Review — commit `98739934fe82361c8b4148ab6c49584f17a82fa8`

Reviewed exactly the requested head (`9873993…`); PR head still matches it (verified again before posting). This head is a merge of the feature branch (`f37e067…`, which I reviewed earlier) with `main`; the PR's own content vs. the base is confined to three files: `replicated/config.yaml`, `replicated/openhands.yaml`, and `scripts/test_replicated_llm_provider_routes.py`.

### What I verified independently

- **Regression suite**: `pytest scripts/test_replicated_llm_provider_routes.py` — **64/64 pass** with helm v3.18.4 (the version CI pins). The 24 new cases (6 providers × alias-forwarding, trim/split/dedup, explicit-legacy, empty-list) plus the 40 existing ones all render the real KOTS templates end-to-end.
- **The six legacy-route regressions genuinely pin the fix**: copied the head's test file onto the previously reviewed commit `f37e067` → all six `test_legacy_model_route_remains_available[*]` fail (`StopIteration` — no alias exists there). The whole suite against base `main` fails 36 cases, since the new provider blocks don't exist at all.
- **Independent render of the alias logic** (not just through the test harness): rendered the OpenAI `model_list` block directly with helm + sprig.
  - Default list → first configured model is the visible route; a hidden `claude-sonnet-4-5-20250929` entry is appended whose `litellm_params` are an exact copy of the first route's and whose `model_info` is `{openhands_hidden: true, openhands_canonical: <first>}`. So an existing saved legacy model continues to work and reaches the real provider route.
  - Legacy name explicitly configured → exactly one route, visible, no duplicate alias.
  - Empty/whitespace-only list → no routes and no alias.
  - Dedup, CRLF/comma/newline splitting, and whitespace trimming behave as documented.
- **Credential wiring**: every `os.environ/<KEY>` referenced in the six new `model_list` blocks is exported by `litellm-env-secrets.yaml` (`OPENAI_API_KEY`, `GOOGLE_API_KEY`, `DEEPSEEK_API_KEY`, `MISTRAL_API_KEY`, `GROQ_API_KEY`, `OPENROUTER_API_KEY`); `test_route_credentials_are_exported_by_the_secret` covers this for all 16 provider variants. The Gemini block correctly keys off the exported `GOOGLE_API_KEY` (sourced from `google_gemini_api_key`).
- **Checksum guards**: `check_secret_checksum.py` (45 password fields) and `check_litellm_checksum.py` (25 deps) both pass; `test_litellm_credentials_checksum.py` 31/31 pass. All six new provider keys were already inside the pod-roll checksum, so key rotation rolls the LiteLLM pod.
- **Gemini gating**: the block is gated on `llm_provider=google` **and** `google_api_type=gemini` (needed because `google_api_type` defaults to `gemini`), and the vertex block is gated only on `google_api_type=vertex` — mutually exclusive, no overlap, verified by the existing vertex/gemini route tests.
- **Formatting**: `git diff --check` is clean and ruff format is clean. `ruff check` surfaces one `PLW1510` (`subprocess.run` without `check`) at `test_replicated_llm_provider_routes.py:146`; this line predates this commit (blamed to `f37e067`), ruff is not configured in the repo, and no CI job runs ruff — so it's not a gate and not introduced here.
- **CI on the head commit**: `test-scripts`, `check-secret-checksum`, `lint` (make lint via Replicated CLI), and the chart lint-and-test jobs all completed **success**.

### Notes

- The PR description's "64 Helm-render regression tests" matches the suite count exactly (40 prior + 24 new).
- Config fields are `required: true` with prefilled `value`s and provider-scoped `when`s, matching the existing azure/bedrock/custom blocks — the empty-models concern raised in the earlier review round was already dispositioned.
- Live inference coverage is limited to OpenRouter (as stated); the other five providers have render coverage only. That's a stated limitation, not an inconsistency.

No findings at the "important" level; the change is correct, follows the established optionalValues pattern, and the validation claims match the evidence.

REVIEWED_HEAD: 98739934fe82361c8b4148ab6c49584f17a82fa8
REVIEW_REQUEST: 1a0ded8dc4e74577901c28504e95ea2f
VERDICT: APPROVED

PR #744, opened 2026-06-24
feat: wire runtime file archive config into Helm and Replicated (APP-2405)

simonrosenberg:
## Description

Wires the workspace **archive-before-delete** feature into the Helm charts and Replicated config so it can be enabled in deployed environments (APP-2405, OpenHands-Cloud#729).

There are two consumers of the same `RUNTIME_FILE_ARCHIVE_*` env, so config is plumbed to both:
- **runtime-api** cleanup CronJob — the primary, pause-time archive of the dominant idle/expiry reap (PLTF-3112, runtime-api#623).
- **OpenHands app-server** — the explicit-delete-while-running path (APP-2403, OpenHands#14953).

Changes:
- `charts/runtime-api`: new `runtimeFileArchive.{enabled,bucket,prefix,format}` → `RUNTIME_FILE_ARCHIVE_*` env in the shared `runtime-api.env` (so the cleanup CronJob and deployment both get it).
- `charts/openhands`: new `runtimeFileArchive.{enabled,bucket,prefix,format,required}` → `RUNTIME_FILE_ARCHIVE_*` env on the app-server.
- `replicated/`: new **Sandbox Configuration** options (archive enabled / bucket / require-on-delete) mapped into both the `runtime-api` subchart and the app-server values in `openhands.yaml`.
- **Default disabled** — existing deployments are unchanged. Bucket write access is granted to the service accounts via Workload Identity (noted in `values.yaml`; GCS: `roles/storage.objectAdmin` on the bucket).

Verified with `helm template` on both charts: with the feature off, **no** `RUNTIME_FILE_ARCHIVE_*` env is rendered; with it on, the vars render on the cleanup CronJob and the app-server. Replicated `config.yaml` parses; the new ConfigOptions map through `openhands.yaml`.

## Helm Chart Checklist

- [x] I have updated the `version` field in `Chart.yaml` for each modified chart (runtime-api 0.3.13 → 0.3.14 + openhands dep pin; openhands 0.7.67 → 0.7.68)
- [x] I have verified backwards compatibility with existing values.yaml configurations (feature defaults off; no env emitted when disabled)
- [ ] I have updated the chart's README.md if there are any breaking changes or new required values (none — additive, optional values; documented inline in values.yaml)

## Additional Notes

This was marked Done in Linear earlier without code; reopened and implemented here. Credentials are intentionally **not** a new secret — archiving uses the existing GCS Workload Identity of the runtime-api / app-server service accounts, consistent with the current `filestore` setup.

### DIFF
--- a/replicated/config.yaml
+++ b/replicated/config.yaml
@@ -894,6 +894,29 @@ spec:
             regex:
               pattern: ^[1-9][0-9]*$
               message: Must be a positive integer (in seconds)
+        - name: sandbox_archive_enabled
+          title: Archive Workspaces Before Deletion
+          help_text: >-
+            Copy each conversation's workspace to object storage before the
+            sandbox is paused/stopped or deleted, so it survives deletion (e.g.
+            for building evaluation datasets). The OpenHands and runtime-api
+            service accounts must have write access to the bucket below.
+          type: bool
+          default: "0"
+        - name: sandbox_archive_bucket
+          title: Archive Bucket
+          help_text: Object-storage bucket that receives workspace archives.
+          type: text
+          when: 'repl{{ ConfigOptionEquals "sandbox_archive_enabled" "1" }}'
+          required: true
+        - name: sandbox_archive_required
+          title: Require Archive to Succeed Before Deleting
+          help_text: >-
+            When enabled, an explicit sandbox deletion is blocked (and retried
+            later) if its workspace archive fails, instead of deleting anyway.
+          type: bool
+          default: "0"
+          when: 'repl{{ ConfigOptionEquals "sandbox_archive_enabled" "1" }}'
         - name: sandbox_storage_size
           title: Storage Size
           help_text: How much persistent storage each sandbox gets.
--- a/charts/openhands/Chart.yaml
+++ b/charts/openhands/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 description: OpenHands is an AI-driven autonomous software engineer
 name: openhands
 appVersion: cloud-1.38.0
-version: 0.7.67
+version: 0.7.68
 dependencies:
   - name: keycloak
     version: 24.7.5
@@ -34,7 +34,7 @@ dependencies:
     condition: replicated.enabled
   - name: runtime-api
     repository: oci://ghcr.io/openhands/helm-charts
-    version: 0.3.13
+    version: 0.3.14
     condition: runtime-api.enabled
   - name: automation
     repository: oci://ghcr.io/openhands/helm-charts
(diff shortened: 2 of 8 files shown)


replicated/config.yaml
```diff
@@ -894,6 +894,29 @@ spec:
             regex:
               pattern: ^[1-9][0-9]*$
               message: Must be a positive integer (in seconds)
+        - name: sandbox_archive_enabled
+          title: Archive Workspaces Before Deletion
+          help_text: >-
+            Copy each conversation's workspace to object storage before the
+            sandbox is paused/stopped or deleted, so it survives deletion (e.g.
+            for building evaluation datasets). The OpenHands and runtime-api
+            service accounts must have write access to the bucket below.
```
jlav, line 903: This is irrelevant for replicated installs. A user cannot configure a service account in replicated, nor wire up access to a bucket.

Unless there's actually a use case for this, I would recommend just not supporting this feature for replicated installs.

jlav (changes requested):

simonrosenberg:
Closing — two findings from review:

1. @jlav is right that the Replicated config is out of scope: a self-hosted/Replicated install can't wire up a GCP service account or bucket IAM, so the toggle can't function there (and enabling `required` would actually block sandbox deletion). Self-hosted customer workspaces also aren't the data this is for — the goal is OpenHands Cloud conversation data for eval/dataset creation.

2. More fundamentally, prod doesn't deploy runtime-api from this repo. Per `deploy/.github/workflows/deploy.yaml` (`deploy-runtime-api` → `chart_path: ./runtime-api/deploy/chart`, `envs_path: ./runtime-api/deploy/envs`), prod's runtime-api — and therefore the **primary** pause/reap capture path — is configured in the runtime-api repo, not in `OpenHands-Cloud/charts/runtime-api`. So this PR's runtime-api chart changes don't configure the dominant path in prod. The real prod config belongs together in `runtime-api/deploy` (primary), `deploy/openhands/envs/production` (app-server secondary), and `infra` (bucket + Workload Identity) — to be done there.

The code PRs (sdk#3867, runtime-api#623, OpenHands/OpenHands#14953, deploy#4709) remain open. Thanks @jlav for the catch.

simonrosenberg:
Superseded by #765, which makes the archive **storage backend selectable** (`storeType: google_cloud | s3 | local`). This directly addresses the objection that closed this PR — self-hosted/Replicated installs can now archive to the in-cluster minio or any S3-compatible store instead of requiring GCS. (GitHub wouldn't let me reopen this PR after the branch was rebased onto current main, hence the new one.)

PR #745, opened 2026-06-24
Prepare integrations-hub for Cloud SQL deployment

neubig:
## Summary
- Add Cloud SQL proxy support to the integrations-hub chart so migrations and the app can use GCP Cloud SQL via Workload Identity.
- Add optional dedicated-host ingress support for integrations-hub, while preserving the existing path-mounted behavior.
- Bump the integrations-hub chart to 0.1.2 and the OpenHands umbrella chart to 0.7.68.

## Testing
- Downloaded local Helm v3.18.6 binary because Helm was not installed in the workspace.
- helm dependency update charts/integrations-hub
- helm template test charts/integrations-hub with gcp.dbInstance/project/region set
- Rendered OpenHands umbrella chart from a temporary copy with integrations-hub dependency rewritten to file://../integrations-hub because 0.1.2 is not published yet
- Verified rendered manifests contain cloud-sql-proxy, 127.0.0.1 DB host, INTHUB_POSTGRES_URL, and integrations.openhands.dev ingress

## Deploy order
1. Merge OpenHands/integrations-hub#277 so the GHCR image workflow exists.
2. Merge this PR and wait for chart publication.
3. Merge the infra PR for DNS/database/secrets.
4. Merge the saas-deploy PR enabling staging/production values.

_This PR was created by an AI agent (OpenHands) on behalf of Graham Neubig._

### DIFF
--- a/charts/openhands/templates/ingress-integrations-hub.yaml
+++ b/charts/openhands/templates/ingress-integrations-hub.yaml
@@ -3,12 +3,18 @@
 
   By default the hub is mounted at integrations-hub.rootPath under the main
   OpenHands host. If integrations-hub.ingress.host is set, the same service is
-  served on that dedicated hostname instead, usually with rootPath="".
+  served on that dedicated hostname instead, usually with rootPath="". Host
+  values are rendered with tpl so environment-specific charts can choose their
+  own naming scheme without this shared chart knowing about it.
 */}}
 {{- $hub := index .Values "integrations-hub" -}}
 {{- if and .Values.enabled .Values.ingress.enabled $hub.enabled }}
 {{- $rootPath := $hub.rootPath | default "/integrations-hub" }}
 {{- $dedicated := and $hub.ingress.enabled $hub.ingress.host }}
+{{- $host := tpl .Values.ingress.host . }}
+{{- if $dedicated }}
+{{- $host = tpl $hub.ingress.host . }}
+{{- end }}
 {{- $ingressPath := $rootPath }}
 {{- if $dedicated }}
 {{- $ingressPath = "/" }}
@@ -28,26 +34,16 @@ spec:
   {{- if and $dedicated $hub.ingress.tls.enabled }}
   tls:
   - hosts:
-    - {{ $hub.ingress.host }}
+    - {{ $host }}
     secretName: {{ $hub.ingress.tls.secretName }}
   {{- else if .Values.tls.enabled }}
   tls:
   - hosts:
-    {{- if .Values.ingress.prefixWithBranch }}
-    - {{ .Values.branchSanitized }}.{{ .Values.ingress.host }}
-    {{- else }}
-    - {{ .Values.ingress.host }}
-    {{- end }}
+    - {{ $host }}
     secretName: app-all-hands-{{ .Values.tls.env }}-tls
   {{- end }}
   rules:
-  {{- if $dedicated }}
-  - host: {{ $hub.ingress.host }}
-  {{- else if .Values.ingress.prefixWithBranch }}
-  - host: {{ .Values.branchSanitized }}.{{ .Values.ingress.host }}
-  {{- else }}
-  - host: {{ .Values.ingress.host }}
-  {{- end }}
+  - host: {{ $host }}
     http:
       paths:
       - path: {{ $ingressPath | quote }}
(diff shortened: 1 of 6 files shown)


charts/integrations-hub/templates/_env.yaml
```diff
@@ -32,8 +32,12 @@
 - name: INTHUB_POSTGRES_URL
   value: {{ .Values.database.url | quote }}
 {{- else }}
+{{- $dbHost := .Values.database.host }}
+{{- if and .Values.gcp.dbInstance (not $dbHost) }}
+{{- $dbHost = "127.0.0.1" }}
+{{- end }}
```
jlav: The charts are meant to be generic and shouldn't assume any particular cloud environment.

Values like `gcp.dbInstance` should just use generic counterparts, like `db.Host` instead.

charts/integrations-hub/templates/deployment.yaml
```diff
@@ -24,6 +24,21 @@ spec:
...
+      - name: cloud-sql-proxy
+        image: {{ .Values.gcp.cloudSqlProxy.image | quote }}
+        restartPolicy: Always
+        args:
+          - "--structured-logs"
+          - "--address=127.0.0.1"
+          - "--port={{ .Values.database.port }}"
+          - "{{ .Values.gcp.project }}:{{ .Values.gcp.region }}:{{ .Values.gcp.dbInstance }}"
+        resources:
+          {{- toYaml .Values.gcp.cloudSqlProxy.resources | nindent 10 }}
```
jlav: Just add this sidecar to the values file for the environment this will execute in. A cloud-sql proxy sidecar is highly specific to our configuration of OHE, but is not applicable to most installs.

charts/openhands/templates/ingress-integrations-hub.yaml
```diff
@@ -42,14 +41,16 @@ spec:
     secretName: app-all-hands-{{ .Values.tls.env }}-tls
   {{- end }}
   rules:
-  {{- if .Values.ingress.prefixWithBranch }}
+  {{- if $dedicated }}
+  - host: {{ $hub.ingress.host }}
+  {{- else if .Values.ingress.prefixWithBranch }}
```
jlav: `prefixWithBranch` is another "us" thing. This is meaningless for every other consumer.

Since this is being used to decide what host name to use, rather than use a conditional, just make the host name templatable via the values file and the consumer can decide what host name they want.

jlav (changes requested):
Left some blocking comments about making these changes generic rather than specific to our cloud environment.

Appreciate your patience on these. I know this is based on patterns you're already finding in our charts, but these are habits that we're trying to move away.

charts/integrations-hub/templates/_env.yaml
neubig: Done in 0b80979: removed the GCP-specific database-host fallback. The chart now reads only the generic database.host/database.url values; environments that use a proxy set database.host themselves. _This reply was created by an AI agent (OpenHands) on behalf of Graham Neubig._

charts/integrations-hub/templates/deployment.yaml
neubig: Done in 0b80979: removed the built-in Cloud SQL proxy branch from the chart and replaced it with generic extraInitContainers rendered from values. Cloud SQL proxy config can now live entirely in the consuming environment values. _This reply was created by an AI agent (OpenHands) on behalf of Graham Neubig._

charts/openhands/templates/ingress-integrations-hub.yaml
neubig: Done in 0b80979: removed the prefixWithBranch conditional from the integrations-hub ingress and render host values with tpl instead, so consumers choose the host naming scheme in values. _This reply was created by an AI agent (OpenHands) on behalf of Graham Neubig._

neubig:
Addressed the chart-genericity review comments in 0b80979. Validation run locally: helm lint charts/integrations-hub, and helm lint/template for openhands with the local integrations-hub subchart plus templated host and extraInitContainers. Ready for re-review. _This comment was created by an AI agent (OpenHands) on behalf of Graham Neubig._

neubig:
Re-requesting AI review on final head 26755d10; the generic chart fixes are in 26755d1. _This comment was created by an AI agent (OpenHands) on behalf of Graham Neubig._

jlav (approved):
Thank you!

PR #764, opened 2026-06-29
feat(charts): map-based warm runtimes + global agent-server image

jlav:
## Description

The warm runtime config in the runtime-api chart was a list (`warmRuntimes.configs`). Helm can't merge lists across value layers, so a consumer who wanted to change one field of one entry had to re-declare the whole list. This converts it to a map (`warmRuntimes.configsByName`) keyed by each entry's `name`, which merges cleanly layer to layer.

The list form still works and takes precedence when it's set, so existing consumers keep working untouched.

It also adds `global.agentServerImage` as the single source for the agent-server image. `runtime.image` and any `configsByName` entry that omits its own `image` both fall back to it, so the proxy / custom-sandbox / local-registry variants only override the repository (and the tag, for the admin-supplied custom image). The replicated wrapper now uses both the map form and the global, which collapses a large block of duplicated warm config.

## Helm Chart Checklist

- [x] I have updated the `version` field in `Chart.yaml` for each modified chart (umbrella `openhands` 0.7.80 -> 0.7.81; runtime-api is an inert embedded `0.0.0`)
- [ ] I have tested the chart upgrade path from the previous version (see notes - I verified render-equivalence, not a live `helm upgrade`)
- [x] I have updated the chart's README.md if there are any breaking changes or new required values (there are none - `global.agentServerImage` ships with defaults, nothing new is required)

## Additional Notes

A few things worth a reviewer's attention:

- **Backwards compatibility.** `configsByName` is the new canonical form, but `configs` still renders and wins when it's non-empty. I confirmed with `helm template` that an existing list-based config renders byte-identical to before.

- **Incidental fix for local-registry installs.** The replicated wrapper's local-registry warm override used to sit at `runtime.warmRuntimes`, a key the chart never reads (warm config lives under `runtime-api`). So local-registry / airgapped installs' warm pods silently fell back to the proxy image path. Routing the warm image through `global.agentServerImage`, which the local-registry block does override, fixes that.

- **The tag now lives only in the chart default.** The replicated wrapper sets the agent-server repository and inherits the tag from `global.agentServerImage.tag` via map-merge. That continues the direction of #758, and it let me delete the whole `update_replicated_openhands_values()` image-bump path (function, 4 regexes, and its tests) since the wrapper no longer carries a literal agent-server tag.

- **Two pre-existing test failures, left as-is.** `test_every_runtime_api_ref_in_real_embedded_values_is_matched` and `test_automation_ref_in_real_embedded_values_is_matched` are red from earlier sha-vs-version tag drift (#758 / #759), unrelated to this change. Everything else passes (248).

### DIFF
--- a/.github/workflows/scan-docker-images.yml
+++ b/.github/workflows/scan-docker-images.yml
@@ -27,14 +27,14 @@ jobs:
           # Extract image tags from chart values using yq
           RUNTIME_API_TAG=$(yq '.image.tag' charts/openhands/charts/runtime-api/values.yaml)
           ENTERPRISE_TAG=$(yq '.image.tag' charts/openhands/values.yaml)
-          RUNTIME_TAG=$(yq '.runtime.image.tag' charts/openhands/values.yaml)
+          AGENT_SERVER_TAG=$(yq '.global.agentServerImage.tag' charts/openhands/values.yaml)
 
           # Build the image list as JSON array
           IMAGES=$(jq -n -c \
             --arg runtime_api "openhands/runtime-api:${RUNTIME_API_TAG}" \
             --arg enterprise "openhands/enterprise-server:${ENTERPRISE_TAG}" \
-            --arg runtime "openhands/agent-server:${RUNTIME_TAG}" \
-            '[$runtime_api, $enterprise, $runtime]')
+            --arg agent_server "openhands/agent-server:${AGENT_SERVER_TAG}" \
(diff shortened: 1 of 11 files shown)


aivong-openhands (approved):
Thanks for updating the script as well!

dylan-openhands:
Nice, this will definitely be good to have!

dylan-openhands (approved):

Issue #729, opened 2026-06-17
Configure GCS/S3 workspace archiving in Helm deployments


neubig:
## Goal

Expose workspace archiving as Helm configuration so a runtime's workspace can be preserved before the runtime is paused, reaped, or explicitly deleted.

## Architecture

There are two archive producers that require compatible configuration:

- **runtime-api cleanup CronJob** — the primary final-state capture path before idle/expiry pause or reap.
- **OpenHands app-server** — captures final state before explicit deletion of a running sandbox and can optionally capture the initial repository state before `setup.sh`.

The deployment integration supports two storage protocols:

- `google_cloud` — native Google Cloud Storage.
- `s3` — AWS S3 or an S3-compatible service such as MinIO.

This is configuration for built-in upload implementations, not a generic storage-plugin mechanism.

## Scope

- Add `runtimeFileArchive` values to the OpenHands chart and its runtime-api subchart.
- Render `RUNTIME_FILE_ARCHIVE_*` environment variables for both archive producers.
- Allow operators to select GCS or S3-compatible storage and supply connection configuration.
- Keep the feature disabled by default.
- Require operators to provision the archive bucket and grant the producer identities write access.

Relevant paths:

- `charts/openhands/templates/_env.yaml`
- `charts/openhands/values.yaml`
- `charts/openhands/charts/runtime-api/templates/_env.yaml`
- `charts/openhands/charts/runtime-api/values.yaml`

## Out of scope

The Replicated installer does not expose this feature. It is not a self-hosted customer workflow, and the installer currently provides neither external archive-storage configuration nor supported access to archives in bundled MinIO.

## Acceptance criteria

- Helm operators can configure backend, existing bucket, prefix, archive format, and relevant connection settings without editing templates.
- Both archive producers receive compatible configuration.
- Initial-state capture can be enabled independently for the app-server path.
- Existing deployments render no archive environment variables while disabled.
- Unsupported backend values fail clearly rather than silently routing elsewhere.
- Values documentation explains bucket and storage-permission responsibilities.

## Implementation

- Helm wiring: #765
- runtime-api S3 upload support: OpenHands/runtime-api#628

simonrosenberg:
[...]

PR #765, opened 2026-06-29
feat(charts): configure GCS/S3 workspace archiving in Helm (APP-2405)

simonrosenberg:
## Summary

Wires workspace archiving into the OpenHands Helm chart and its runtime-api subchart (APP-2405, #729).

The chart supports two storage protocols:

- `google_cloud` — native GCS using the pod ServiceAccount / Workload Identity.
- `s3` — AWS S3 or an S3-compatible service such as MinIO.

`storeType` selects one of these built-in upload implementations; this is not a generic storage-plugin mechanism.

This supersedes #744, whose GCS-only wiring could not support S3-compatible deployments.

## Archive producers

Configuration is projected to both producers:

- **runtime-api cleanup CronJob** — the primary final-state capture path before idle/expiry pause or reap. S3 upload support is provided by OpenHands/runtime-api#628.
- **OpenHands app-server** — captures final state before explicit deletion of a running sandbox and can optionally capture the initial repository state before `setup.sh` (OpenHands/OpenHands#14953).

### App-server

`runtimeFileArchive.storeType` can be set to `google_cloud` or `s3`. When empty, the chart selects `s3` for `filestore.ephemeral=true` and `google_cloud` otherwise.

The ephemeral S3 path reuses the app-server's existing `AWS_S3_*` environment. For an external S3-compatible store, the operator supplies endpoint and credentials through the pod's normal environment, mounted Secret, or AWS workload-identity mechanism.

### runtime-api cleanup

`runtime-api.runtimeFileArchive.storeType` defaults to `google_cloud` and can be set to `s3`.

For `s3`, operators can configure `s3.endpoint`, `s3.region`, inline credentials, or `s3.existingSecret`. When no endpoint and no external-store marker (`region` or `existingSecret`) is present, the endpoint defaults to the bundled in-cluster MinIO service. An endpoint-less external AWS configuration uses boto3's regional endpoint and TLS defaults.

The runtime-api and app-server have separate pod configuration, so an external destination must be configured for both producers.

## Values added

- App-server: `runtimeFileArchive.{enabled,storeType,bucket,prefix,path,format,required}`.
- Optional initial snapshot: `runtimeFileArchive.initial.{enabled,format,deadline}`.
- runtime-api: the corresponding archive values plus `runtimeFileArchive.s3.*` connection settings.

The feature is disabled by default. When disabled, no `RUNTIME_FILE_ARCHIVE_*` environment variables are rendered.

## Explicit infrastructure contract

- The chart does not create archive buckets. The configured GCS, S3, or MinIO bucket must already exist.
- GCS: grant the relevant producer ServiceAccount object-write permission, such as `roles/storage.objectAdmin` on the archive bucket.
- S3-compatible storage: grant object-write access and provide endpoint/region/TLS and credentials or pod identity as appropriate.

## Replicated scope

This PR intentionally does not expose workspace archiving through the Replicated installer. The feature serves OpenHands-operated data capture rather than a self-hosted customer workflow, and Replicated customers currently have neither external-storage configuration nor a supported way to retrieve archives from bundled MinIO.

The generic chart values remain disabled by default, so Replicated installations do not activate this feature.

## Verification

- `helm lint charts/openhands` — clean.
- Feature disabled: no archive environment variables rendered.
- GCS values render for both producers.
- S3-compatible values render for both producers, including the bundled-MinIO endpoint behavior.
- `git diff --check` — clean.

## Helm chart checklist

- [x] Bumped `charts/openhands` from `0.7.83` to `0.7.84`.
- [ ] Tested the chart upgrade path from the previous version.
- [x] Verified backward compatibility: archiving is disabled by default.
- [x] Documented all additive values inline.

Refs #729

### DIFF
--- a/replicated/config.yaml
+++ b/replicated/config.yaml
@@ -924,6 +924,17 @@ spec:
           type: bool
           default: "0"
           when: 'repl{{ ConfigOptionEquals "sandbox_archive_enabled" "1" }}'
+        - name: sandbox_archive_initial
+          title: Also Snapshot Initial Workspace State
+          help_text: >-
+            In addition to the final state, snapshot each conversation's INITIAL
+            workspace (the repository exactly as cloned, before setup runs) — the
+            "base" state needed to reconstruct an evaluation from the archive.
+            Captured at conversation start (bounded, best-effort), so it adds a
+            small amount of startup latency to repo-backed conversations.
+          type: bool
+          default: "0"
+          when: 'repl{{ ConfigOptionEquals "sandbox_archive_enabled" "1" }}'
         - name: sandbox_storage_size
           title: Storage Size
           help_text: How much persistent storage each sandbox gets.
--- a/replicated/config.yaml
+++ b/replicated/config.yaml
@@ -894,6 +894,34 @@ spec:
             regex:
               pattern: ^[1-9][0-9]*$
               message: Must be a positive integer (in seconds)
+        - name: sandbox_archive_enabled
+          title: Archive Workspaces Before Deletion
+          help_text: >-
+            Copy each conversation's workspace to object storage before the
+            sandbox is paused/stopped or deleted, so it survives deletion (e.g.
+            for building evaluation datasets). Archives are written to the
+            bundled object store (minio) by default; set a dedicated bucket
+            below. Disabled by default.
+          type: bool
+          default: "0"
+        - name: sandbox_archive_bucket
+          title: Archive Bucket
+          help_text: >-
+            Object-storage bucket that receives workspace archives (under the
+            "workspace-archives/" prefix). Defaults to the bundled sessions
+            bucket; point at a dedicated, pre-created bucket to keep archives
+            separate.
+          type: text
+          default: "openhands-sessions"
+          when: 'repl{{ ConfigOptionEquals "sandbox_archive_enabled" "1" }}'
+        - name: sandbox_archive_required
+          title: Require Archive to Succeed Before Deleting
+          help_text: >-
+            When enabled, an explicit sandbox deletion is blocked (and retried
+            later) if its workspace archive fails, instead of deleting anyway.
+          type: bool
+          default: "0"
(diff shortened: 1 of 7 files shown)


replicated/config.yaml
```diff
@@ -894,6 +894,47 @@ spec:
...
+          default: "0"
+        - name: sandbox_archive_bucket
+          title: Archive Bucket
+          help_text: >-
+            Object-storage bucket that receives workspace archives (under the
+            "workspace-archives/" prefix). Defaults to the bundled
+            "openhands-archives" bucket, which is provisioned for you and
+            preserved across upgrades. Do NOT use "openhands-sessions" — that
+            bucket is emptied on every upgrade. To use your own bucket, create
+            it first in the object store.
```
jlav: A replicated consumer has no mechanism to provision another bucket minio bucket, so exposing this configuration isn't useful.

If the intent is to allow storage to an external bucket, then there would need to be configuration options to add access keys.

replicated/config.yaml
```diff
@@ -894,6 +894,47 @@ spec:
             regex:
               pattern: ^[1-9][0-9]*$
               message: Must be a positive integer (in seconds)
+        - name: sandbox_archive_enabled
```
jlav: We don't give replicated users direct access to the minio storage bucket in the cluster.

It seems like this is a really esoteric feature that makes sense for us, but I'm struggling to see why this would be useful for self hosted customers.

If there actually is a use case, then there's going to need to be some way, via the replicated installer, to either configure an external bucket, or we need to wire up a way for consumers to get access to the minio bucket where this data is stored.

jlav (changes requested):
Thanks for remembering to consider the Replicated installer when getting this change together!

I think the approach for Replicated either needs to be re-worked a little given the constraints that I mentioned, or we should just not support this feature in Replicated if there's no clear use case for it in self hosted setups.

jlav (approved):

PR #766, opened 2026-06-29
feat(postgres): migrate embedded postgres from emptyDir to a PVC

jlav:
## Description

Embedded postgres stored its data directory on an emptyDir volume, so any pod reschedule permanently destroyed the database.

The root cause was a silent nesting bug. The Replicated layer set `postgresql.persistence.enabled`, but the bitnami chart only honors `postgresql.primary.persistence.enabled` and ignores the top-level key. Persistence was never actually on, so the StatefulSet fell back to emptyDir.

This moves embedded postgres onto a PVC so it's durable. The hard part is existing installs: `volumeClaimTemplates` is immutable, so you can't flip persistence on and upgrade in place. So this also carries the existing data across, in a single KOTS deploy with no manual steps.

External postgres is unaffected.

## Helm Chart Checklist

- [x] I have updated the `version` field in `Chart.yaml` for each modified chart (openhands 0.7.80 -> 0.7.81)
- [x] I have tested the chart upgrade path from the previous version (validated end to end on a test instance, see below)
- [x] I have verified backwards compatibility with existing values.yaml configurations (migration renders only for embedded postgres; disabled by default in the chart; external postgres renders nothing new)
- [x] I have updated the chart's README.md if there are any breaking changes or new required values (none - the new `postgresMigration` values all have defaults)

## Additional Notes

The migration is the non-obvious part. Three pre-upgrade hooks run, only for embedded postgres:

1. An orchestrator job guards on cluster state (no-op once the StatefulSet is already PVC-backed or a marker ConfigMap exists), then pre-creates the target PVC.
2. It launches a transient seeder pod that dumps the live database with `pg_dumpall` and restores it into a fresh, bitnami-initialized PGDATA on the PVC, checking the database count matches before going on.
3. It deletes the old emptyDir StatefulSet. Helm recreates it PVC-backed in the same upgrade, and the new pod adopts the seeded PVC.

A few decisions worth flagging:

- **The seeder pod is created imperatively by the orchestrator at runtime.** Once migrated, the PVC is held read-write by the live postgres pod, so a static Helm hook that mounted it would hit a multi-attach error and brick every later upgrade. Creating the seeder only when migration is actually needed keeps no-op upgrades from touching the live PVC.
- **No config toggle and no configurable size.** `volumeClaimTemplates` and PVC size are immutable, so a post-hoc toggle or resize would brick a future upgrade with an immutable-field error. The feature is always on for embedded postgres and guarded to a no-op once done. Size is set to 50Gi, which is nominal - the default `openebs-hostpath` SC enforces no quota, so the node disk is the real bound.
- **No write quiescing.** The dump is transactionally consistent as of its start, so writes committed to the old database during the dump window (a few seconds) aren't captured. Run the migration during a quiet period. I'd lean toward adding scale-to-zero of writers if that window ever matters, but it couples to deployment names and widens the outage, so I left it out.

Validation: ran this end to end on a test instance with embedded postgres on emptyDir. One deploy migrated it to a PVC with all 5 databases and exact row counts preserved. Deleting the postgres pod afterward (the event that used to wipe the data) kept everything intact. KOTS didn't roll the release back, which confirms the hook-delete-then-recreate of the immutable StatefulSet works under KOTS native helm.

### DIFF
--- a/charts/openhands/Chart.yaml
+++ b/charts/openhands/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 description: OpenHands is an AI-driven autonomous software engineer
 name: openhands
 appVersion: cloud-1.40.0
-version: 0.7.80
+version: 0.7.81
 dependencies:
   - name: keycloak
     version: 24.7.5
--- a/charts/openhands/templates/postgres-pvc-migration.yaml
+++ b/charts/openhands/templates/postgres-pvc-migration.yaml
@@ -1,0 +1,372 @@
+{{- /*
+One-time migration of embedded PostgreSQL from an emptyDir volume to a PVC.
+
+Why this exists: the embedded Bitnami `postgresql` StatefulSet historically ran
+its data directory on an emptyDir, so any pod reschedule destroyed the database.
+The fix (postgresql.primary.persistence.enabled) makes the StatefulSet
+PVC-backed, but `volumeClaimTemplates` is immutable — an in-place upgrade of an
+existing emptyDir StatefulSet would fail. These pre-upgrade hooks carry the
+existing data across in a single deploy:
+
+  1. (-20) RBAC: a ServiceAccount + namespaced Role for the orchestrator.
+  2. (-18) ConfigMap: the seed script and the PVC / seeder-pod manifests.
+  3. (-10) Orchestrator Job: guards (no-op once migrated), pre-creates the
+     target PVC, launches a *transient* seeder pod that hydrates the PVC from
+     the live database, then deletes the old emptyDir StatefulSet. Helm then
+     recreates the StatefulSet PVC-backed and it ADOPTS the seeded PVC.
+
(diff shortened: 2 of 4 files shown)


aivong-openhands (approved):

dylan-openhands:
I guess this is a perfectly safe migration because the current state today involves permanent destruction already 😬 ?

LGTM holler if there is any manual steps needed to migrate and an extra set of eyes would help

dylan-openhands (approved):

jlav:
@dylan-openhands The Replicated installer can install postgres into the k8s cluster that it manages. That embedded postgres instance is really only intended for testing and trial scenarios, and we recommend using an external postgres instance before using the Replicated installer in production.

That said, we definitely didn't intend for the data in that instance to be saved to non-durable storage, so this PR will migrate any existing data into durable storage if it exists already.

PR #767, opened 2026-06-30
feat(infra): add coordinator to relax sandbox PDBs for k0s upgrades

jlav:
## Description

runtime-api gives every sandbox a PodDisruptionBudget with `maxUnavailable=0`, so a sandbox pod is never voluntarily evicted. On embedded-cluster installs that protection has a side effect: it also blocks the node drains the k0s upgrade performs. Autopilot drains nodes through the eviction API, and a `maxUnavailable=0` PDB rejects every eviction, so a k8s version upgrade stalls on a cordoned node and never finishes.

This adds a small controller, the sandbox-upgrade-coordinator, so the upgrade can drain the node without giving up that PDB protection the rest of the time. While a k0s upgrade is rolling nodes, it relaxes the sandbox PDBs to `maxUnavailable=100%` so the drain can evict the pods, then sets them back to `maxUnavailable=0` once the upgrade finishes.

It's baked into the infra chart, disabled by default, and enabled for embedded-cluster installs - where the upgrade happens and the `Installation` CRD it watches exists. The sandbox Deployments stay at `replicas=1`, so each pod reschedules on its own once the node uncordons, and the node-local workspace PVC survives the reboot. The interruption is the pod being killed and recreated.

## Approach

A few choices worth flagging:

- **It acts on the PDB.** The PDB is the only thing blocking the drain. The obvious alternative - scaling the sandbox Deployment to zero - would fight runtime-api, which owns `spec.replicas` and the sandbox lifecycle state in its database, and it would bypass runtime-api's own pause/stop bookkeeping. Relaxing the PDB is the minimal change that respects that ownership.
- **It runs as a DaemonSet.** k0s autopilot and `kubectl drain` pass `IgnoreAllDaemonSets`, so this pod survives the very drain it has to act during. A Deployment's pod would get evicted with everything else and wouldn't be around to restore the PDBs afterward.
- **It's a standalone controller that reacts to `Installation` state.** A chart pre-upgrade hook would fire too late: embedded-cluster upgrades k0s before the application charts, so the node roll has already stalled by the time a hook runs. The controller watches the `Installation` state and reacts at the start of the k0s upgrade.

The selector only matches PDBs labeled `runtime_id`, so the app's own PDBs (keycloak, postgres, redis) are left alone. The relax is idempotent and only widens a PDB during a detected upgrade. If the controller is down, the worst case is the upgrade stalls on a cordoned node - autopilot retries the drain every 120s - so the PDB protection is never silently dropped.

## Helm Chart Checklist

- [x] I have updated the `version` field in `Chart.yaml` for each modified chart (infra `0.1.2` -> `0.1.3`)

The new `sandboxUpgradeCoordinator` values are additive and default to `enabled: false`, so existing `values.yaml` configs are unaffected. There are no breaking changes and no new required values, so the README doesn't need an update.

## Additional Notes

Validated on an embedded-cluster install:

- The coordinator deploys and runs as a DaemonSet.
- Against a real sandbox PDB (a warm-pool runtime), the scoping is right: it targets only the `runtime_id` PDB and leaves keycloak/postgres/redis alone.
- At steady state it issues zero patches and logs nothing, so there's no flapping.
- The k0s upgrade path is tested end-to-end: with a sandbox present, the upgrade drains the node and completes instead of stalling on the sandbox PDB.

### DIFF
--- a/charts/infra/templates/sandbox-upgrade-coordinator.yaml
+++ b/charts/infra/templates/sandbox-upgrade-coordinator.yaml
@@ -10,18 +10,30 @@ eviction API, which the PDB rejects, and the upgrade stalls on a cordoned node.
 Rather than drop the PDB, this controller removes the *pods* ahead of the drain
 so there is nothing for the PDB to protect, then brings them back afterwards.
 
-Why a Deployment and not a Helm hook: embedded-cluster upgrades k0s BEFORE it
-upgrades the application Helm charts (operator Upgrade(): upgradeK0s → addons →
-extensions). A chart pre-upgrade hook would therefore fire only AFTER the node
-roll already happened — too late. This controller runs continuously and reacts
-to the embedded-cluster `Installation` state, which flips to a drain state at
-the very start of the k0s upgrade (controllers roll before workers, so there is
-ample lead time before the first worker drain).
+Why it reacts continuously and not via a Helm hook: embedded-cluster upgrades
+k0s BEFORE it upgrades the application Helm charts (operator Upgrade():
+upgradeK0s → addons → extensions). A chart pre-upgrade hook would therefore fire
+only AFTER the node roll already happened — too late. This controller runs
+continuously and reacts to the embedded-cluster `Installation` state, which
+flips to a drain state at the very start of the k0s upgrade.
+
+Why a DaemonSet and not a Deployment: the upgrade drains the node, and most
+installs are single-node, so a Deployment's pod would be evicted along with
+everything else and never be around to scale the sandboxes down — deadlock,
+because the sandbox PDBs then block the very drain that would let the node
+upgrade. `kubectl drain` and k0s autopilot both pass IgnoreAllDaemonSets, so a
+DaemonSet pod is NOT evicted by the drain. It stays running on the draining node
+and removes the sandbox pods, which unblocks the drain. After the node reboots
+into the new k0s, the DaemonSet pod is recreated and restores the sandboxes.
 
 Safety: the PDB stays as a hard interlock. If this controller is down or slow,
 the worst case is the upgrade stalls on a cordoned node (autopilot's drain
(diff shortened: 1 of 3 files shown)


aivong-openhands (approved):

charts/infra/templates/sandbox-upgrade-coordinator.yaml
```diff
@@ -0,0 +1,274 @@
...
+  labels:
+    app.kubernetes.io/name: sandbox-upgrade-coordinator
+    app.kubernetes.io/instance: {{ .Release.Name }}
+data:
+  coordinator.sh: |
+    #!/bin/sh
+    # POSIX sh — alpine/k8s ships a shell and kubectl. Do NOT swap in
+    # rancher/kubectl here: it is FROM scratch and has no shell.
+    set -u
+
```
dylan-openhands: Thoughts on a quick top-level summary for the steps being taken here? Double check me on these if you agree

```suggestion

    # Sandbox upgrade coordinator — keeps sandbox PDBs out of the way during
    # k0s upgrades, then locks them back down. Control loop:
    #
    #   1. Every INTERVAL seconds, read status.state of all Installation CRDs.
    #   2. If any is in a drain state (upgrade rolling nodes) → relax every
    #      sandbox PDB to maxUnavailable=100% so the node can drain.
    #   3. Otherwise (idle) → restore every sandbox PDB to maxUnavailable=0
    #      to protect running sandboxes.
    #   4. If state is unreadable (CRD missing / API error) → leave PDBs as-is.
    #
    # Only PDBs matching the sandbox selector are touched, and only when the
    # value actually differs — steady state writes nothing and logs nothing.
    # 
```

dylan-openhands (approved):
One non-blocking suggestion. Otherwise good call for enabling a smooth-as-possible upgrade procedure

PR #768, opened 2026-06-30
ci: adopt release-please for per-chart helm chart releases

jlav:
## Description

Today every merge to main that touches `charts/**` publishes a stable chart version, and contributors have to hand-bump `version` in `Chart.yaml` (CI enforces it). That couples "merged" with "released": a chart ships a stable version the moment any change to it lands.

This adopts release-please, via the shared `OpenHands/release-actions` reusable workflows, to decouple releasing from merging:

- Every merge to main publishes an **alpha** chart to GHCR instead of a stable version, versioned off release-please's next candidate for that chart. An unreleased feat that moves openhands to 0.8.0 publishes `0.8.0-alpha.<run>`, a prerelease of the version that's genuinely coming next.
- **Stable** charts publish only when a per-chart release PR is merged. release-please tags `<chart>/<version>`, and that tag drives the stable publish, so the deploy repos and the `crd-check` OCI dependency keep resolving unchanged.
- Each chart (crd-check, image-loader, openhands, openhands-secrets) is its own release line with its own release PR, tag, and release. Embedded subcharts under `charts/openhands/charts/**` roll up into the openhands line, so there's one release PR per published chart, not per subchart.
- release-please owns the chart `version` now, so I removed the manual version-bump enforcement (`validate-chart-versions.yml` and its `pr-checks` job). Image-tag bump PRs move from `chore(...)` to `feat(...)` so they fold into a chart's release PR under Features (a minor bump, since feats move the minor on these pre-1.0 charts).

I validated this with a read-only release-please dry-run against live history. Each chart produced correctly path-scoped, github-style notes with the right bump (e.g. openhands 0.7.84 to 0.7.85), and nothing was created.

## Approach

A few choices worth a reviewer's attention:

- **The alpha version comes from release-please's candidate, computed after it runs.** A `publish-alpha` job in `release.yml` `needs` the per-chart release jobs, so each chart's release PR reflects the merge before the alpha publishes. It reads the proposed next version off that chart's release-PR branch (the manifest on `release-please--branches--main--components--<chart>`) and publishes `<candidate>-alpha.<run>`. This stays self-contained in this repo, with no extra output wired through the shared workflow. A chart with no pending release (a chore-only change, say) has no candidate and publishes nothing, and a guard skips publishing when the candidate already equals main's released version, which covers a release that was just cut.
- **One notes engine, so the release PR body and the published release match by construction.** The reusable workflow runs release-please in library mode with a custom changelog type (`github-scoped`, opted in with `github-scoped-notes: true`) that renders github-style notes (`## What's Changed`, `by @author in <PR>`, New Contributors, Full Changelog) scoped to a single chart's path. release-please copies the release body verbatim out of the merged PR, so what you review is what ships. Categories come from PR labels read live, not from git, so a mislabel is fixable without rewriting history.
- **Pinned at `@main`.** The github-scoped engine isn't in release-actions' `v0.2.0` tag, and the reusable workflow's internal action ref is hardcoded `@main`, so pinning the caller at `@main` keeps the workflow and its scripts in lockstep.
- **Tags keep the legacy `<chart>/<version>` format**, via `tag-separator: "/"` and `include-v-in-tag: false`. A nice side effect: release-please picks up the existing `<chart>/<version>` tags as each chart's last release, so the first release PR diffs from the real last published version (openhands from 0.7.84, and so on) rather than the start of history. I seeded the manifests to the live versions to match, and kept `bootstrap-sha` (current main HEAD) as a fallback for any chart with no matching tag - today just crd-check, which has no `crd-check/0.1.0` tag.
- **One repo setting changed (already applied):** `squash_merge_commit_title` is now `PR_TITLE`, so the conventional PR title is always the squash subject release-please reads. Without it, a single-commit PR keeps its raw, unlinted commit subject, and release-please silently drops anything non-conventional. The new `pr.yml` title lint backs this up.

## Helm Chart Checklist

No chart templates or values change in this PR, so this checklist doesn't apply. The first item is in fact what this PR removes: contributors no longer bump `Chart.yaml` versions by hand, since release-please owns them.

## Additional Notes

`release-replicated.yml` is untouched and still runs on every main push, labeling the Replicated release with the openhands `Chart.yaml` version. That version is now static between openhands releases, so consecutive Unstable releases share a label (Replicated still tells them apart by channel sequence). Not broken, but I'm happy to rework it separately if we want that label to move per build.

Per-PR previews (`preview-helm-charts.yml`) still version off a patch-bump of the current `Chart.yaml` plus `-alpha.<PR>`. release-please has no candidate for an unmerged branch, so the candidate scheme doesn't apply there. I left previews as they were; I'm happy to revisit if we want them on the same footing.


.github/workflows/release.yml
```diff
@@ -0,0 +1,306 @@
...
+          done
+
+  # Edge alpha publish. Waits for the release-please jobs so each chart's release
+  # PR (and therefore its proposed next version) reflects this push, then
+  # publishes an alpha of every changed chart at that chart's next candidate
+  # version. Main-only — release/** pushes just refresh their release PRs.
+  publish-alpha:
+    needs: [crd-check, image-loader, openhands, openhands-secrets, detect-changes]
+    if: github.ref == 'refs/heads/main'
+    runs-on: ubuntu-latest
```
aivong-openhands: We should use a fixed ubuntu image as new images can unexpectedly break a build for us

.github/workflows/release.yml
```diff
@@ -0,0 +1,306 @@
...
+
+      - name: Set up Helm
+        if: steps.check.outputs.should_publish == 'true' && steps.candidate.outputs.has_candidate == 'true'
+        uses: azure/setup-helm@v3
+        with:
+          version: 'latest'
+
+      - name: Install yq
+        if: steps.check.outputs.should_publish == 'true' && steps.candidate.outputs.has_candidate == 'true'
+        uses: mikefarah/yq@v4
```
aivong-openhands: I think yq is already available in the ubuntu image?

aivong-openhands (approved):

PR #772, opened 2026-07-02
ci: cut replicated releases in step with openhands

jlav:
## Description

`openhands` and the `replicated` config are versioned together, but the release pipeline treated them separately. release-please only bumped the openhands chart when files under `charts/openhands/` changed, so a change to `charts/infra` or `replicated/` didn't move the openhands version. And the Replicated app only published bleeding-edge main to the Unstable channel, so a cut openhands release didn't produce a matching Replicated release.

This ties the two together:

- release-please now treats the whole repo as openhands, excluding the three charts with their own release lines (`crd-check`, `image-loader`, `openhands-secrets`). A `feat` or `fix` under `charts/infra/` or `replicated/` folds into the next openhands release instead of going unversioned.
- When an openhands release is cut (the `openhands/<version>` tag), a new workflow builds the Replicated app from that commit and publishes it to the Beta channel, then tags the commit `Replicated/Beta/<sequence>` with the release sequence Replicated assigned.

Channel progression is now Unstable = every push to main (unchanged), Beta = every cut openhands release.

## Helm Chart Checklist

N/A - no chart templates or values changed. This only touches release automation: the openhands release-please config + manifest, `release.yml`, and a new `release-replicated-beta.yml`.

## Additional Notes

release-please has no "watch these extra paths" option for a component. The only path control is `exclude-paths`, a denylist. So the openhands package is rooted at the repo root (`.`) and excludes the three independently-released charts.

Rooting at `.` changes where the version write lands, since the helm strategy targets `<package>/Chart.yaml`:

- The bump is redirected to `charts/openhands/Chart.yaml` via an `extra-files` entry. The root `Chart.yaml` the helm strategy would otherwise write doesn't exist, and release-please skips a missing update file instead of failing.
- The manifest key moves from `charts/openhands` to `.`. Release continuity is keyed on the `openhands` tag, so the existing `openhands/0.7.84` line carries over.

Two supporting edits in `release.yml` keep alpha publishing correct: the alpha candidate lookup reads the `.` manifest key for openhands, and change-detection counts openhands as changed whenever a touched file sits outside those three charts.

One consequence to expect: the next openhands release PR will retroactively pick up any infra or replicated `feat`/`fix` landed since `openhands/0.7.84` was tagged.

**The Beta publish.** `release-replicated-beta.yml` fires on the `openhands/*` tag, checks out that commit, and runs `make release CHANNEL=Beta`. `CHANNEL=Beta` overrides the Makefile's branch-to-channel default, which is empty on a tag checkout (detached HEAD, no branch). The `VERSION` label comes from `charts/openhands/Chart.yaml` at the tagged commit, so the Beta release's version matches the openhands release.

**The sequence tag.** `replicated release create` prints `SEQUENCE: <n>` for the release it cuts. That's the global release sequence, the one `replicated release inspect <n>` takes. The workflow parses it and tags the same commit `Replicated/Beta/<n>`. `Replicated/*` is a new tag namespace and matches no workflow's tag triggers, so it doesn't kick off another run.

**Re-runs.** `replicated release create` has no version idempotency, so re-running the Beta workflow for the same openhands tag cuts a fresh Beta release with a new sequence and tag. That's the same behavior the existing Unstable-on-push workflow already has.

Scope boundary: promotion from Beta to Stable stays manual. This automates the Unstable and Beta lines only.


.github/workflows/release-replicated-beta.yml
```diff
@@ -0,0 +1,118 @@
...
+          tar -xzf replicated_*_linux_amd64.tar.gz -C /tmp/replicated-cli replicated
+          sudo mv /tmp/replicated-cli/replicated /usr/local/bin/
+
+      - name: Build and publish to Beta
+        id: release
+        env:
+          REPLICATED_APP: ${{ secrets.REPLICATED_APP }}
+          REPLICATED_API_TOKEN: ${{ secrets.REPLICATED_API_TOKEN }}
+        run: |
+          set -o pipefail
```
aivong-openhands: This step might benefit from shell: bash to get set -o pipefail by default

.github/workflows/release-replicated-beta.yml
```diff
@@ -0,0 +1,118 @@
...
+          ref: ${{ steps.tag.outputs.tag }}
+          fetch-depth: 0
+
+      - name: Set up Helm
+        uses: azure/setup-helm@v3
+        with:
+          version: 'latest'
+
+      - name: Install yq
+        uses: mikefarah/yq@v4
```
aivong-openhands: yq seems included in the ubuntu image?

.github/workflows/release-replicated-beta.yml
```diff
@@ -0,0 +1,118 @@
...
+        required: true
+
+# Never let two Beta publishes for the same release tag race.
+concurrency:
+  group: release-replicated-beta-${{ github.ref_name }}
+  cancel-in-progress: false
+
+jobs:
+  beta:
+    runs-on: ubuntu-latest
```
aivong-openhands: Can we fix the image to a version?

jlav: Done - pinned to `ubuntu-24.04` to match `release.yml`.

.github/workflows/release-replicated-beta.yml
jlav: Right, mikefarah's yq ships preinstalled on the ubuntu runner. Dropped the install step; `make` uses the runner's yq directly.

.github/workflows/release-replicated-beta.yml
jlav: Added `shell: bash` and removed the manual `set -o pipefail`. Confirmed the explicit shell runs with `-eo pipefail` while the omitted default is just `bash -e {0}`.

aivong-openhands (approved):

PR #774, opened 2026-07-02
feat: Expose Laminar v1 API ingress for self-hosted analytics

juanmichelini:
## Summary
- add an OpenHands-managed Laminar API ingress that exposes only `/v1` on the analytics hostname
- enable that ingress for self-hosted analytics in Replicated values
- include the ingress in Replicated expected resources and add regression tests

## Testing
- `uvx pytest scripts/test_laminar_api_ingress.py scripts/test_keycloak_realm_template.py`

## Why

Exposing /v1/ paths was suggested by Din from laminar see https://openhands-ai.slack.com/archives/C06U8UTKSAD/p1782941063706259?thread_ts=1782843637.758159&cid=C06U8UTKSAD

The concern is in the same thread, by Aarushi, currently we are not allowing api server access and it's needed to process laminar traces externally.

### DIFF
--- a/charts/openhands/templates/ingress-laminar-api.yaml
+++ b/charts/openhands/templates/ingress-laminar-api.yaml
@@ -1,0 +1,33 @@
+{{- $apiIngress := .Values.laminar.apiIngress | default dict -}}
+{{- $tls := $apiIngress.tls | default dict -}}
+{{- if and .Values.laminar.enabled $apiIngress.enabled $apiIngress.hostname }}
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: laminar-api-ingress
+  {{ with $apiIngress.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  {{- if $apiIngress.className }}
+  ingressClassName: {{ $apiIngress.className }}
+  {{- end }}
+  {{- if $tls.enabled }}
+  tls:
+    - hosts:
+        - {{ $apiIngress.hostname }}
+      secretName: {{ $tls.secretName }}
+  {{- end }}
+  rules:
+    - host: {{ $apiIngress.hostname }}
+      http:
+        paths:
+          - path: {{ $apiIngress.path | default "/v1" }}
+            pathType: {{ $apiIngress.pathType | default "Prefix" }}
+            backend:
+              service:
+                name: laminar-app-server-service
+                port:
+                  number: 8000
+{{- end }}
--- a/charts/openhands/values.yaml
+++ b/charts/openhands/values.yaml
@@ -498,6 +498,20 @@ laminar:
         clusterIssuer: ""
       annotations: {}
 
+  # OpenHands-managed ingress that exposes only Laminar's public ingestion API
+  # paths on the analytics hostname. This avoids exposing the full app-server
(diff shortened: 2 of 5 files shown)


charts/openhands/values.yaml
```diff
@@ -498,6 +498,20 @@ laminar:
         clusterIssuer: ""
       annotations: {}
 
+  # OpenHands-managed ingress that exposes only Laminar's public ingestion API
+  # paths on the analytics hostname. This avoids exposing the full app-server
+  # surface through the upstream Laminar appServer.ingress, which routes "/".
+  apiIngress:
```
aivong-openhands, line 514: The laminar chart already has helm values to support app server ingress https://github.com/OpenHands/saas-deploy/blob/main/app/openhands/environments/staging/values.yaml#L170

aivong-openhands: Ohh I see your comment

aivong-openhands: I think it would be worth talking to the Laminar team to see if their chart can support only exposing the ingestion API paths. What is the concern with exposing the full Laminar api server though?

juanmichelini: Exposing /v1/ paths was suggested by Din from laminar see https://openhands-ai.slack.com/archives/C06U8UTKSAD/p1782941063706259?thread_ts=1782843637.758159&cid=C06U8UTKSAD

The concern is in the same thread, by Aarushi, currently we are not allowing api server access and it's needed to process laminar traces externally.

aivong-openhands: Oh great there was already a discussion

charts/openhands/templates/ingress-laminar-api.yaml
```diff
@@ -0,0 +1,33 @@
...
+  {{- end }}
+  rules:
+    - host: {{ $apiIngress.hostname }}
+      http:
+        paths:
+          - path: {{ $apiIngress.path | default "/v1" }}
+            pathType: {{ $apiIngress.pathType | default "Prefix" }}
+            backend:
+              service:
+                name: laminar-app-server-service
```
aivong-openhands, line 30: It would be great if we could use variables here like LMNR_BASE_URL

charts/openhands/templates/ingress-laminar-api.yaml
```diff
@@ -0,0 +1,33 @@
...
+                port:
+                  number: 8000
```
aivong-openhands, line 32: LMNR_HTTP_PORT we have a helm value for that already I think? Could we use that here?

charts/openhands/templates/ingress-laminar-api.yaml
aivong-openhands: That could be a follow up

aivong-openhands (approved):
Just a couple of improvements for follow up. LGTM

charts/openhands/templates/ingress-laminar-api.yaml
juanmichelini: Addressed in follow-up PR #777 by deriving the ingress backend service from `env.LMNR_BASE_URL`.

charts/openhands/templates/ingress-laminar-api.yaml
juanmichelini: Addressed in follow-up PR #777 by using `laminar.httpPort` for the ingress backend port.

PR #796, opened 2026-07-07
Configure sandbox spec agent-server image

ak684:
## Summary

- Configure the default remote sandbox spec from the chart-rendered agent-server image values.
- Carry the default warm runtime command, working directory, and initial environment into the static sandbox spec.
- Keep the existing `AGENT_SERVER_IMAGE_*` env vars for compatibility with older OpenHands app images.

## Context

Companion for OpenHands/OpenHands#15168. That PR stops reading `AGENT_SERVER_IMAGE_REPOSITORY` and `AGENT_SERVER_IMAGE_TAG`, so OHE needs to provide the image through `OH_SANDBOX_SPEC_*` instead.

## Validation

- `helm dependency build charts/openhands`
- `helm template oh-main charts/openhands`
- `helm template oh-main charts/openhands --set-string global.agentServerImage.repository=local.registry/ns/agent-server --set-string global.agentServerImage.tag=9.9.9-python --set-string agentServerEnv.REQUESTS_CA_BUNDLE=/etc/test-ca.pem --set-string env.LLM_TIMEOUT=3600`
- `git diff --check`

### DIFF
--- a/charts/openhands/templates/_env.yaml
+++ b/charts/openhands/templates/_env.yaml
@@ -1,3 +1,44 @@
+{{- define "openhands.agentServerImage" -}}
+{{- printf "%s:%s" (.Values.runtime.image.repository | default .Values.global.agentServerImage.repository) (.Values.runtime.image.tag | default .Values.global.agentServerImage.tag) -}}
+{{- end -}}
+
+{{- define "openhands.defaultWarmRuntimeConfig" -}}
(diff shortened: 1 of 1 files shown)


ak684:
Closing this draft because current `main` already uses the dynamic sandbox-spec path, so this is not the right companion change.

The follow-up Cloud change should instead preserve explicit OHE/Replicated image overrides while avoiding default `AGENT_SERVER_IMAGE_*` env vars.

PR #806, opened 2026-07-08
PLTF-3196: Configure OpenHands resolver label for Replicated

aivong-openhands:
## Summary
- Adds a Replicated GitHub auth config option for the OpenHands resolver label, defaulting to openhands for existing installs.
- Maps the option to OH_RESOLVER_LABEL in the OpenHands optional values block.
- Adds focused manifest tests for the config option and env mapping.

## Tests
- uv run --with pytest pytest scripts/test_replicated_github_resolver_label.py
- ruby -e 'require "yaml"; ARGV.each { |path| YAML.load_file(path); puts "#{path}: ok" }' replicated/config.yaml replicated/openhands.yaml
- make lint

### DIFF
--- a/replicated/config.yaml
+++ b/replicated/config.yaml
@@ -718,6 +718,15 @@ spec:
           type: file
           when: 'repl{{ ConfigOptionEquals "github_auth_enabled" "1" }}'
           required: true
+        - name: github_resolver_label
+          title: OpenHands Resolver Label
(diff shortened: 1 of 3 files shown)


aivong-openhands:
Closing to restart this work on the resolver label scoped globally rather than GitHub-only.

PR #807, opened 2026-07-08
feat: PLTF-3196 Configure global OpenHands resolver label

aivong-openhands:
## Why
Replicated needs a single resolver trigger label that applies to all label/macro-based integrations, not just GitHub. Upstream OpenHands reads `OH_RESOLVER_LABEL` through the shared resolver-label helper used by GitHub, GitLab, Bitbucket/Data Center, Jira/Data Center, and Azure DevOps, while Slack stays separate.

Adding a global Advanced Options field that defaults to `openhands` preserves existing installs and lets operators set a custom label when they need an instance to stop responding to the default trigger.

## Validation
- Targeted manifest tests passed for the global config option and base `OH_RESOLVER_LABEL` env mapping.
- Helm rendering with `--set env.OH_RESOLVER_LABEL=team-label` showed the custom label flowing through the OpenHands env consumers.


replicated/openhands.yaml
```diff
@@ -70,6 +70,7 @@ spec:
       # Enterprise reads the FEATURE_FLAGS-prefixed name; bare name kept for OSS.
       OH_WEB_CLIENT_FEATURE_FLAGS_ALLOW_USER_LLM_CONFIGURATION: 'repl{{ ConfigOptionEquals "allow_user_llm_configuration" "1" }}'
       OH_ALLOW_USER_LLM_CONFIGURATION: 'repl{{ ConfigOptionEquals "allow_user_llm_configuration" "1" }}'
+      OH_RESOLVER_LABEL: 'repl{{ ConfigOption "openhands_resolver_label" }}'
```
jlav: Nit: rather than wire straight into the env var, we should expose this value in the chart under the integrations section, then wire that through to the env var.

jlav (approved):

replicated/openhands.yaml
aivong-openhands: I see, that is a nice and flexible quality of life improvement! Done in 4f97f5fcdc2e06c7f416269af34b48b5775c88e7

PR #843, opened 2026-07-10
ci: PLTF-3193 dispatch development chart bumps after publish

aivong-openhands:
## Why

Stable OpenHands chart publishes currently notify staging but not development, so a newly published chart cannot reach the development environment immediately through the new direct-bump receiver. This adds an independent development dispatch after a successful `openhands` publish, authenticated by the dedicated low-privilege `dev-chart-bump-dispatcher` App and restricted to the verified release actor.

The publish path now checks out the immutable push-event SHA, pins its Actions and Helm inputs, and documents the sender contract and failure semantics.

## Validation

- A fake `gh` execution verified the exact `bump-chart-to-development` payload and that API failures remain fatal.
- All 12 manual sender safety mutants were caught; `zizmor`, a security static analyzer for GitHub Actions workflows, reported no findings.

### DIFF
--- a/.github/workflows/publish-release-charts.yml
+++ b/.github/workflows/publish-release-charts.yml
@@ -49,7 +49,7 @@ jobs:
       - name: Checkout (at the release tag)
         uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
         with:
-          ref: ${{ github.ref_name }}
+          ref: ${{ github.sha }}
           fetch-depth: 0
           persist-credentials: false
 
@@ -60,7 +60,7 @@ jobs:
       - name: Set up Helm
         uses: azure/setup-helm@5119fcb9089d432beecbf79bb2c7915207344b78 # v3
         with:
-          version: 'latest'
+          version: 'v3.18.4'
 
       - name: Lowercase registry owner
         id: owner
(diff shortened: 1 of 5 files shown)


.github/workflows/publish-release-charts.yml
```diff
@@ -47,19 +47,22 @@ jobs:
           echo "Publishing ${COMPONENT} ${VERSION} from charts/${COMPONENT}"
 
       - name: Checkout (at the release tag)
-        uses: actions/checkout@v4
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
```
dylan-openhands, line 50: Question (non-blocking):
Is it expected for these to move to a particular `sha` instead of a version pin?

dylan-openhands (approved):
One small question but otherwise LGTM!

.github/workflows/publish-release-charts.yml
aivong-openhands: Yes version pins are tags that can move so we have settled on sha pins being a better practice!

dylan-openhands: oof that's scary, they _shouldn't_ be doing that but i'm on board for that practice if it's bitten us before!

PR #738, opened 2026-06-18
Always emit LMNR_HTTP_PORT and align chart-default warm LMNR env

ak684:
## Problem (#737): every conversation cold-starts on OHE

The runtime-api warm matcher (`warm_runtime_matches` -> `clean_env_for_match`) does **strict env equality** (keys *and* values), and `LMNR_*` keys are **not** in `KEYS_TO_CLEAN`. So if the request env and a warm pool entry differ on any `LMNR_*` key, the warm runtime is rejected and the conversation cold-starts.

`LMNR_BASE_URL` / `LMNR_FORCE_HTTP` / `LMNR_PROJECT_API_KEY` are always emitted (`=""` when off) on both sides and match. But **PR #692 (`ec2c290`)** wrapped `LMNR_HTTP_PORT` in `charts/openhands/templates/_env.yaml` so the **request side OMITS it** when `laminar.httpPort` is unset, while the **KOTS warm template** (`replicated/openhands.yaml:356`, anchor `&lmnr_http_port` reused at `:625`/`:913`) **still always emits it** (`=""` when analytics off):

```
only_in_warm=['LMNR_HTTP_PORT']  ->  no warm match  ->  cold start
```

## Fix (quick, teammate-endorsed)

Make the `LMNR_*` env **key set identical** on the request side and on **both** warm paths (KOTS and chart-default), in **both** toggle states:

1. **Request side** — `charts/openhands/templates/_env.yaml`: restore `LMNR_HTTP_PORT` to **always-emit** (`{{ if and .Values.laminar.enabled .Values.laminar.httpPort }}…{{ else }}""{{ end }}`), matching the three `LMNR_*` keys directly above it. This reverts #692's key omission.
2. **Chart-default warm config** — `charts/runtime-api/values.yaml` (`warmRuntimes.configs[].environment`) and `charts/openhands/values.yaml` (`runtime-api.warmRuntimes` override): add the full `LMNR_*` set (`=""`). Reverting #692 alone would re-break the chart-default/SaaS path (request `LMNR_HTTP_PORT=""` but chart-default warm had **no** `LMNR_*` keys) — which is what #692 originally papered over. Adding `LMNR_*` here makes the always-emit revert **safe for SaaS**.
3. **KOTS warm template** — `replicated/openhands.yaml` already always-emits all 4 `LMNR_*` via shared anchors (`:354-357` defined, `:623-626` + `:911-914` reused). Verified consistent; left unchanged.

[...]

PR #866, opened 2026-07-13
fix(openhands): mirror agentServerEnv into warm-runtime env so warm pools stay claimable

ak684:
## Problem

On any KOTS install whose LLM provider populates `agentServerEnv` (azure api-key, custom, vertex, bedrock static keys — or `proxy_enabled`), the warm-runtime pool is **never claimable**: the app injects those env vars (e.g. `CUSTOM_API_KEY`/`CUSTOM_API_BASE`, `AZURE_API_KEY`/`AZURE_API_BASE`/`AZURE_API_VERSION`) into every per-session runtime request via `OH_AGENT_SERVER_ENV`, but the `warm-runtimes-config` ConfigMap env is rendered from a separate values path that never sees them. runtime-api claims warm pods by exact env match, so every claim fails with

```
environment: match=False only_in_warm=[] only_in_request=['CUSTOM_API_BASE', 'CUSTOM_API_KEY'] value_differs=[]
No warm runtimes found for API key default with session ID ...
```

and every conversation cold-starts (~1 min). Reported by a customer running the Custom and Azure providers; confirmed from their support bundle (warm CM has 21 infra keys, zero provider keys; all session runtimes carry the AZURE_* trio, the warm runtime doesn't). Same failure class as the LMNR_HTTP_PORT asymmetry fixed in #738 — this closes the whole class for provider creds instead of adding another per-key mirror.

## Fix

Move the canonical agent-server env map to `global.agentServerEnv` so both consumers read one source:

- `replicated/openhands.yaml`: the base CA-bundle map and the five provider/proxy `agentServerEnv` blocks now merge under `global.agentServerEnv`.
- `charts/openhands/templates/_env.yaml`: `OH_AGENT_SERVER_ENV` is assembled from `global.agentServerEnv`, with the legacy `.Values.agentServerEnv` still merged on top for compatibility.
- `charts/openhands/charts/runtime-api/templates/warm-runtimes-configmap.yaml`: each warm entry's `environment` now `mergeOverwrite`s `global.agentServerEnv`, so the warm env always matches what the app sends.

## Verified

- `helm template` subchart: warm CM env picks up `global.agentServerEnv` keys and keeps entry defaults; without global set, output is byte-identical to before.
- `helm template` umbrella: `global.agentServerEnv.CUSTOM_API_*` lands in both `OH_AGENT_SERVER_ENV` and the warm CM in one render; legacy `agentServerEnv` keys still reach `OH_AGENT_SERVER_ENV`.
- Duplicate-key YAML check on `replicated/openhands.yaml`; `helm lint` on both charts.

## Notes

- Provider creds now also appear in the warm ConfigMap. Incremental exposure is limited — the identical values already sit as literal env in the openhands deployment spec (`OH_AGENT_SERVER_ENV`) and in every session runtime pod spec — but worth noting; the warm spec chain (ConfigMap → warm_runtimes.py → V1EnvVar) has no secret-ref support today.
- On key rotation the CM regenerates immediately; warm pods serve the old env until the per-minute warm-runtimes CronJob replaces them (brief self-healing cold-start window).
- Per-user BYOK is unaffected: user keys ride the conversation payload, not the runtime env, so request env stays identical across users.

### DIFF
--- a/charts/openhands/charts/runtime-api/templates/warm-runtimes-configmap.yaml
+++ b/charts/openhands/charts/runtime-api/templates/warm-runtimes-configmap.yaml
@@ -15,6 +15,16 @@
 {{- end }}
 {{- /* Default image for any entry that omits one: the global agent-server image. */ -}}
 {{- $agentServerImage := printf "%s:%s" .Values.global.agentServerImage.repository .Values.global.agentServerImage.tag }}
+{{- /*
+  The app injects global.agentServerEnv (LLM provider creds, proxy vars, CA
+  paths) into every per-session runtime request, and runtime-api claims warm
+  pods only on an exact env match — so those keys must be in the warm env too,
+  or the pool is never claimable and every conversation cold-starts.
+*/ -}}
+{{- $globalAgentEnv := dict }}
+{{- if .Values.global }}
+{{- $globalAgentEnv = (get .Values.global "agentServerEnv") | default dict }}
+{{- end }}
 apiVersion: v1
 kind: ConfigMap
 metadata:
@@ -36,7 +46,7 @@ data:
           "count": {{ $config.count }},
           {{- end }}
           "command": {{ $config.command | toJson }},
-          "environment": {{ $config.environment | toJson }}
+          "environment": {{ mergeOverwrite (deepCopy ($config.environment | default dict)) (deepCopy $globalAgentEnv) | toJson }}
           {{- with $config.run_as_user }},
           "run_as_user": {{ $config.run_as_user }}
           {{- end }}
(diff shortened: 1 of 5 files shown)


ak684:
Live-validated on an embedded-cluster install of this branch's release (llm_provider=custom, warm count 1):
- Deployed `warm-runtimes-config` now carries `CUSTOM_API_BASE`/`CUSTOM_API_KEY`, matching the openhands deployment's `OH_AGENT_SERVER_ENV` exactly.
- A `/start` request with the app-identical spec was claimed by the warm pool in ~130ms: `Claimed warm runtime: lxymahnmseonlpdp`, response `runtime_id=lxymahnmseonlpdp` — the same matcher path that previously rejected every claim with `only_in_request=['CUSTOM_API_BASE','CUSTOM_API_KEY']`.
- Without `global.agentServerEnv` set, rendered output is byte-identical to the current chart (back-compat check).

dylan-openhands (approved):

PR #874, opened 2026-07-13
test: PLTF-1257 add native Helm smoke test with KinD CI

aivong-openhands:
## Why

A native Helm test (`helm.sh/hook: test`) for the OpenHands release, run against ephemeral KinD clusters in CI. The official Helm chart-test framework is post-install verification. What this tier uniquely catches: manifests that render as valid YAML but are rejected by the Kubernetes API, and resources that deploy successfully but are wired together incorrectly at runtime.

The KinD matrix installs the chart with both the ephemeral and persistent values profiles (the same profiles the render contracts assert against) using `helm install --wait --wait-for-jobs`, so a successful install verifies API acceptance and workload/job readiness; the persistent profile also confirms storage provisioning for MinIO, PostgreSQL, and Redis. Each release then runs the native Helm test twice: the test Pod requests `http://openhands-service:3000/health`, verifying in-cluster Service discovery, routing, and application health, and the second run verifies the hook recreates cleanly. The same runner drives local debugging (`ci/run-kind-helm-tests.sh`, documented in `ci/README.md`).

## Validation

The negative-control runs validate the job's two distinct gates: the KinD-backed install checks Kubernetes API acceptance, while the native Helm test checks post-install Service routing and application health.

- [Shared local/CI runner](https://github.com/OpenHands/OpenHands-Cloud/actions/runs/29303946767): the ephemeral and persistent KinD profiles both installed the OpenHands chart and passed the native Helm test using `ci/run-kind-helm-tests.sh`.
- [Broken Service routing](https://github.com/OpenHands/OpenHands-Cloud/actions/runs/29291209011/job/86954990324): proves the native `helm test` catches a deployed release whose valid Kubernetes resources are wired together incorrectly. Kubernetes accepted the release, but the in-cluster health request failed with `Connection refused`.
- [Broken chart installation](https://github.com/OpenHands/OpenHands-Cloud/actions/runs/29291957406/job/86957286919): proves the KinD-backed `helm install` catches Kubernetes API errors that syntactically valid rendered YAML can contain. Kubernetes rejected the unsupported Service type before `helm test` ran.

_This PR was drafted by an AI agent on behalf of the user._

PR #875, opened 2026-07-14
test: PLTF-1257 add OpenHands Helm render contract tests

aivong-openhands:
## Why

Validate rendered manifests against expected output, cover key values combinations, and exercise the schema, `required`, and `fail` safeguards — can only be checked at render time: those guards abort rendering before any resource reaches a cluster, so the official `helm.sh/hook: test` framework (post-install, in-cluster) cannot exercise them. This adds that unit tier as pytest render contracts against `helm template` (Helm 3.21.3), covering Service-to-workload ports, OpenHands probes, storage output, and the safeguard rejections for the ephemeral and persistent profiles. It complements the live-cluster smoke tier in #874 and reuses the same values profiles that job installs, so the fast and KinD checks stay aligned without changing chart runtime behavior.

## Validation

- [The render-contract job](https://github.com/OpenHands/OpenHands-Cloud/actions/runs/29303952793/job/86993388330) completed all 13 cases on the current diff, including both values profiles and the expected rejection of PostgreSQL and Redis schema violations, a missing sandbox gateway hostname, and incomplete Laminar AWS Secrets Manager values.

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -5,8 +5,10 @@ on:
     paths:
       - '.github/workflows/pr-checks.yml'
       - 'charts/**'
+      - 'ci/**'
       - 'replicated/config.yaml'
       - 'replicated/openhands.yaml'
+      - 'scripts/test_openhands_helm_chart_render.py'
       - 'scripts/test_keycloak_realm_template.py'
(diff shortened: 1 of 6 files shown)


aivong-openhands:
We want helm unit tests instead

PR #895, opened 2026-07-14
feat: PLTF-3198 enable the Replicated SDK by default for helm installs

aivong-openhands:
## Why
License enforcement requires the Replicated SDK to run in plain helm installs, not just KOTS/Embedded Cluster. Helm installs are served through the Replicated registry, which injects the customer license at chart-pull time, so flipping `replicated.enabled` to `true` is all the chart needs — no license values are added. The flip also turns on the preflight/support-bundle Secrets for helm installs (they were gated on the same flag), which is what enables troubleshoot CLI support bundles. The postgres/redis support-bundle collectors and analyzers now additionally require their password value to be set, so default helm installs don't render collector URIs with empty credentials (KOTS always sets them). Installs that should not run the Replicated SDK can set `replicated.enabled=false`.

## Validation
- `helm template` with default values renders the SDK subchart's resources and both troubleshoot Secrets (`troubleshoot.sh/kind: preflight` / `support-bundle`); with `replicated.enabled=false` neither renders — locked in as render tests in `scripts/test_troubleshoot_secrets_template.py`, alongside collector-credential gating in both directions.
- Live check on a kind cluster (bootstrapped with the kind harness from the closed #874 — cluster setup, seeded secrets, and kind values reused as-is): installed the current Unstable release (chart 0.22.0) from `oci://registry.replicated.com` with a dev license and `--set replicated.enabled=true`. The SDK pod came up licensed and the instance appeared in the Vendor Portal (`client: Replicated-SDK 1.9.0`, `appStatus: ready`, not KOTS/EC). `support-bundle --load-cluster-specs` discovered the chart and SDK specs and produced a bundle (203 files, all app pod logs + analyzers); `preflight secret/openhands/openhands-preflight` ran and passed. The generated bundle was then uploaded with `support-bundle upload` (authenticating with the license ID) and appeared on the customer's bundles page in the Vendor Portal — note the license needs the support-bundle-upload entitlement enabled for this step.
- In that live run the released chart's postgres/redis analyzers errored on empty-credential URIs — the exact failure mode the collector gating in this PR removes.

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -22,7 +22,9 @@ jobs:
       - name: Set up Helm
         uses: azure/setup-helm@v3
         with:
-          version: 'latest'
+          # 'latest' falls back to helm v3.9.0 when the release lookup is
+          # rate-limited, which is too old to render the replicated SDK chart.
+          version: 'v3.21.3'
       - name: Build chart dependencies
         run: |
           helm repo add lmnr https://lmnr-ai.github.io/lmnr-helm
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -24,7 +24,7 @@ jobs:
         with:
           # 'latest' falls back to helm v3.9.0 when the release lookup is
           # rate-limited, which is too old to render the replicated SDK chart.
-          version: 'v3.21.3'
+          version: 'v3.18.4'
       - name: Build chart dependencies
         run: |
           helm repo add lmnr https://lmnr-ai.github.io/lmnr-helm
--- a/.github/workflows/preview-helm-charts.yml
+++ b/.github/workflows/preview-helm-charts.yml
@@ -95,7 +95,7 @@ jobs:
         if: steps.check.outputs.should_publish == 'true'
         uses: azure/setup-helm@v3
         with:
-          version: 'latest'
+          version: 'v3.18.4'
 
       - name: Compute preview version
         if: steps.check.outputs.should_publish == 'true'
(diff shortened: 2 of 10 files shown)


jpshackelford (approved):
The part of this PR I didn't really understand is this:

> The postgres/redis support-bundle collectors and analyzers now additionally require their password value to be set, so default helm installs don't render collector URIs with empty credentials (KOTS always sets them).

Would you explain? How will this impact helm or vm installs? Are these need in both in cluster and external db and redis installs?

aivong-openhands:
Good question — the gating only affects the two DB health checks inside the support bundle, nothing else.

The troubleshoot v1beta2 `postgresql`/`redis` collectors connect to the database and report health in the bundle. The spec format has no `secretRef`, so credentials must be inlined into the URI string in the Secret — that's the only reason `replicated.postgresUsername/Password/Database` and `replicated.redisPassword` values exist. Only the KOTS installer populates them (templated from config options at deploy time).

Before this PR they were gated on `replicated.enabled` alone, which was effectively a proxy for "this is a KOTS install, so the credentials are set". Flipping `replicated.enabled` on by default breaks that proxy: a default helm install would render `postgresql://:@openhands-postgresql...` — a guaranteed auth failure. (Confirmed in the live kind test against the current release: the bundle's postgres/redis analyzers all errored with "Cannot connect".) Requiring the password restores the old behavior: the collector renders only when it can actually work.

Impact by install type:
- **VM / Embedded Cluster (KOTS)**: no change. The installer always sets the credentials, so the collectors and analyzers render exactly as before.
- **Helm**: the DB collectors are omitted until `replicated.postgres*`/`redisPassword` are provided in values; everything else in the bundle (pod logs, cluster resources, all other analyzers) is unaffected. Setting the values opts the DB health checks in.

On in-cluster vs external: these collectors only ever target the **in-cluster** databases — the URIs are hardcoded to the chart's service hostnames (`<release>-postgresql`, `<release>-redis-master`) and each is additionally gated on `postgresql.enabled`/`redis.enabled` (the in-chart DBs). Installs using an external postgres never rendered them, before or after this PR. External-DB health checks would be a separate addition if we want them.

PR #917, opened 2026-07-16
feat: add root ingress redirect support

malhotra5:
## Summary
- Add `ingress.redirects` values for redirecting legacy root ingress hostnames.
- Render a Traefik `Middleware` plus redirect-only `Ingress` per configured redirect host.
- Add `ingress.authHost` so deployments can keep auth on an existing hostname while moving the canonical app host.
- Add Helm unit coverage for redirect rendering.

## Why
The staging domain migration needs the Helm chart to own the app ingress/redirect behavior. `saas-deploy` can then consume this as preview chart `0.23.2-alpha.917` and configure `staging.openhands.dev` as canonical while redirecting `staging.all-hands.dev`.

## Validation
- `helm dependency update charts/openhands`
- `helm unittest charts/openhands`
- Rendered `templates/ingress-root-redirects.yaml` with sample redirect values.
- Rendered deployment env with `ingress.authHost` and confirmed `AUTH_WEB_HOST` uses the override.

_This PR was updated by an AI agent (OpenHands) on behalf of the user._

### DIFF
--- a/charts/openhands/templates/ingress-root-redirects.yaml
+++ b/charts/openhands/templates/ingress-root-redirects.yaml
@@ -1,0 +1,41 @@
+{{- if and .Values.enabled .Values.ingress.enabled .Values.ingress.redirects }}
+{{- range $redirect := .Values.ingress.redirects }}
+{{- $name := required "ingress.redirects[].name is required" $redirect.name }}
+{{- $host := required (printf "ingress.redirects[%s].host is required" $name) $redirect.host }}
+{{- $to := required (printf "ingress.redirects[%s].to is required" $name) $redirect.to }}
+{{- $resourceName := printf "openhands-%s-redirect" $name | trunc 63 | trimSuffix "-" }}
+---
+apiVersion: traefik.io/v1alpha1
+kind: Middleware
+metadata:
+  name: {{ $resourceName }}
+spec:
+  redirectRegex:
+    regex: {{ printf "^https://%s(.*)" ($host | replace "." "\\.") | quote }}
+    replacement: {{ $to | quote }}
+    permanent: {{ ternary $redirect.permanent true (hasKey $redirect "permanent") }}
+---
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: {{ $resourceName }}
+  annotations:
+    traefik.ingress.kubernetes.io/router.middlewares: {{ printf "%s-%s@kubernetescrd" $.Release.Namespace $resourceName | quote }}
+    {{- with $redirect.annotations }}
+    {{ . | toYaml | nindent 4 }}
+    {{- end }}
+spec:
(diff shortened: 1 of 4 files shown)


charts/openhands/templates/ingress-root-redirects.yaml
```diff
@@ -0,0 +1,41 @@
...
+---
+apiVersion: traefik.io/v1alpha1
+kind: Middleware
+metadata:
+  name: {{ $resourceName }}
+spec:
+  redirectRegex:
+    regex: {{ printf "^https://%s(.*)" ($host | replace "." "\\.") | quote }}
+    replacement: {{ $to | quote }}
+    permanent: {{ ternary $redirect.permanent true (hasKey $redirect "permanent") }}
```
jlav, line 16: Our charts should be agnostic about the ingress that our consumers are using, so we should avoid baking ingress-related CRDs into the charts. I know there may be some places where we assume traefik or create CRDs already, but for our future selves we're trying to move away from that.

We're adding this redirect for a concern that only exists in our environment, not necessarily the environment of our consumers, so my recommendation is to create the middleware ourselves through https://github.com/OpenHands/saas-deploy

You can add any additional resources you want to render along with each app into each environment. For example, you could put this middleware in `app/openhands/environments/staging/templates/middleware.yaml`, and it will get deployed.

charts/openhands/values.yaml
```diff
@@ -208,6 +208,8 @@ ingress:
   enabled: false
   # REQUIRED: Update to a hostname in a DNS domain you own
   # host: "app.example.com"
+  # Optional override when auth is served on a different hostname than auth.<host>.
+  authHost: ""
```
jlav, line 212: Absent of `ingress-root-redirects.yaml`, I don't think this configuration belongs under the `ingress` value.

A really awesome change I would encourage, though, is to break out a separate `auth` key in the values file and then refactor all the conditional logic that surrounds `AUTH_WEB_HOST` so that the auth host is just a templated value that the consumer plugs in and we don't try to derive all the possible configurations based on other values set in the charts.

As much as we can, we want to try to keep conditional logic out of the charts in favor of just allowing consumers to template in the specific values that they need for their cluster.

jlav (changes requested):
Sorry to request changes on this one. I recognize that a lot of this is following existing patterns, but I'm trying to stop the bleeding that is making these charts increasingly difficult to maintain.

Following my recommendation to define the redirect middleware in https://github.com/OpenHands/saas-deploy, you may find that you can make a very minimal change - or maybe even no change at all - in the charts to accomplish what you're looking for.

malhotra5:
Closing this after review feedback. The Traefik redirect and staging-specific routing have been moved into `OpenHands/saas-deploy` instead, so the generic OpenHands Cloud chart no longer needs these changes. See OpenHands/saas-deploy#268.

_This comment was posted by an AI agent (OpenHands) on behalf of the user._

PR #926, opened 2026-07-20
feat: Added environment variable for RUNTIME_API_BASE_URL

tofarr:
In most environments, it is the same as RUNTIME_BASE_URL - It differs in our staging, dev and prod. 
e.g. for my-openhands.com, with derived URLs:
* `RUNTIME_API_BASE_URL=https://runtime.my-openhands.com`
* `RUNTIME_BASE_URL=runtime.my-openhands.com` (So runtimes will be at https://<RUNTIME_ID>.runtime.my-openhands.com)

For our prod:
* `RUNTIME_API_BASE_URL=https://runtime.app.all-hands.dev`
* `RUNTIME_BASE_URL=prod-runtime.all-hands.dev` (So runtimes will be at https://<RUNTIME_ID>.prod-runtime.all-hands.dev)

## Helm Chart Checklist

### DIFF
--- a/replicated/openhands.yaml
+++ b/replicated/openhands.yaml
@@ -314,7 +314,7 @@ spec:
         DB_USER: '{{repl if ConfigOptionEquals "postgres_type" "external_postgres"}}{{repl ConfigOption "external_postgres_username"}}{{repl else}}postgres{{repl end}}'
         DB_NAME: "runtime_api_db"
         RUNTIME_BASE_URL: '{{repl if ConfigOptionEquals "hostname_mode" "derive"}}runtime.{{repl ConfigOption "base_domain"}}{{repl else}}{{repl ConfigOption "runtime_base_hostname"}}{{repl end}}'
-        RUNTIME_API_BASE_URL: '{{repl if ConfigOptionEquals "hostname_mode" "derive"}}runtime.{{repl ConfigOption "base_domain"}}{{repl else}}{{repl ConfigOption "runtime_base_hostname"}}{{repl end}}'
+        RUNTIME_API_BASE_URL: '{{repl if ConfigOptionEquals "hostname_mode" "derive"}}https://runtime.{{repl ConfigOption "base_domain"}}{{repl else}}{{repl ConfigOption "runtime_base_hostname"}}{{repl end}}'
         RUNTIME_IMAGE_PULL_SECRETS: '{{repl if and (ConfigOptionEquals "custom_sandbox_image_enabled" "1") (ne (ConfigOption "custom_sandbox_image_registry_server") "") (ne (ConfigOption "custom_sandbox_image_registry_username") "") (ne (ConfigOption "custom_sandbox_image_registry_password") "") }}sandbox-image-pull-secret{{repl else}}{{repl ImagePullSecretName }}{{repl end}}'
         STORAGE_CLASS: "openebs-hostpath"
         INGRESS_CLASS: "traefik"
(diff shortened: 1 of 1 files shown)


hieptl (approved):
Thank you! 🙏

simonrosenberg (approved):

replicated/openhands.yaml
```diff
@@ -314,6 +314,7 @@ spec:
         DB_USER: '{{repl if ConfigOptionEquals "postgres_type" "external_postgres"}}{{repl ConfigOption "external_postgres_username"}}{{repl else}}postgres{{repl end}}'
         DB_NAME: "runtime_api_db"
         RUNTIME_BASE_URL: '{{repl if ConfigOptionEquals "hostname_mode" "derive"}}runtime.{{repl ConfigOption "base_domain"}}{{repl else}}{{repl ConfigOption "runtime_base_hostname"}}{{repl end}}'
+        RUNTIME_API_BASE_URL: '{{repl if ConfigOptionEquals "hostname_mode" "derive"}}https://runtime.{{repl ConfigOption "base_domain"}}{{repl else}}{{repl ConfigOption "runtime_base_hostname"}}{{repl end}}'
```
tofarr, line 317: 🔴 Critical: `https://` is currently inside the `derive` branch only, so in `custom` mode the rendered value is the bare hostname from `runtime_base_hostname` (e.g. `runtime.example.com`) with no scheme. That value then gets fed into `f'{RUNTIME_API_BASE_URL}/runtimes/{id}/storage'` in `OpenHands/runtime-api/runtimes.py`, producing a scheme-less URL that HTTP clients will reject as relative — silently breaking the fusey broker whenever the operator uses `hostname_mode=custom` (which is the default for self-hosted installs).

The config option `runtime_base_hostname` in `replicated/config.yaml` is documented as a hostname (default `"runtime.example.com"`, no scheme), and the existing `RUNTIME_URL_PATTERN` on line 968 already encodes the correct pattern — `https://` **outside** the conditional so both branches get a scheme:

```suggestion
        RUNTIME_API_BASE_URL: 'https://{{repl if ConfigOptionEquals "hostname_mode" "derive"}}runtime.{{repl ConfigOption "base_domain"}}{{repl else}}{{repl ConfigOption "runtime_base_hostname"}}{{repl end}}'
```

(Equivalent fix: also put `https://` in the `else` branch, matching the `openhandsApiUrl`/`automationBaseUrl` pattern on lines 633–634. The above is preferable because it stays symmetric with `RUNTIME_URL_PATTERN` right below it in the same env block.)

tofarr (commented):
Thanks for wiring up the new env var! The motivation is sound — runtime-api consumes `RUNTIME_API_BASE_URL` as a full URL (see `OpenHands/runtime-api/runtimes.py` at the `broker_url = f'{RUNTIME_API_BASE_URL}/runtimes/{workspace_runtime_id}/storage'` line) and in our prod/staging/dev the runtime-api host differs from the per-runtime hostname pattern. However the current template only emits the `https://` prefix in the `derive` branch, which silently produces a host-without-scheme in the `custom` branch and breaks the URL construction. See the inline comment for the fix.

A couple of secondary notes:

- The `## Helm Chart Checklist` in the description is misleading — this PR only touches `replicated/openhands.yaml`, not any Helm chart. The checklist items should be removed (or replaced with a Replicated-manifest equivalent).
- No tests were added. The existing scripts under `scripts/` (e.g. `test_replicated_resolver_label.py`) don't cover env-var wiring, so there's nothing to extend, but it would be worth a small assertion (alongside the other test scripts) that `RUNTIME_API_BASE_URL` starts with a scheme under both `hostname_mode` values — that would have caught this bug.

PR #931, opened 2026-07-20
feat: enable the pending-runtime reaper on OHE installs

ak684:
Turns on the runtime pending-runtime reaper for OHE installs, and pins runtime-api to the first release that has it.

- Bumps the runtime-api image from `0.5.2` to `0.5.3` (runtime-api#655, the reaper, shipped in that release).
- Sets `RUNTIME_PENDING_REAP_ENABLED: "true"` in `replicated/openhands.yaml` (Replicated installer only, since OHE clusters are fixed-size with no autoscaler).
- Keeps it `false` in `charts/openhands/values.yaml` so it never reaches SaaS, which is autoscaled and shares this umbrella chart via saas-deploy (per review).

With both, an OHE install reaps sandbox pods stuck Pending/Unschedulable instead of letting them pile up (the Verizon issue, PLTF-3281). Effective reap is bounded by the idle grace: about 60 min for a new runtime on OHE, ~10 min for a resumed one, versus days today. This rides the next OHE release.

### DIFF
--- a/charts/openhands/values.yaml
+++ b/charts/openhands/values.yaml
@@ -819,6 +819,9 @@ runtime-api:
     CPU_REQUEST: "500m"
     # Default; KOTS overrides via the sandbox_ephemeral_storage_size ConfigOption.
     EPHEMERAL_STORAGE_SIZE: "10Gi"
+    # OHE clusters are fixed-size (no autoscaler), so an unschedulable sandbox
+    # never schedules — reap it instead of letting it pile up until dead_seconds.
+    RUNTIME_PENDING_REAP_ENABLED: "true"
 
 jira:
   enabled: false
--- a/charts/openhands/values.yaml
+++ b/charts/openhands/values.yaml
@@ -819,9 +819,11 @@ runtime-api:
     CPU_REQUEST: "500m"
     # Default; KOTS overrides via the sandbox_ephemeral_storage_size ConfigOption.
     EPHEMERAL_STORAGE_SIZE: "10Gi"
-    # OHE clusters are fixed-size (no autoscaler), so an unschedulable sandbox
-    # never schedules — reap it instead of letting it pile up until dead_seconds.
-    RUNTIME_PENDING_REAP_ENABLED: "true"
+    # Off here so it never reaches SaaS (this umbrella chart is shared, and SaaS
+    # is autoscaled where an unschedulable pod is what triggers a node). The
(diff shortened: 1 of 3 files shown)


charts/openhands/values.yaml
```diff
@@ -812,6 +812,9 @@ runtime-api:
     MEMORY_LIMIT: "3072Mi"
     CPU_REQUEST: "500m"
     EPHEMERAL_STORAGE_SIZE: "10Gi"
+    # OHE clusters are fixed-size (no autoscaler), so an unschedulable sandbox
+    # never schedules — reap it instead of letting it pile up until dead_seconds.
+    RUNTIME_PENDING_REAP_ENABLED: "true"
```
aivong-openhands: Since the chart default is true, we may need to set this to off for SaaS?

ak684: Good one to check, but I don't think it reaches SaaS. This is the OHE (Replicated) umbrella chart, and SaaS deploys runtime-api from `runtime-api/deploy/chart` instead (that's what `deploy/.github/workflows/deploy-eval-runtime.yaml` points at). There the reaper stays off by default, since `RUNTIME_PENDING_REAP_ENABLED` unset falls back to `false` in code. So this `"true"` only hits OHE installs, which are the fixed-size clusters we actually want it on for. Let me know if there's a SaaS path I'm not seeing though.

jlav: Let's set this to false here and then only set it to true inside of the Replicated `openhands.yaml` file. That's the most straightforward place to set a value that we want to be applied for the Replicated installer only.

aivong-openhands: We deploy runtime API with the openhands umbrella chart in this repo for SaaS environments https://github.com/OpenHands/saas-deploy/blob/main/app/runtime-api/environments/staging/Chart.yaml#L9

ak684: Ah, you're both right, thanks. I had the SaaS path backwards, `saas-deploy` does pull in this umbrella chart so the chart default would have reached SaaS. Went with jlav's suggestion: `false` here in the chart so SaaS stays off (it's autoscaled), and `RUNTIME_PENDING_REAP_ENABLED: "true"` in `replicated/openhands.yaml` so it only applies to the Replicated installer. Pushed.

aivong-openhands (approved):

PR #937, opened 2026-07-21
fix(openhands): namespace-qualify bundled litellm url for sandboxes

### LINKED ISSUE #932: charts/openhands: namespace-qualify bundled in-cluster URLs so runtimes can run in a separate namespace
## Summary

Running agent runtimes (sandboxes) in a dedicated namespace, separate from the chart's control-plane workloads, is blocked: the agent can't reach the bundled LiteLLM from another namespace because the chart hands it a bare Service name that only resolves in the release namespace.

## Details

The runtime namespace is already selectable via `runtime-api.env.K8S_NAMESPACE`, and placement works — the sandbox's Deployment, Pod, Service, Ingress, and PVC all land in the target namespace and the cluster-wide RBAC covers it.

But the agent is given the bare LiteLLM URL `http://<release>-litellm:4000` (`charts/openhands/templates/_env.yaml:283`). A bare Service name only resolves inside the caller's own namespace, so from a sandbox in a different namespace DNS fails and every conversation errors with a LiteLLM connection error.

Reproduced on a local kind install with `runtime-api.env.K8S_NAMESPACE` pointed at a dedicated namespace: the sandbox came up Ready there, but the agent's LLM calls failed with `Cannot connect to host <release>-litellm:4000 [Name or service not known]`. Adding a namespace-scoped DNS alias so the bare name resolved to the real LiteLLM Service produced successful completions — confirming the bare-name resolution is the only blocker. Same-namespace installs (today's default) resolve the bare name and are unaffected.

## Why it matters

Sandboxes execute untrusted agent-authored code, so isolating them from the app, database, Keycloak, and secrets is a sensible posture. This gap blocks recommending a dedicated runtime namespace in the Helm install docs; until it's fixed, that recommendation would require a per-namespace DNS-alias workaround.

## Proposed fix

Namespace-qualify the bundled URL — `charts/openhands/templates/_env.yaml:283` → `http://<release>-litellm.<release-namespace>.svc.cluster.local:4000`. It resolves from any namespace and is a no-op for same-namespace installs.

It has to be a chart default rather than a values override, because the URL is persisted into each user's LLM profile at first use — overriding it later doesn't fix existing profiles, so a fresh install needs to bake in a resolvable URL from the start. LiteLLM is the confirmed case; other bundled in-cluster URLs handed to sandboxes should get the same audit.

_This issue was filed by an AI agent on behalf of the user._

jlav:
## Description

Fixes #932. Running agent sandboxes in a dedicated namespace, separate from the control-plane workloads, was blocked. The app hands each sandbox the bundled litellm URL as a bare service name (`http://<release>-litellm:4000`), which only resolves inside the caller's own namespace. From a sandbox in another namespace, DNS fails and every conversation errors with a litellm connection error.

This namespace-qualifies that URL to an FQDN (`http://<release>-litellm.<release-namespace>.svc.cluster.local:4000`) so it resolves from any namespace. It's a no-op for same-namespace installs (today's default).

It has to be a chart default rather than a values override, because the URL is persisted into each user's LLM profile at first use. Overriding it later wouldn't fix existing profiles, so a fresh install needs a resolvable URL baked in from the start. This only affects newly-persisted profiles - I didn't touch data that's already been written.

While here, I promoted the sandbox namespace to a proper `runtime-api.sandbox_namespace` value. It was a static `K8S_NAMESPACE: "openhands"` buried in the `runtime-api.env` passthrough map. Since values files aren't templated, that literal could never track a differently-named release namespace. The new value defaults to the release namespace and is the clean knob for pointing sandboxes at a dedicated namespace. I named it `sandbox_namespace` rather than `namespace` so it's clear it controls where sandboxes are created, not where runtime-api itself runs.

## Approach

`K8S_NAMESPACE` is now emitted from `sandbox_namespace | default .Release.Namespace`, guarded the same way as `DB_SSL_MODE` so a legacy `env.K8S_NAMESPACE` override still wins with no double-emission. For the standard deploy (release namespace `openhands`) this resolves to `openhands`, matching the previous hardcoded default.

I checked the two other bare `<release>-...` service URLs in that app env block - `REDIS_HOST` and the Postgres `DB_HOST`. Neither is handed to sandboxes, and per the isolation goal a sandbox shouldn't reach redis or postgres at all, so litellm was the only URL that needed qualifying.

## Helm Chart Checklist

## Additional Notes

No breaking changes and no new required values - `sandbox_namespace` defaults to the release namespace, so nothing needs README changes. `helm lint` passes. I verified backwards compat by rendering: default release namespace resolves `K8S_NAMESPACE` to `openhands` (prior behavior), a legacy `env.K8S_NAMESPACE` override still wins, and `--set runtime-api.sandbox_namespace=<ns>` places sandboxes and the litellm FQDN in the expected namespace.

### DIFF
--- a/charts/openhands/charts/runtime-api/templates/_env.yaml
+++ b/charts/openhands/charts/runtime-api/templates/_env.yaml
@@ -12,6 +12,13 @@
 - name: PGSSLMODE
   value: {{ $dbSslMode | quote }}
 {{- end }}
+{{- if not (hasKey .Values.env "K8S_NAMESPACE") }}
+# Namespace runtime-api creates sandboxes in (not where runtime-api itself
+# runs). Defaults to the release namespace (sandbox_namespace empty); set
+# sandbox_namespace to create sandboxes in a dedicated one.
+- name: K8S_NAMESPACE
+  value: {{ .Values.sandbox_namespace | default .Release.Namespace | quote }}
+{{- end }}
 - name: RUNTIME_IDLE_SECONDS
   value: {{ .Values.cleanup.idle_seconds | default 1200 | quote }}
 - name: RUNTIME_DEAD_SECONDS
--- a/charts/openhands/charts/runtime-api/values.yaml
+++ b/charts/openhands/charts/runtime-api/values.yaml
@@ -27,6 +27,13 @@ serviceAccount:
 runtimeInSameCluster: false
 defaultAPIKeySecretName: default-api-key
 
+# Namespace runtime-api creates sandbox pods (and their Service, Ingress, PVC)
+# in, surfaced to the pod as K8S_NAMESPACE. This is NOT where runtime-api
+# itself runs. Empty defaults to the release namespace. Point it at a dedicated
+# namespace to isolate untrusted sandbox workloads from the control plane.
+# Ignored if env.K8S_NAMESPACE is set.
+sandbox_namespace: ""
+
 # Mount a trust-manager / user-supplied CA bundle into the runtime-api pod and
 # point the Python TLS stack (httpx + requests) at it. Required when the
 # runtime ingress serves a cert signed by a CA that isn't in the OS trust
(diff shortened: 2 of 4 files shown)


aivong-openhands (approved):
Thank you!

PR #942, opened 2026-07-21
revert(openhands): PLTF-3258 remove ingress.host fail guard (#923)

aivong-openhands:
## Why

This reverts #923, which added a render-time guard in `validations.yaml` failing the install/upgrade when `ingress.enabled: true` but `ingress.host` is empty.

The guard rejects a value combination that was previously accepted, so it is a breaking change for any consumer that renders with `ingress.enabled: true` and no host set. Reverting restores the prior render behavior while the guard is reworked.

## Validation

Rendered with `helm template`:

- `charts/openhands/templates/validations.yaml` removed, restoring the tree to its pre-#923 state
- ingress on, no host → renders again instead of failing
- ingress off (default) and ingress on + host set → unchanged

### DIFF
--- a/charts/openhands/templates/validations.yaml
+++ b/charts/openhands/templates/validations.yaml
@@ -1,8 +1,0 @@
-{{/*
-Render-time configuration guards. Emits no resources; each block fails the
-render with a clear message when values are invalid or conflicting, so the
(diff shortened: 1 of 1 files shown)


aivong-openhands:
Closing as the alternative solution is to look into syncing canvas in staging and prod

PR #945, opened 2026-07-21
docs(assets): PLTF-3286 add automations auth-flow diagram

aivong-openhands:
## Why
The Automations UI/API runs at a subpath (`/automations`, `/api/automation`) on the **same host** as the main OpenHands app, and how an already-logged-in user gets authenticated there is undocumented and non-obvious: there is no second login and no token hand-off — the browser just sends the same-origin `keycloak_auth` session cookie, and the Automations service (which cannot validate that cookie itself) forwards it verbatim to the Enterprise Server's `GET /api/v1/users/me` and trusts the identity that comes back. This adds a sequence diagram (`.mmd` + rendered PNG) capturing that flow, in the same style as the existing `docs/assets/*-oauth-flow` diagrams, and wires it into `docs/build-diagrams.sh` so it re-renders alongside them. The diagram uses `BASE_DOMAIN` placeholders only — no environment-specific hosts, IPs, or secrets.

## Validation
- Verified every step of the diagram against a live install (instance-1, v0.28.0): the `keycloak_auth` cookie shape (HS256 JWS, `access_token`/`refresh_token`/`accepted_tos` payload, 1-week expiry, `Domain=app.<BASE_DOMAIN>` `Path=/`); the SPA being public (`GET /automations/` with no cookie → 200); the delegated API auth (no cookie → 401 *"Authentication required…"*, bogus cookie → 401 *"Invalid or expired session cookie"*, valid session → 200); and the Enterprise Server receiving the forwarded `GET /api/v1/users/me`.
- Rendered the PNG via `docs/build-diagrams.sh` (`mmdc`); the `.mmd` parses and renders cleanly.


lilagrc (approved):
🎉

PR #900, opened 2026-07-14
feat: mount Agent Canvas under /canvas

malhotra5:
## Summary

- Change the Agent Canvas umbrella ingress from dedicated-host mode to subpath mode on the main OpenHands host.
- Remove `agent-canvas.ingress.host` and dedicated Canvas TLS secret values from chart defaults.
- Pass `--base-path /canvas` to the Agent Canvas static server deployment.
- Update the Agent Canvas chart README examples for `/canvas` deployment.

## Related

- Requires the Agent Canvas base-path support from https://github.com/OpenHands/agent-canvas/pull/1796.

## Validation

- Reviewed rendered template logic and searched for stale Canvas subdomain values.
- Attempted `helm template`, but Helm is not installed in this environment.

This PR was created by an AI agent (OpenHands) on behalf of the user.

@malhotra5 can click here to [continue refining the PR](https://app.all-hands.dev/conversations/9a6cc2b2-2b39-4126-99c6-c2f8f057bb91)

PR #954, opened 2026-07-22
feat: wire Agent Canvas through Replicated/Helm installs

lilagrc:
HUMAN:
Manually verified this in shared-3 replicated instance. Agent Canvas loads when navigating to `/canvas`, and prompts a user through the agent/llm onboarding flow (mirroring exactly the behavior in saas/cloud. Verified that a conversation in UI is present in the other. Agent Canvas is turned on by default, confirmed that is desired with @jpelletier1. 

## Summary

Agent Canvas (the `/canvas` UI) was added to the Helm chart in #900 — the `agent-canvas` subchart, the `templates/ingress-agent-canvas.yaml` umbrella ingress, and an `agent-canvas:` block in `charts/openhands/values.yaml`. However, that PR never touched `replicated/`, so the feature was **never wired into the Replicated (KOTS) release**. Replicated installs (and Helm installs derived from the KOTS values) are driven by `replicated/openhands.yaml`, which had zero agent-canvas configuration, so the subchart stayed at its `enabled: false` default and never deployed. There was also no config toggle, no status informers, and no airgap image-bundling entry.

A latent bug compounded this: the subchart's `image.tag` defaulted to `""`, so even a bare `agent-canvas.enabled: true` would render an invalid image reference (`ghcr.io/openhands/agent-canvas:`) and fail to pull.

This PR wires Agent Canvas all the way through to Replicated/Helm customers.

## Changes

- **`charts/openhands/charts/agent-canvas/values.yaml`** — pin a real default `image.tag` to the semver release `1.5.2` (agent-canvas publishes release-tagged images via Release Please) so the chart is deployable on its own. Environments can drop their image-tag override and fall back to this chart default.
- **`replicated/config.yaml`** — add an `agent_canvas_configuration` group with an `agent_canvas_enabled` bool, defaulted **on** (`"1"`) so Agent Canvas is on by default, which was validated by Product.
- **`replicated/openhands.yaml`** —
  - add an `optionalValues` block (gated on `agent_canvas_enabled`) that enables the subchart, points the image at the licensed proxy registry, adds the image pull secret, enables the ingress on the **primary app host** at `/canvas` with the existing `openhands-tls` secret, and sets `staticServer.lockToCloud` to the instance's own app URL (so the Canvas UI targets the customer's backend, not all-hands.dev);
  - add `agent-canvas` to the `builder` block for airgap image bundling.
- **`replicated/application.yaml`** — add conditional `statusInformers` for the agent-canvas deployment, service, and ingress.
- **`charts/openhands/tests/agent_canvas_ingress_test.yaml`** — new helm-unittest suite covering disabled cases, `/canvas` routing, TLS on/off, and the required-host failure.

## Validation

- `helm unittest charts/openhands` -> **31 passed** (including 5 new agent-canvas tests).
- `make manifests` builds all Replicated manifests cleanly with the canvas wiring present.
- All edited `replicated/*.yaml` files parse as valid YAML.
- Subchart deployment renders a valid image (`ghcr.io/openhands/agent-canvas:1.5.2`) and correct `--base-path` / `--lock-to-cloud` / `--auth-required` args.
- Published to an isolated branch-named Replicated channel (`wire-agent-canvas-replicated`, release sequence 1291) to validate the build + `replicated release lint` in the real pipeline.

## Notes

- **Same-host routing**: Canvas shares the app host and relies on Traefik longest-prefix routing (`/canvas` -> canvas, `/` -> app), matching production `app.all-hands.dev/canvas`.

---


lilagrc (commented):
## Initial code review (via `/codereview` repo skill)

Reviewed against the repo's `code-review` guidelines. Overall this follows the established Replicated wiring patterns well — the `optionalValues` block, `statusInformers`, `builder` entry, and config toggle all mirror `plugin-directory` and `automation`. No blocking (important+) issues found. A few points worth a decision before this leaves draft:

### 1. Default-on vs. opt-in (guideline #7: "Defaults serve the self-hosted user") — most substantive
`agent_canvas_enabled` defaults to `"1"`, whereas the comparable feature toggles `plugin_directory_enabled` and `automations_enabled` both default to `"0"`. Enabling a relatively new preview UI by default for every self-hosted/Replicated user is a notable posture change. Unless there's a product decision to ship Canvas on by default, defaulting to `"0"` would be more consistent with the other optional components and safer for self-hosted users. (Flagged in the PR description as an open question.)

### 2. TLS secret name is inconsistent with the other chart ingresses
The Canvas ingress uses `secretName: openhands-tls`, but every other ingress in the `openhands` chart (`ingress-root`, `-mcp`, `-automation`, `-integrations-hub`, `-plugin-directory`) uses `app-all-hands-{{ .Values.tls.env }}-tls`. Both secrets are created with the same cert by `openhands-secrets/templates/tls-certificate.yaml`, so this is functional, not a bug — but the inconsistency is a small future-maintenance trap. Consider aligning Canvas with the same `app-all-hands-<env>-tls` secret the sibling ingresses use. (The agent-canvas ingress template takes an explicit `secretName`, so this is purely a value choice in `replicated/openhands.yaml`.)

### 3. Pinned `sha-` image tag will go stale
The replicated block overrides `image.repository` but intentionally inherits `image.tag` from the subchart default (`sha-ef3b28a`). Because agent-canvas only publishes `sha-`/`pr-` tags and there's no image-tag-bump workflow wiring it into this repo, this pin will drift over time unless the deploy workflow always overrides it or a bump automation is added. Worth confirming the intended source of truth for the Canvas tag and whether a follow-up (semver releases or an auto-bump caller) should track it.

### 4. Config grouping (minor)
`plugin_directory_enabled` lives under the **Experimental** group ("Preview features … Enable at your own risk"). If Canvas is still preview-quality, placing `agent_canvas_enabled` there rather than in a new dedicated group may set clearer expectations; if it's considered GA, a standalone group is fine.

### Nits
- None blocking. The comments on the `optionalValues` block and image tag are clear and helpful.

Tests and build validation look solid (`helm unittest` 31 passed incl. the new ingress suite; `make manifests` builds cleanly).

---
This review was generated by an AI agent (OpenHands) on behalf of the user.

charts/openhands/charts/agent-canvas/values.yaml
```diff
@@ -9,8 +9,11 @@
 
 image:
   repository: ghcr.io/openhands/agent-canvas
-  # Tag is supplied at deploy time via the OpenHands deploy workflow.
-  tag: ""
+  # Default tag so the chart is deployable on its own (helm/replicated
+  # installs). The OpenHands Cloud deploy workflow overrides this at deploy
+  # time. agent-canvas only publishes sha-/pr- tags, so this tracks a known
+  # good main build rather than a semver release.
+  tag: "sha-ef3b28a"
```
aivong-openhands: The comment seems out of date as agent canvas is set up with Release Please so it does publish release tagged images https://github.com/OpenHands/agent-canvas/releases 🤔 I recommend getting a semver release tag here so that we can start shipping the agent canvas chart with a release tag. We can then remove the agent canvas image tag override in environments like prod https://github.com/OpenHands/saas-deploy/blob/main/app/openhands/environments/production/version.yaml#L9 and instead use the chart default

replicated/config.yaml
```diff
@@ -1143,6 +1143,16 @@ spec:
...
 
+    - name: agent_canvas_configuration
+      title: Agent Canvas
+      description: Configure the Agent Canvas UI
+      items:
+        - name: agent_canvas_enabled
+          title: Enable Agent Canvas
+          help_text: Deploy the Agent Canvas UI, served at /canvas on your application hostname. The UI is locked to this instance's OpenHands backend.
+          type: bool
+          default: "1"
```
aivong-openhands, line 1206: As the code review suggested, I do recommend setting the default to off for an opt in. In case there are issues when enabling the feature, fewer customers are impacted since it's an opt in like automations

charts/openhands/charts/agent-canvas/values.yaml
lilagrc: Good catch, thanks — the comment was stale. Fixed in 7be3514: the default `image.tag` now points at the semver release `1.5.2` (the ghcr tag has no `v` prefix, matching the other subcharts like `1.2.0`/`1.1.5`), and the comment now explains it tracks a published Release Please release. This should let prod/staging drop their agent-canvas image-tag override and fall back to the chart default.

_Reply by an AI agent (OpenHands) on behalf of the user._

replicated/config.yaml
lilagrc: Agreed — changed the default to `"0"` in 7be3514 so Agent Canvas is opt-in, matching the automations pattern. Fewer customers are impacted if we hit issues after enabling.

_Reply by an AI agent (OpenHands) on behalf of the user._

lilagrc:
Thanks for the early feedback, @aivong-openhands !

aivong-openhands:
> Thanks for the early feedback, @aivong-openhands !

No problem!

replicated/config.yaml
```diff
@@ -1195,6 +1195,16 @@ spec:
...
```
lilagrc, line 1206: Default is `"1"` (on) here. This was set to `"0"` in `7be3514` per the opt-in review feedback and then flipped back to `"1"` in `d0eacab`. If on-by-default is the intended final decision, please reconcile the PR description (which still says "defaulted off, matching the automations pattern") and the resolved review thread, since `automations_enabled` defaults to `"0"` — so this no longer matches that pattern.

_AI-generated comment (OpenHands)._

replicated/openhands.yaml
```diff
@@ -1123,6 +1123,32 @@ spec:
...
+            enabled: true
+            host: '{{repl if ConfigOptionEquals "hostname_mode" "derive"}}app.{{repl ConfigOption "base_domain"}}{{repl else}}{{repl ConfigOption "app_hostname"}}{{repl end}}'
+            className: traefik
+            path: /canvas
+            tls:
+              enabled: true
+              secretName: openhands-tls
+          staticServer:
+            basePath: /canvas
+            authRequired: false
```
lilagrc, line 1150: `authRequired: false` overrides the subchart's `authRequired: true` default, so the static bundle is served unauthenticated at `/canvas` on the app host. Assuming the backend still enforces auth and no secrets ship in the static bundle, this is fine — but a short inline comment explaining why auth is intentionally disabled here (unlike the subchart default) would help future readers.

_AI-generated comment (OpenHands)._

lilagrc (commented):
## Code review — Agent Canvas wiring for Replicated/Helm

Overall this is a clean, well-scoped change that follows the established Replicated wiring patterns nicely:

- The `optionalValues` block gated on `agent_canvas_enabled` mirrors the automations/plugin-directory pattern (licensed proxy repository, `ImagePullSecretName`, `hostname_mode` derive/explicit hostname).
- The `statusInformers` names (`deployment/agent-canvas`, `service/agent-canvas`, `ingress/openhands-agent-canvas-ingress`) match the subchart's actual resource names.
- Pinning the subchart `image.tag` to the semver release `1.5.2` (no `v` prefix, consistent with other subcharts) fixes the real latent bug where a bare `enabled: true` rendered `agent-canvas:` with an empty tag.
- The new helm-unittest suite gives good coverage (disabled cases, `/canvas` routing, TLS on/off, required-host failure).

I have one important consistency issue and one question below — no blocking chart problems.

### Important — default-on contradicts the resolved review thread and the PR description
The `agent_canvas_enabled` config default is currently `"1"` (on). The commit history shows it was set to `"0"` in `7be3514` in response to @aivong-openhands's opt-in request (and you replied on that now-resolved thread confirming you changed the default to `"0"` to match the automations pattern), but a later commit `d0eacab` ("Turns on agent canvas by default") flipped it back to `"1"`.

The PR description also still states it is "defaulted **off** (`"0"`) ... matching the automations pattern" — but `automations_enabled` defaults to `"0"`, so the shipped default no longer matches that pattern.

I see the human note that on-by-default was confirmed with @jpelletier1, so this may be an intentional product decision. If so, please:
1. Update the PR description so it no longer claims the default is off / matches automations.
2. Add a note on the resolved thread with @aivong-openhands, since the resolution reply no longer reflects the code.

Worth a sanity check against the guideline that shipped defaults should serve the self-hosted user: enabling a brand-new feature by default widens the blast radius if issues surface, which was exactly @aivong-openhands's original concern.

### Question — `authRequired: false` served on the primary app host
`staticServer.authRequired: false` in the Replicated values overrides the subchart default of `true`, so the static Canvas frontend is served unauthenticated at `/canvas` on the primary app host. Combined with `lockToCloud` pointing at the instance's own backend, I assume the intent is that the backend still enforces auth and the static bundle carries no secrets. Can you confirm that's the case, and that serving the bundle unauthenticated is acceptable here? A one-line comment in `openhands.yaml` explaining why auth is disabled (unlike the subchart default) would help future readers.

_This review was generated by an AI agent (OpenHands) on behalf of the repository maintainer._

aivong-openhands (approved):

PR #973, opened 2026-07-27
feat(openhands): PLTF-1247 migrate embedded Redis to Valkey

aivong-openhands:
## Why

Bitnami's license change froze the free Redis chart and images, so the pinned `redis 20.3.0` and `bitnamilegacy/redis` get no CVE patches and the index is not guaranteed to stay reachable. This swaps the embedded cache to the official Valkey chart (0.11.0, BSD-3-Clause), keeping the existing `redis` / `redis-password` secret and the `REDIS_*` variable names so operators create nothing new and the app needs no code change. Cutover discards the cache, which invalidates in-flight OAuth and Slack flows and briefly cold-starts the identity caches.

**Consumers that pin cache settings must translate them in the same change.** The values key becomes `valkey:`, and a leftover `redis:` block is silently ignored rather than rejected. All four internal SaaS environments pin `redis.master.resources` and none define `valkey:`, so a chart bump alone would quietly drop production's cache from 500m/1Gi to the chart defaults of 100m/128Mi. The chart README now carries the key-by-key translation, a copy-paste block, and the rollback command.

This is a narrower second attempt at #521, whose Valkey values shape, service rename and support-bundle changes carry over here. That one spanned 33 files because it bundled a CloudNativePG Postgres migration alongside the cache swap, and it renamed the secret to `valkey` / `valkey-password` with matching changes to `openhands-secrets`. Keeping the existing secret name is the deliberate reversal: `openhands-secrets` stays untouched, and no install has to create a secret before upgrading or retain the old one in order to revert.

## Validation

- **Upgrade and rollback on every consumer surface** — 0.28.0 to this change and back, on published charts, on the Replicated registry via the customer path (install from Stable, switch channel, upgrade), and on Embedded Cluster. Each direction confirmed by a completed conversation writing rate-limiter keys into the live cache.
- **The credential survives untouched** — the `redis` secret's uid and value hash were identical after every upgrade and rollback on both surfaces, so no install needs a new secret and no revert needs the old password recovered.
- **Nothing is left pointing at the old workload** — the support-bundle analyzer moved from `statefulsetStatus` on `openhands-redis-master` to `deploymentStatus` on `openhands-valkey` and back on revert, no StatefulSet, service or PVC survives the swap, and no subchart carries cache configuration of its own.
- **A stale `redis:` block is silently ignored** — the upgrade succeeds and serves conversations, but the cache runs on chart defaults instead of the pinned values (100m in place of 50m). Nothing warns, because the values schema constrains neither key. This is the basis for the consumer note above.
- **The added unit tests fail when the change is reverted** — verified by mutating the templates: restoring the old service name breaks the cache-env assertion, and restoring `statefulsetStatus` breaks both the analyzer test and the guard against any lingering `openhands-redis-master` reference.


dylan-openhands (approved):

aivong-openhands:
Closing in favor or https://github.com/OpenHands/OpenHands-Cloud/pull/1007 and https://github.com/OpenHands/OpenHands-Cloud/pull/1008 to first ship a chart that defaults valkey to off

PR #979, opened 2026-07-28
feat(openhands): PLTF-1247 note when values still set the removed redis key

aivong-openhands:
## Why

An operator whose values still set `redis:` gets the cache on chart defaults, because Helm accepts the unused block without complaint. That is a silent downgrade of whatever sizing they had configured, and it only surfaces later as an undersized cache.

This prints a note after `helm install` and `helm upgrade`. It does not guard the render: the template still succeeds, because the cache does run and only its sizing is wrong, so blocking an otherwise healthy upgrade would be disproportionate. The note names the key mapping and states that the `redis` Secret is unchanged, so nobody rotates a credential they do not need to.

Worth knowing where this does and does not reach. Notes are printed by the Helm CLI, so this covers operators running `helm upgrade` themselves, which is where an untranslated values file is most likely. GitOps and KOTS deploys never render notes, so they are not covered by this and need their values translated directly.

## Validation

- **Asserted in both directions** — the notes suite compiles `NOTES.txt` and checks the note appears when values set a `redis:` key and is absent when they do not.

### DIFF
--- a/charts/openhands/templates/NOTES.txt
+++ b/charts/openhands/templates/NOTES.txt
@@ -36,3 +36,22 @@ Next steps:
 
 Docs: https://docs.all-hands.dev
 {{- end }}
+{{- if .Values.redis }}
+
+WARNING: your values still set `redis:`, which this chart no longer reads.
+
+    The cache is now the Valkey subchart, configured under `valkey:`. Helm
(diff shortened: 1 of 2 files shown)


aivong-openhands:
Closing in favor of a render guard helm unit test https://github.com/OpenHands/OpenHands-Cloud/pull/980

PR #983, opened 2026-07-29
feat(rustfs): PLTF-1250 optional in-cluster object store

aivong-openhands:
## Why

This adds RustFS as a chart dependency and settles its values shape, ahead of replacing the bundled MinIO store with it. Keeping it gated on `rustfs.enabled` and defaulting to false means nothing renders and no existing install changes, so the swap that follows is a small diff against a values shape already exercised. Pointing a deployment at RustFS needs no new template logic, since it speaks S3 and the existing external-store values already accept an arbitrary endpoint.

The values block overrides four upstream defaults that suit a multi-node cluster rather than a bundled store: a four-replica distributed topology, the `local-path` storage class, 256Mi volumes, and no resource limits. The empty storage class depends on standalone mode, which omits `storageClassName` so the cluster default applies.

## Validation

- **The values shape works on a fresh install:** enabled into a clean namespace on a local kind cluster, it came up ready in 15s as a single-replica Deployment, with both PVCs bound at the shipped sizes on the cluster default storage class. An S3 create-bucket, upload and read-back returned the object intact.
- **Release packaging:** `make release` built a release from this branch, confirming the release path resolves the new chart repository.

### DIFF
--- a/charts/openhands/Chart.lock
+++ b/charts/openhands/Chart.lock
@@ -20,6 +20,9 @@ dependencies:
 - name: replicated
   repository: oci://registry.replicated.com/library
   version: 1.9.0
+- name: rustfs
+  repository: https://charts.rustfs.com/
+  version: 0.10.0
 - name: crd-check
   repository: oci://ghcr.io/openhands/helm-charts
   version: 0.1.0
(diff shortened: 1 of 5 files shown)


charts/openhands/Chart.lock
```diff
@@ -20,6 +20,9 @@ dependencies:
 - name: replicated
   repository: oci://registry.replicated.com/library
   version: 1.9.0
+- name: rustfs
+  repository: https://charts.rustfs.com/
+  version: 0.10.0
```
jlav, line 25: Any reason not to use 0.11.0, which is the current latest?

jlav (approved):

charts/openhands/Chart.lock
aivong-openhands: rustFS chart versions map to beta versions and preview versions so I picked a beta before the latest to not grab the cutting edge beta. I thought the 0.10.0 chart would have had more baking time to be battle tested

dylan-openhands (approved):

PR #994, opened 2026-07-30
feat(filestore): PLTF-1250 run the bundled object store on rustfs

aivong-openhands:
## Why

MinIO CE is in maintenance and the bundled store needs to move to RustFS. This makes RustFS a selectable bundled
backend and, on the Replicated installer, copies the existing objects across during the upgrade.

`templates/_objectstore.tpl` holds the backend precedence — RustFS wins over MinIO when both are enabled, which is
the state a migration runs in, with every consumer already pointed at RustFS while the old objects survive. RustFS
uses the same credentials as `minio.svcaccts[0]`, so a switch moves only the endpoint and never rotates what the app
holds; validation rejects a mismatch. The upstream RustFS chart provisions no bucket, so `rustfs-bucket.yaml` does it
with a signed PUT, create-or-409 with no delete path — deliberately, because the bundled MinIO bucket job once
shipped with purge enabled and erased conversation data on every upgrade.

**The copy in this PR does not work, and should not be merged as it stands.** Testing on a Helm install and on
Embedded Cluster found three problems, all measured, listed under Validation. In short: objects written while the copy
runs are silently lost, the completeness check cannot detect that, and the window during which conversations return
no history grows linearly with object count until the copy exhausts the upgrade timeout outright.

The root cause is the design premise. The copy has no freeze because objects are immutable per key, so nothing can be
copied half-written. That is true, and it is the wrong property: it guarantees everything copied is *correct*, and
says nothing about whether everything is *copied*. Objects created after the copy's source listing are simply never
seen. Completeness is what freezing solves, and object storage has that problem exactly as a database does — which is
why the CloudNativePG migration in #986 freezes its writers.

Three further consequences make the shape unworkable rather than merely incomplete:

- **`helmUpgradeFlags` is not a fix.** Dropping `--wait` only buys a head start bounded by rollout duration, so it is
  a fixed discount that stops mattering as data grows.
- **Retries cannot converge.** `mc cp` has no skip-existing option and re-transfers everything, so a copy that
  exceeds the timeout redoes identical work on every redeploy and dies at the same point. A plain Helm operator can
  raise `--timeout`; a Replicated customer cannot, because that flag is baked into the release.
- **The cost recurs.** The copy re-runs in full on every later upgrade until MinIO is disabled, spending the same time
  inside the timeout on unrelated configuration changes.

**What remains valid:** the backend selector, bucket provisioning, the validation guards, the `_env.yaml` wiring and
the tests. Recommend reducing this PR to those — RustFS as a selectable backend, claiming no migration — and to
land the migration separately as a `pre-upgrade` orchestrator that freezes the writers, copies only the key-set
difference so retries converge, and verifies against a source that cannot change. That follows #986's pattern.

## Validation

- **Objects written during the copy are lost, silently.** Twice on the released chart: 2,571 objects listed, 120
  written after the listing, **0 of them copied**, and the Job reported `migration complete: 2571 objects in RustFS`.
  Confirmed not eventual consistency — `mc stat` on a missing key returned `Object does not exist` afterwards.
- **The completeness check cannot fail.** It compares the destination count against the count from the same stale
  listing; both read 2,571 while 120 objects were missing. It is also inert whenever the destination legitimately
  holds more than the source, which is the normal state after any post-cutover activity.
- **The window scales linearly and is invisible.** On Embedded Cluster at 22,125 objects: 159 s window, of which
  157 s returns HTTP 200 with zero events, no 5xx at all, and no errors or warnings in the application log for the
  entire time. Roughly 13 s fixed plus 152 objects/second, bound by object count rather than bytes. At about 60,000
  objects the copy alone exhausts a 600 s timeout.
- **What does work, and carries forward:** the copy has no delete path and is additive, so it is safe to re-run;
  `mc cp --recursive` succeeds where `mc mirror` does not, because RustFS rejects mirror's recursive list comparison;
  the store is genuinely the read path, proven by deleting a sandbox, its Deployment and its PVC and finding the
  conversation still rendered; and a copy restricted to the key-set difference works against RustFS, which is what
  makes retries converge.


aivong-openhands:
@openhands fix conflicts

aivong-openhands:
Merge conflicts resolved ✅

Merged latest `main` into `aivong/rustfs-bundled-store` (commit `eeeaa0e`). The only conflict was in `charts/openhands/Chart.lock`, caused by `main` adding the `device-plugin` dependency while this branch added `rustfs`. I resolved it by including both dependencies and regenerating the lock file with `helm dependency update`, which produced a correct digest for the merged `Chart.yaml`.

The PR is now mergeable.

PR #1006, opened 2026-08-03
feat(charts): device-plugin subchart for kvm/fuse passthrough

jlav:
## Description

Ports the generic-device-plugin support from OpenHands/runtime-api#685 into this repo, since the runtime-api chart is deprecated. Sandbox pods need host device nodes (/dev/fuse for fuse mounts, /dev/kvm for hardware-accelerated QEMU) exposed as extended resources so they can request them without running privileged.

Two things beyond a straight port:

- The kvm passthrough previously used smarter-device-manager, which isn't well maintained. Both devices are now served by [squat/generic-device-plugin](https://github.com/squat/generic-device-plugin).
- The daemonset has to run in the cluster where sandboxes are scheduled, which isn't necessarily the cluster running the rest of OpenHands. It now lives in its own `device-plugin` subchart, off by default. Same-cluster installs (replicated) enable it through the umbrella; split-cluster installs run it standalone in the sandbox cluster with `helm install device-plugin charts/openhands/charts/device-plugin`.

The replicated installer keeps advertising the legacy `smarter-devices/kvm` resource name via `device-plugin.domain: smarter-devices`. Pause/resume reuses a runtime's deployment as-is (resume just scales replicas 0 -> 1), so pre-upgrade pod templates keep requesting that name. Renaming it would leave every pre-upgrade kvm sandbox unschedulable on its next resume. Fresh installs of the subchart default to `squat.ai/*`.

This removes the `runtime-api.kvm.*` values. Installs that set those directly need to move to the `device-plugin` block and update `KVM_DEVICE_RESOURCE` to match (the replicated installer is already updated here).

## Helm Chart Checklist

## Additional Notes

- The per-node device `count` is a synthetic admission cap (every claimant shares the same device node). Defaults are 256 now, since replicated nodes run tens of sandboxes and every sandbox pod requests the kvm resource when `sandbox_kvm_enabled` is on. The old smarter-device-manager config capped this at 20.
- Upstream publishes no semver image tags, only short git shas and a moving `latest`, so the image is pinned to `2cc50b0` (current upstream HEAD).
- `priorityClassName` defaults to empty because GKE quota-restricts `system-node-critical` outside kube-system. The replicated values set it explicitly since k0s has no such restriction.
- helm lint, `helm dependency build`, the unittest suites (7 new subchart tests), and the replicated template invariant tests all pass locally.
- Upgrade path validated on a replicated embedded-cluster install: baseline (smarter-device-manager, kvm on) upgraded to this branch. A sandbox paused before the upgrade resumed with working kvm, a new sandbox worked, and an unprivileged pod performed a real FUSE mount through the new daemonset. Under sysbox (how sandboxes actually run), the device resource request is the only thing the pod spec needs.
- The chart README never documented the kvm values, so there's no README change for the removal.


tofarr (approved):
🍰 - Let's see this on dev

PR #1007, opened 2026-08-03
feat(openhands): PLTF-1247 offer Valkey as an opt-in cache backend

aivong-openhands:
## Why

Bitnami's license change froze the free Redis chart and images, so the pinned `redis 20.3.0` and
`bitnamilegacy/redis` get no CVE patches. The replacement is the official Valkey chart (0.11.0, BSD-3-Clause).
This adds it as a second cache backend with a selector helper and leaves it **off**, so the swap becomes a values
change each environment makes when it chooses to rather than something an upgrade performs on everyone.

Redis stays the active cache while both are enabled, so enabling Valkey by itself deploys it without moving the
app. Disabling Redis is what performs the cutover:

```yaml
redis:
  enabled: false
valkey:
  enabled: true
```

Valkey reuses the existing `redis` Secret and the app keeps its `REDIS_*` variable names, so no install has to
create a secret before cutting over, no revert has to recover the old password, and there is no application
change. The cache is discardable, so there is nothing to migrate — reverting is the same two values back.

## Validation

- **Nothing moves for anyone who does not opt in** — `helm template` output is byte-identical to `origin/main` at
  chart defaults, with cache resources pinned the way the internal SaaS environments pin them, and with
  `replicated.enabled` on. The support-bundle analyzer messages still say "Redis" for that reason; they become
  wrong only after a cutover, so they move with it.
- **The cutover renders correctly** — `redis.enabled: false` with `valkey.enabled: true` produces the
  `openhands-valkey` Deployment and Service, points `REDIS_HOST` at it, still reads the password from the `redis`
  Secret's `redis-password` key, leaves no reference to `openhands-redis-master`, and swaps the readiness analyzer
  from `statefulsetStatus` to `deploymentStatus` — Valkey is a Deployment, so an analyzer left on the old kind
  would report permanently unready.
- **Eight unit tests cover all four states** — defaults on Redis, cutover on Valkey, both-enabled staying on
  Redis, and neither dropping the cache env entirely.


dylan-openhands (approved):
This may already be covered but I see that we can have both redis and valkey enabled.

I think this is fine and makes sense. However anything else we can tack-on maybe as some dynamic helm notes to call out "hey you've got both of these enabled, is that intentional"

This is just to catch the possibility of operators simply forgetting to actually cutover. Anyways **non-blocking** and this might be handled in a way i'm not seeing

aivong-openhands:
@openhands fix conflicts

aivong-openhands:
Good catch, and agreed that a hard guard would be too strong here — the consequence is a redundant pod, not a broken install. Added the dynamic note in `7abc704`:

```
NOTE: both cache backends are enabled, so two caches are deployed and Redis is
the one in use. If you meant to move to Valkey, also set:

    redis.enabled=false
```

It sits above the normal notes and is deliberately not scoped to `.Values.enabled`, since the cache subcharts install independently of the enterprise-server, so a server-less release can be misconfigured the same way. Four unit tests cover it: both enabled, both enabled with the app off, and silence at the chart defaults.

aivong-openhands:
Resolved the conflicts by merging current `main` and regenerating `Chart.lock` so both the Valkey and device-plugin dependencies are preserved. The resolution is included in `672ae59`.

Local validation passed: 95 Helm tests and 19 Python tests, plus 74 parent-chart tests after integrating the concurrent notes update. GitHub now reports the PR as mergeable; CI checks are still running.

aivong-openhands:
> This may already be covered but I see that we can have both redis and valkey enabled.
> 
> I think this is fine and makes sense. However anything else we can tack-on maybe as some dynamic helm notes to call out "hey you've got both of these enabled, is that intentional"
> 
> This is just to catch the possibility of operators simply forgetting to actually cutover. Anyways **non-blocking** and this might be handled in a way i'm not seeing

Added a NOTES.txt warning in https://github.com/OpenHands/OpenHands-Cloud/pull/1007/changes/7abc7042ce9db02108161a06c1b3b32d626c3561!

PR #1017, opened 2026-08-04
feat(filestore): PLTF-1250 deploy bundled RustFS consumed by nothing

aivong-openhands:
## Why

Deploys the RustFS subchart as a bundled object store that nothing consumes. A new backend selector keeps MinIO in precedence while both stores are enabled, so the app's endpoint and credentials do not change and no data can be stranded. RustFS gets its own release because the migration that follows copies data into it from a `pre-upgrade` hook, and Helm refuses to adopt a resource a hook created, so the store has to already exist and already be Helm-managed before that hook runs. Plain Helm installs are unaffected, since the chart default stays `rustfs.enabled: false` and only Replicated installs turn it on. A render guard rejects credentials that differ between the two stores, because the app is handed one pair for whichever store is active. The change also bounds the store's log volume, which the subchart otherwise leaves to grow without limit: the subchart's own `log_rotation` knobs do not work, as the server reads no rotation-size setting and accepts only `minutely`, `hourly` or `daily`, so the documented `hour` silently meant daily. Retention is therefore set with the byte keys the server does read, plus the file-age gate that makes those ceilings take effect at all.

## Validation

- **An upgrade leaves the application where it was** — on a Replicated Helm install, all three consumers still resolve `AWS_S3_ENDPOINT` to MinIO across every container including init containers, every app Deployment's `metadata.generation` and every app pod UID are unchanged, and the MinIO bucket holds the identical object set before and after, compared by count, bytes and a sha256 of the sorted key list.
- **Same result on Embedded Cluster** — RustFS came up with its Service and both PVCs bound at the configured sizes, the bucket Job completed, the app stayed on MinIO with its PV from before the upgrade, both store images resolved through the proxy registry, and all preflights passed.
- **RustFS is Helm-managed, not orphaned** — its Deployment, Service and both PVCs carry `app.kubernetes.io/managed-by: Helm` with the release-name and release-namespace annotations, which is what lets a later release's hook adopt them instead of creating them.
- **At the `warn` level this ships, log volume tracks faults rather than traffic** — 5000 object writes plus 400 client errors produced zero log bytes, because a missing bucket or key is a normal S3 response and only the per-request logging at `info` records those, while a genuine internal fault does reach the file at `warn` at roughly 320 bytes a line. The byte ceilings then only bind because the file-age gate is set: with that gate left at its default, a scaled-down 8 MiB ceiling still let the log directory reach 1.27 GB across 1220 files in 92s, since nothing younger than an hour is eligible for cleanup.

### DIFF
--- a/charts/openhands/templates/_env.yaml
+++ b/charts/openhands/templates/_env.yaml
@@ -48,11 +48,12 @@
 - name: SANDBOX_CLOSE_DELAY
   value: "1800"
 {{- if eq .Values.filestore.type "s3" }}
-{{- /* Empty endpoint targets the bundled MinIO when it is deployed, otherwise
-       real AWS (boto3 uses the regional endpoint). */}}
+{{- /* Empty endpoint targets the bundled object store (MinIO or RustFS) when
+       one is deployed, otherwise real AWS (boto3 uses the regional endpoint).
+       See templates/_objectstore.tpl for backend selection. */}}
 {{- $endpoint := .Values.filestore.endpoint }}
-{{- if and (not $endpoint) .Values.minio.enabled }}
-{{- $endpoint = printf "http://%s-minio:9000" $.Release.Name }}
+{{- if and (not $endpoint) (include "openhands.objectStore.backend" .) }}
+{{- $endpoint = include "openhands.objectStore.endpoint" . }}
 {{- end }}
 - name: FILE_STORE
   value: s3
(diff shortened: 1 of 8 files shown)


dylan-openhands (approved):

aivong-openhands:
- [ ] evaluate RustFS PVC sidecar approach

aivong-openhands:
Closing in favor of a one release upgrade https://github.com/OpenHands/OpenHands-Cloud/pull/1028

PR #1019, opened 2026-08-04
feat(filestore): PLTF-1250 migrate bundled object store to RustFS

aivong-openhands:
## Why

Copies the bundled object store from MinIO into RustFS and cuts consumers over, in the release after RustFS was deployed empty. `filestore.migration.enabled` flips the backend selector and runs a `pre-upgrade` orchestrator: it waits for RustFS (deployed by the earlier release, never created here), compares the two stores read-only, and freezes writers only when something is actually outstanding. It then copies each bucket's key-set difference in a transient `mc` Pod, verifies by recomputing that difference against the frozen source, and lets the apply repoint consumers and bring the writers back. MinIO stays deployed and stops receiving writes, so it is a clean rollback source for the state at cutover until a later release retires it.

Safety rests on three properties. The copy is a `pre-upgrade` hook, so any failure aborts before the apply and consumers are never repointed. The copy is incremental by key-set difference and additive with no delete path, so a retry resumes instead of restarting. And the decision to copy comes from the live difference rather than a completion marker, because a marker outlives a return to the earlier release while consumers keep writing to MinIO, so trusting one would repoint them onto a store missing everything written since. Zero outstanding exits before any writer is frozen, which matters because the switch stays on for every later release.

One coupling is invisible from either template alone. Helm skips `post-upgrade` hooks when an upgrade fails, so a failure here would leave the CNPG migration's freeze un-resumed. The orchestrator therefore restores every writer it finds at zero, including ones it did not freeze itself.

## Validation

- **The migration completes on the shipping surface** — on Embedded Cluster with two populated buckets, the copy Pod ran once per bucket (3028 session objects, 41 automation objects), sorted-key sha256 matched on both stores with an empty key-set difference in each direction, a planted key containing spaces survived intact, MinIO's fingerprint was unchanged afterwards, and all four writers repointed across their main and init containers and returned to 1/1 from the apply. Writers were down 120s, of which the copy was 85s, using 4.7% of the hook's budget.
- **The frozen set is every writer** — an independent scan of all 18 workloads, covering every container, init container and `envFrom`, found the four the orchestrator freezes are the only ones that write to the store. The scheduled jobs that carry the same object-store environment from the shared env template do not use it: the maintenance task every install runs daily works only against the database, and the GitLab webhook installer, off by default, does API and database work.
- **Going back and forward again copies only the delta** — the completion marker survived the return to the earlier release, so the hazard it guards is real on this surface. The check then reported 25 objects outstanding, exactly what had been written while back on the earlier release, copied 25 of 3053 with the automation bucket correctly reporting 0 to copy, rewrote the marker from 3028 to 3053, and left a MinIO-minus-RustFS difference of 0. The retry cost 2s of copying in a 16s hook, against 85s and 91s for the initial migration. Repeating the same upgrade with nothing outstanding froze no writers and finished in 18s with the app Deployment's generation unchanged.
- **Two unrelated failures each abort before anything is repointed** — deploying with RustFS missing aborted at the hook's first action, the readiness check, with a message naming the fix and before the restore trap was even installed. Separately, blocking the copy Pod's network aborted later, at the read-only comparison step, on a DNS timeout. Each left the release failed with healthy workloads, consumers unchanged, no freeze record, no copy Pod, no Deployment generation churn, and the app serving throughout. Two different routes to the same safe outcome, so the guarantee does not rest on one check.
- **The bucket list matches what is actually deployed** — the sandbox workspace archive is the one object-store consumer KOTS never enables, so it was exercised on a Helm install with the archive resolving to s3 through an empty store type: its 61 objects reached RustFS with a key-set difference of 0 in both directions, including one at a path containing spaces, and the marker recorded both buckets. Rendering across six configurations confirms only the cases that should add a bucket do, no empty entry is appended, and a duplicate collapses.

### DIFF
--- a/charts/openhands/templates/_objectstore.tpl
+++ b/charts/openhands/templates/_objectstore.tpl
@@ -9,8 +9,14 @@ Two bundled backends, in this precedence order:
 MinIO wins when both are enabled, so enabling RustFS by itself deploys it
 without repointing the app — the same shape templates/_cache.tpl uses for
 redis/valkey, where the incumbent keeps precedence. Deploying RustFS therefore
-never moves a consumer onto an empty store; the migration switch (a later
-release) is what repoints consumers, once it has copied the data across.
+never moves a consumer onto an empty store.
+
+The migration switch is what repoints consumers. filestore.migration.enabled
+flips precedence to RustFS: the pre-upgrade hook copies the data across before
+Helm's apply lands, and the apply then renders every consumer pointed at RustFS.
+MinIO stays deployed (minio.enabled) as the rollback source, exactly as the
+CNPG migration keeps the old database server deployed while consumers already
+resolve to the new one.
 
 Neither backend is bundled when both are disabled: that install uses an
 external store (S3/GCS) or none, and these helpers render nothing.
@@ -22,7 +28,9 @@ so switching backends does not rotate anything the app holds.
 
 {{/* "minio", "rustfs", or "" when no bundled store is deployed. */}}
 {{- define "openhands.objectStore.backend" -}}
-{{- if .Values.minio.enabled -}}
+{{- if and .Values.rustfs.enabled .Values.filestore.migration.enabled -}}
+rustfs
+{{- else if .Values.minio.enabled -}}
 minio
 {{- else if .Values.rustfs.enabled -}}
 rustfs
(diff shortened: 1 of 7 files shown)


aivong-openhands:
Closing in favor of a one release upgrade https://github.com/OpenHands/OpenHands-Cloud/pull/1028

PR #1022, opened 2026-08-05
feat(filestore): PLTF-1250 retire the bundled MinIO

aivong-openhands:
## Why

Removes the bundled MinIO, the release after its data was copied into RustFS and every consumer repointed. MinIO stayed deployed through that release as the rollback source, and this is the deliberate retirement it was waiting for. Going with it: the subchart dependency and its values, the two-backend precedence selector, the purge guard that stopped MinIO's bucket job wiping conversation data on every upgrade, the MinIO branches of the support-bundle collector and analyzer, and the migration hook itself, whose source no longer exists. The KOTS values lose the MinIO image, its airgap builder entries, the migration images and the `NO_PROXY` host.

Two parts are not simple deletions. The app's bundled-store credentials were sourced from `minio.svcaccts[0]` even while RustFS was the active backend, and now come from `rustfs.secret.rustfs`. The two were seeded with the same pair and a render guard rejected a mismatch, so an install on chart defaults sees no change and nothing it holds is rotated; an install that overrode the MinIO pair has to set the same values under `rustfs.secret.rustfs`. And a values file still carrying a `minio:` block would have it silently ignored once the key is unknown, leaving an install that depended on it with no bundled store and an empty `FILE_STORE`. A new guard fails the render with that instruction instead, taking the retired purge guard's place.

This is only safe once every install has run the migration, which promoting that release as required enforces.

## Validation

- **Nothing MinIO survives in a rendered release** — rendering with the bundled store enabled produces 57 resources and zero MinIO references, with consumers on `rustfs-svc:9000` and `AWS_ACCESS_KEY_ID` resolving from the RustFS pair, so the credential move is visible in the output rather than inferred from the diff.
- **The stale-values guard fires on the case that would otherwise be silent** — a values file setting `minio.enabled` with RustFS off fails the render with the instruction to enable RustFS or configure an external store. It stays quiet once RustFS is enabled, and quiet for an external-store release that sets no `minio` block, so it catches the broken combination without blocking the working ones.
- **`Chart.lock` matches `Chart.yaml`** — regenerated after dropping the dependency, which is what CI's `helm dependency build` verifies and the thing a hand-edited lock breaks.
- **The suites that asserted the old behaviour now assert the new** — 16 suites and 85 tests pass after rewriting the four that covered MinIO precedence, the credential-match guard, the purge guard and the migration switch, and deleting the migration suite along with its template.


aivong-openhands:
Closed in favor of https://github.com/OpenHands/OpenHands-Cloud/pull/1029

PR #1028, opened 2026-08-05
feat: PLTF-1250 migrate bundled object store to RustFS in one release

aivong-openhands:
## Why

OpenHands ships MinIO as its bundled object store; this migrates it to RustFS in a single upgrade. A pre-upgrade hook freezes the writers, stands up RustFS backed by a new PVC (`openhands-rustfs-data`), copies every bucket from MinIO with a transient `mc` pod, and Helm's apply repoints the app and automation at RustFS while MinIO stays as the rollback source until a later upgrade retires it.

That single upgrade turns on one non-obvious piece: the RustFS PVC is declared in the ordinary template only on install or when migration is off, while during a migrating upgrade the template emits nothing and the pre-upgrade hook creates and owns the claim. Helm adopts it in place on the next upgrade that turns migration off — necessarily the MinIO retirement, since with migration off the selector routes consumers back to MinIO unless MinIO is also disabled, so migration can only be turned off in the release that retires it. The split is forced by Helm's ordering — it snapshots a release's resources before its hooks run — and the CloudNativePG migration in #986 orders it the same way; `templates/rustfs-data-claim.yaml` carries the full rationale.

On Replicated/Embedded Cluster these lifecycle switches are values the vendor sets per release, and this release ships the migrate configuration — RustFS on, migration on, MinIO kept on as the rollback source, and the automation endpoint moved to the RustFS service — so every Replicated upgrade to it runs the freeze + copy and repoints consumers onto RustFS. It is meant to ship in one release alongside the CloudNativePG migration in #986, whose orchestrator (pre-upgrade hook weight -2) runs ahead of this one (-1) in a shared freeze window. A follow-up release then turns migration off — the point at which Helm adopts the claim — and retires MinIO. A direct Helm install owns these values itself, so its operator flips `rustfs.enabled` and `filestore.migration.enabled` and migrates whenever they are ready.

Automation's endpoint can't be templated from the parent chart, so it is a static value that each migrate and retire release must set in lock-step with the switches; the app and archive repoint automatically via the selector.

## Validation

- **Adoption ordering holds on both Helm majors** — the migrating upgrade keeps the PVC out of the release manifest (the hook creates it) and the retirement upgrade adopts it in place with an unchanged UID and no RustFS restart. Walked baseline→migrate→retire on Helm 3.21 (kind) and baseline→migrate→retire→abort on Helm 4 (Embedded Cluster).
- **The copy is byte-identical and non-destructive** — every bucket matched on RustFS by object count, bytes, and key set, including keys with spaces and nested prefixes, and MinIO was left byte-identical as the rollback source.
- **Both consumers follow the active backend** — app sessions and automation repointed MinIO→RustFS and back, with automation's writes landing on the live store at each step, and the runtime archive (when a Helm user enables it) rode the app's `AWS_S3_*`; this is the basis for the lock-step note in Why.

### DIFF
--- a/replicated/openhands.yaml
+++ b/replicated/openhands.yaml
@@ -456,9 +456,9 @@ spec:
         repository: 'proxy/{{repl LicenseFieldValue "appSlug"}}/docker.io/bitnamilegacy/redis'
       auth:
         password: '{{repl ConfigOption "redis_password" | Base64Encode }}'
-    # Conversation/session files live on the bundled MinIO, which is durable
-    # here (persistence below). With no endpoint or credentials configured,
-    # the app targets the bundled instance automatically.
+    # Conversation/session files live on the bundled object store, durable here
+    # (persistence below). With no endpoint or credentials configured, the app
+    # targets the active bundled backend automatically (see templates/_objectstore.tpl).
     filestore:
       type: s3
       # Switches for the bundled object store: they select the backend and run the
@@ -467,7 +467,7 @@ spec:
       # MinIO kept as the rollback source; turning minio off afterwards retires it
       # and Helm adopts the claim. Move the automation endpoint (below) in lock-step.
       migration:
-        enabled: false
+        enabled: true
         images:
           kubectl:
             repository: 'images.r9.all-hands.dev/proxy/{{repl LicenseFieldValue "appSlug"}}/docker.io/alpine/k8s'
@@ -494,7 +494,7 @@ spec:
     # (templates/_objectstore.tpl); deployed alone it changes nothing, since MinIO
     # keeps precedence until the migration runs.
     rustfs:
-      enabled: false
+      enabled: true
       image:
         rustfs:
           repository: 'images.r9.all-hands.dev/proxy/{{repl LicenseFieldValue "appSlug"}}/docker.io/rustfs/rustfs'
@@ -718,7 +718,7 @@ spec:
           # undeployed.
           filestore:
             type: s3
-            endpoint: http://openhands-minio:9000
+            endpoint: http://openhands-rustfs-svc:9000
             # The bundled store's bucket job only creates the parent chart's
             # session bucket, so the service provisions its own on startup.
             createBucket: true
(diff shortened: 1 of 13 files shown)


replicated/openhands.yaml
```diff
@@ -473,6 +490,21 @@ spec:
...
+      image:
+        rustfs:
+          repository: 'images.r9.all-hands.dev/proxy/{{repl LicenseFieldValue "appSlug"}}/docker.io/rustfs/rustfs'
+        # RustFS runs a busybox init container even in standalone mode.
+        initImage:
+          repository: 'images.r9.all-hands.dev/proxy/{{repl LicenseFieldValue "appSlug"}}/docker.io/library/busybox'
+      imagePullSecrets:
+        - name: '{{repl ImagePullSecretName }}'
+      storageclass:
+        dataStorageSize: 100Gi
```
aivong-openhands: - [x] maybe this could be a configurable value? CC @dylan-openhands

aivong-openhands: Went with hardcoding it, but bumped it to **500Gi to match MinIO** (was 100Gi) — done in ede517d57.

Two reasons not to make it a config option:

- **It should never need to be larger than the store we migrate from.** The bundled MinIO PVC is 500Gi (the minio subchart default; our overlay enables persistence but never overrode the size). RustFS only ever holds what we copy out of MinIO, so MinIO's own size is both the floor and the ceiling — there's no case where an operator needs it bigger. At 100Gi we were actually sizing the *target* smaller than the *source*, which on a capacity-enforcing storage class (EBS/PD/hyperdisk) could fail the copy with out-of-space. 500Gi removes that risk and is just parity with MinIO — no regression, since MinIO was already 500Gi there.
- **It's create-time only.** A PVC's size is fixed at provisioning — immutable except grow-in-place, and only where the storage class allows expansion. A config value would only ever take effect at first install and couldn't resize an existing volume, so the knob wouldn't buy the flexibility it implies.

(On Replicated's default hostpath the request is metadata-only and not enforced, so it's nominal there either way; it only becomes a real allocation on cloud block storage — where MinIO was already 500Gi too.)

Carried the same change into the retire release in #1029.

_Posted by an AI agent (OpenHands) on behalf of @aivong._

PR #1035, opened 2026-08-07
feat(chart): OHE-3021 : expose OH_SANDBOX_MAX_NUM_SANDBOXES as a ConfigOption

tofarr:
## Description

Exposes `OH_SANDBOX_MAX_NUM_SANDBOXES` as a configurable value, letting operators cap the number of concurrent sandboxes the app will start. This bounds resource consumption on shared or memory-constrained clusters; once the cap is reached, new conversations wait for an existing sandbox to be freed rather than spawning a new one.

The change follows the same plumbing path as the existing `RUNTIME_URL_PATTERN` variable, applied in three coordinated places:

- **`charts/openhands/values.yaml`** — added the env var to the `env:` block (default `"0"` = no cap), so non-Replicated installs are uncapped unless an operator opts in.
- **`replicated/config.yaml`** — added a `sandbox_max_num_sandboxes` ConfigOption in the `sandbox_configuration` group, modeled on the adjacent `sandbox_warm_runtime_count` (text type with a `^(0|[1-9][0-9]*)$` non-negative-integer regex validation). Default `"0"`.
- **`replicated/openhands.yaml`** — wired the ConfigOption into the app-server env block. The path-routing `optionalValues` block uses `recursiveMerge: true` and only overrides `RUNTIME_URL_PATTERN`, so the new var is inherited there automatically — no repetition needed.

## Helm Chart Checklist

## Additional Notes

- The default of `"0"` (uncapped) preserves the chart's prior behavior — no cap existed before — so this is a pure opt-in addition with no behavioral change for existing installs.
- Verification: `helm template` renders `OH_SANDBOX_MAX_NUM_SANDBOXES` in the app-server, runtime-api, and warm-runtime envs (mirrored everywhere `RUNTIME_URL_PATTERN` goes). All existing env-wiring test suites pass (`filestore env`, `smtp env`, `valkey env`, `sandbox hostname layout`). The 4 pre-existing subchart test failures (missing `device-plugin` daemonset templates) are unrelated and present on `main` without these changes.

### DIFF
--- a/charts/openhands/values.yaml
+++ b/charts/openhands/values.yaml
@@ -88,6 +88,11 @@ env:
   # each sandbox ingress. The default nests sandboxes under the runtime base,
   # matching the runtime-api's own "." separator.
   RUNTIME_URL_PATTERN: "https://{runtime_id}.runtime.app.example.com"
+  # Maximum number of concurrent sandboxes the app will start (0 = no cap).
+  # The Replicated installer exposes this as the sandbox_max_num_sandboxes
+  # ConfigOption; leave at 0 here so non-Replicated installs are uncapped by
+  # default and operators opt in to a limit.
+  OH_SANDBOX_MAX_NUM_SANDBOXES: "0"
   # replace prod/claude-3-7-sonnet-20250219 with your LLM model and uncomment or override this variable
   # LITELLM_DEFAULT_MODEL: "litellm_proxy/prod/claude-3-7-sonnet-20250219"
 
--- a/replicated/config.yaml
+++ b/replicated/config.yaml
@@ -1239,6 +1239,20 @@ spec:
             regex:
               pattern: ^(0|[1-9][0-9]*)$
               message: Must be a non-negative integer
+        - name: sandbox_max_num_sandboxes
+          title: Max Concurrent Sandboxes
+          help_text: >-
+            Maximum number of concurrent sandboxes the app will start across the
+            cluster. Set to 0 (default) for no cap. Use this to bound resource
+            consumption on shared or memory-constrained clusters; once the cap is
+            reached, new conversations wait for an existing sandbox to be freed
+            rather than spawning a new one.
+          type: text
+          default: "0"
+          validation:
+            regex:
+              pattern: ^(0|[1-9][0-9]*)$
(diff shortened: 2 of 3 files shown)


aivong-openhands (approved):
### Code review — APPROVE

Clean, minimal, opt-in addition that follows the established chart conventions. Verified against the repo's patterns and by rendering the chart.

**What I checked**
- **Pattern adherence:** The new `OH_SANDBOX_MAX_NUM_SANDBOXES` env var is wired exactly like the existing `RUNTIME_URL_PATTERN` (chart `env:` block), and the `sandbox_max_num_sandboxes` ConfigOption mirrors the adjacent `sandbox_warm_runtime_count` — same `type: text`, same `^(0|[1-9][0-9]*)$` non-negative-integer regex, same `repl{{ ConfigOption ... }}` reference form.
- **Override-friendly:** The value flows through the `openhands.env` de-duplication helper in `_env.yaml`, so consumers can override it from their site values without collisions.
- **Render check:** `helm template charts/openhands` renders `OH_SANDBOX_MAX_NUM_SANDBOXES` in all 11 locations that carry `RUNTIME_URL_PATTERN` (app-server deployment, cronjobs, init containers).
- **Path-routing inheritance:** The `runtime_routing_mode=path` `optionalValues` block uses `recursiveMerge: true` and only overrides `RUNTIME_URL_PATTERN`, so the new var is inherited there automatically — confirmed, no repetition needed.
- **Default `"0"` (uncapped):** Preserves prior behavior and is the right default for a self-hosted user — a pure opt-in with no behavioral change for existing installs.
- **Consumer-friendly config:** ConfigOption title/help_text describe the effect ("Max Concurrent Sandboxes", cap semantics) rather than internal conventions.

**Non-blocking FYI**
- The PR description says the var is mirrored into the "runtime-api, and warm-runtime envs." In practice it renders into the parent chart's `.Values.env` consumers (app-server + cronjobs + init containers), same as `RUNTIME_URL_PATTERN` — which is correct, since the app server is what enforces the cap. Just a wording nit in the description; the code is right.
- `build/` is a gitignored generated artifact, so no update needed there.

Nice work — approving.

PR #1042, opened 2026-08-10
feat(keycloak): wire SMTP config into realm for forgot-password

tofarr:
## Description

Propagate SMTP configuration through to the Keycloak default realm so that forgot-password and email-verification workflows work. The realm template's `smtpServer` was hard-coded; it now reads every field via envsubst from `KEYCLOAK_SMTP_*` env vars, and the realm flags `loginWithEmailAllowed` and `resetPasswordAllowed` are enabled.

### Changes

- **`values.yaml`**: New `keycloak.smtp` block (`host`, `port`, `from`, `fromDisplayName`, `user`, `ssl`, `starttls`) with Resend defaults so existing installs keep working unchanged.
- **`allhands-realm-github-provider.json.tmpl`**: `smtpServer` fields are now envsubst-driven (`$KEYCLOAK_SMTP_*`); added `port` field; set `loginWithEmailAllowed: true`, `resetPasswordAllowed: true`.
- **`_env.yaml`**: 7 new `KEYCLOAK_SMTP_*` env vars, gated on `keycloakProvisionRealm` so they only render when the realm-config init container runs.
- **`keycloak-config-script.yaml`**: envsubst variable list extended with the new `KEYCLOAK_SMTP_*` vars.
- **`replicated/config.yaml`**: New `keycloak_smtp_configuration` config group with host, port, from, from display name, user, SSL, STARTTLS items.
- **`replicated/openhands.yaml`**: Wires the ConfigOptions into `keycloak.smtp` values.
- New helm-unittest suite (`keycloak_smtp_env_test.yaml`) covering env injection, custom overrides, and the disabled-realm case.

## Helm Chart Checklist

## Additional Notes

- The build config items (`build/config.yaml`, `build/openhands.yaml`) were wired in a prior commit on this branch and are not changed here.
- Existing Python realm-template tests (18) and helm-unittest suites (15) all pass; the 4 pre-existing subchart test failures are unrelated (missing subchart dependencies in the parent-chart run, run separately in CI).
- Defaults target Resend so existing installs that relied on the hard-coded Resend values keep working; operators override via the Replicated `keycloak_smtp_*` ConfigOptions for customer SMTP servers.


aivong-openhands (commented):
### Code review: `feat(keycloak): wire SMTP config into realm for forgot-password`

Reviewed against the OpenHands-Cloud Helm chart review guidelines. The core SMTP change is well-executed and CI is green, but there are a couple of things worth addressing before merge, so I'm leaving this as a comment rather than an approval.

#### Important

**1. The PR bundles a second, undocumented feature.**
There are two commits here:
- `feat(keycloak): wire SMTP config into realm for forgot-password`
- `feat(chart): expose OH_SANDBOX_MAX_NUM_SANDBOXES as a ConfigOption`

The description only documents the SMTP work, but the diff also adds the sandbox-cap wiring (`OH_SANDBOX_MAX_NUM_SANDBOXES` in `values.yaml`, `sandbox_max_num_sandboxes` in `replicated/config.yaml`, and the mapping in `replicated/openhands.yaml`). That is an unrelated concern that isn't mentioned in the "Changes" list at all. Please either split it into its own PR or, at minimum, document it in the description so reviewers and the changelog reflect what actually ships.

#### Consider

**2. `resetPasswordAllowed: true` plus Resend/All-Hands defaults vs. the self-hosted user (Principles 7 and 8).**
The realm now ships with `loginWithEmailAllowed: true` and `resetPasswordAllowed: true` unconditionally, while the SMTP defaults point at All Hands' own provider (`host: smtp.resend.com`, `from: admin@all-hands.dev`, `user: resend`). For a self-hosted operator who hasn't configured a working SMTP server, this means the login page now advertises a "forgot password" flow that will silently fail (the required password still comes from the `keycloak-realm` secret).

The backwards-compat rationale for keeping the Resend defaults is reasonable since those values were previously hard-coded in the template, so this isn't a blocker. But it's worth deciding: should `resetPasswordAllowed` be gated on SMTP actually being configured, so we don't surface a broken flow by default to self-hosted installs? An All-Hands-specific `from` address as a shipped default is exactly the kind of value Principle 7 suggests belongs in our environment overrides rather than the chart default.

#### Minor

**3. Branch is behind `main`.**
The branch's `charts/openhands/Chart.yaml` is at `0.41.0` while `main` is at `0.49.0`. CI (`check-appversion-drift`, `publish-charts`) is passing, so it's not blocking, but a rebase before merge would avoid surprises with the version bump.

#### What looks good
- Env var naming follows the conventions: service-owned vars use the `KEYCLOAK_SMTP_*` prefix, and they're correctly gated behind `keycloakProvisionRealm` so they only render with the realm-config init container.
- Good helm-unittest coverage, including the disabled-realm case (`notMatchRegexRaw`) and the custom-override case.
- Follows the existing `envsubst` template pattern rather than inventing a new mechanism.
- Types are consistent (`port` as string, `ssl`/`starttls` as booleans that get quoted for env injection), and defaults preserve existing behavior.

---

PR #548, opened 2026-04-14
TEST a random change

aivong-openhands:
## Helm Chart Checklist

- [ ] I have updated the `version` field in `Chart.yaml` for each modified chart
- [ ] I have tested the chart upgrade path from the previous version
- [ ] I have verified backwards compatibility with existing values.yaml configurations
- [ ] I have updated the chart's README.md if there are any breaking changes or new required values

PR #1043, opened 2026-08-10
fix(automation-chart): reject legacy ephemeral GCS storage

neubig:
## Problem

The legacy automation chart accepted this contradictory configuration:

```yaml
filestore:
  ephemeral: true
  type: gcs
```

The old `ephemeral` rendering path forcibly selected S3/MinIO and ignored `filestore.type`, so the deployment worked while the values claimed GCS. This is the exact inconsistency previously present in saas-deploy feature values.

## Fix

Add one narrowly scoped render-time guard to the automation subchart:

- when legacy `filestore.ephemeral` is `true`, `filestore.type` must be `s3`

Add a regression test for `ephemeral: true` plus `type: gcs`.

No general backend-field validation or unrelated chart validation is included.

## Validation

- Automation chart: **8 tests passed**
- OpenHands chart: **81 tests passed**
- `git diff --check` passed

### DIFF
--- a/charts/openhands/charts/automation/templates/_env.yaml
+++ b/charts/openhands/charts/automation/templates/_env.yaml
@@ -39,6 +39,9 @@
 # to "gcs", the service's native FILE_STORE value.
 {{- $store := .Values.filestore.type }}
 {{- if eq $store "google_cloud" }}{{- $store = "gcs" }}{{- end }}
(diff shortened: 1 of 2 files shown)


neubig:
Closing this PR because the guard targets the removed  field in the current chart main. The affected deployed chart configuration is handled by saas-deploy PR #548, which fixes the feature values directly. This action was performed by an AI agent (OpenHands) on behalf of the user.

PR #1092, opened 2026-08-14
test(e2e): make returning/new-user roles opt-in via *_GITHUB_USERNAME

tofarr:
## Description

The e2e harness previously required both the Returning User and New User GitHub credential sets on every run. This is awkward for scenarios like a freshly spun-up test cluster (e.g. via Replicated) where there are no existing users — only brand-new accounts exist, so the "returning" path doesn't apply and placeholder credentials shouldn't be required.

Each user role is now **opt-in via its `*_GITHUB_USERNAME` env var**. Leave either `RETURNING_GITHUB_USERNAME` or `NEW_GITHUB_USERNAME` unset (or empty) to skip that role entirely — its setup project, the Keycloak cleanup step, and all of its paired test projects (`chromium/firefox/webkit` × `:returning`/`:new-user`) are excluded from the run. When both are set, behavior is unchanged (all 15 projects run).

### How it works

- `utils/config.ts` — adds `isUserEnabled(user)`, which checks whether the role's `*_GITHUB_USERNAME` is set.
- `playwright.config.ts` — each project's `testMatch` is gated through `matchFor(enabled, pattern)`, returning the real pattern when the role is enabled or `/$a/` (matches nothing) when disabled. This ensures a disabled role's test project matches zero specs and never tries to load a storage-state file that was never produced.
- `tests/setup/{setup-returning,setup-new-user,keycloak-cleanup}.ts` — each setup test begins with a `setup.skip()` guard as defense-in-depth for explicit `--project=` invocations without creds.

I verified the key Playwright behavior before implementing: a project whose `testMatch` matches zero tests does **not** error on a missing `storageState` file, and this approach keeps runs green across all combinations (neither / only returning / only new-user / both).

| Scenario | `RETURNING_GITHUB_USERNAME` | `NEW_GITHUB_USERNAME` | Projects run |
| --- | --- | --- | --- |
| Fresh cluster (no users) | unset | unset | 0 — clean, no errors |
| Only returning | set | unset | 7 — `setup:returning` + `*:returning` specs |
| Only new user | unset | set | 8 — `keycloak-cleanup` + `setup:new-user` + `*:new-user` specs |
| Both (original behavior) | set | set | 15 — unchanged |

`npm run lint` (typecheck + eslint + prettier) passes.

## Helm Chart Checklist

> No Helm chart changes in this PR — the checklist above is N/A. Changes are confined to the `e2e_tests/` Playwright harness. Backwards compatibility is preserved: existing runs that set both `*_GITHUB_USERNAME` vars see no change in project topology.

## Additional Notes

- This PR was created by an AI agent (OpenHands) on behalf of the repository contributor.
- The `setup.skip()` guards in the setup files are belt-and-suspenders: in normal operation the gated `testMatch` already excludes those tests, but the skips make explicit `--project=setup:new-user` runs (without creds) fail cleanly with a skip rather than a credential error.

### DIFF
--- a/e2e_tests/README.md
+++ b/e2e_tests/README.md
@@ -9,6 +9,8 @@ This directory contains the Playwright harness and release-level end-to-end test
 - An explicit non-production or intentionally selected release target
 - Three credential sets: Keycloak admin, Returning User (GitHub), and New User (GitHub) — or pre-generated Playwright storage states for the two user roles
 
+> **Opt-in roles:** Each user role is enabled by setting its `*_GITHUB_USERNAME` env var. Leave either unset (or empty) to skip that role entirely — its setup project, Keycloak cleanup, and test projects are excluded from the run. This is useful for fresh clusters (e.g. a spun-up test cluster with no existing users) where the "returning" path doesn't apply, or where only one role is relevant.
+
 ## Install
 
 ```bash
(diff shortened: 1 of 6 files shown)


hieptl (approved):
Thank you! 🙏

PR #1093, opened 2026-08-14
Added local dev test

tofarr:
## Helm Chart Checklist

## Additional Notes

`npm run test:dev` now yields:
* A UI for debugging
* Only chromium tests
* Only returning user

image


hieptl (approved):
Thank you! 🙏

PR #1094, opened 2026-08-14
test(e2e): migrate Stripe credit purchase into billing suite

tofarr:
## Description

Port the `should be able to purchase $10 credits via Stripe` test from `saas_deploy`'s `e2e_tests/tests/smoke.spec.ts` into a dedicated `tests/billing.spec.ts`, adapted to this harness's conventions (`runUser` annotation, `HomePage` page object, prettier formatting, `STRIPE_TEST_CARD` constant).

Add a reusable billing feature-flag gate in `utils/billing.ts`: an auto fixture fetches `/api/v1/web-client/config` once (cached for the worker) and skips every billing spec when `body.feature_flags.enable_billing` is falsy. Billing specs import the gated `{ test, expect }` from there instead of `@playwright/test`, so the whole suite is skipped automatically on deployments that don't expose billing — no separate config flag to set, no false failures.

## Helm Chart Checklist

- [x] N/A — no Helm chart changes; this is an e2e test addition only.

## Screenshots

image

## Additional Notes

- `utils/billing.ts` exports `fetchWebClientConfig`, `billingEnabled`, plus the gated `test`/`expect`. The config is fetched once per worker and cached, so only the first test pays for the request.
- Verified: `tsc --noEmit`, `eslint`, and `prettier --check` all pass; `playwright test --list` discovers the test across chromium/firefox/webkit under the returning-user role (and new-user when enabled).
- The local `saas_deploy` checkout wasn't available in this sandbox, so the source `smoke.spec.ts` was pulled from `OpenHands/deploy` on GitHub; the ported test matches that file.

### DIFF
--- a/e2e_tests/tests/billing.spec.ts
+++ b/e2e_tests/tests/billing.spec.ts
@@ -1,0 +1,122 @@
+import { test, expect } from "../utils/billing";
+import { HomePage } from "../pages";
+import { runUser } from "../utils/config";
+
+/**
+ * Billing specs.
+ *
+ * Ported from saas_deploy's `e2e_tests/tests/smoke.spec.ts` (Stripe credit
+ * purchase flow). Every spec here is gated on `feature_flags.enable_billing`
+ * — see `utils/billing.ts`. When that flag is falsy the whole suite is
+ * skipped automatically, so these specs never run against a deployment that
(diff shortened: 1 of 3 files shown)


hieptl (approved):
Thank you! 🙏

PR #1100, opened 2026-08-15
feat: add configurable daily conversation limit to chart

neubig:
Adds an optional dailyConversationLimit Helm value, defaulting to null/unlimited, and wires it to OH_DAILY_CONVERSATION_LIMIT. Includes schema and chart tests.

Coordinates with OpenHands/enterprise#179 and OpenHands/saas-deploy#639.

This pull request was created by an AI agent (OpenHands) on behalf of the user.

### DIFF
--- a/charts/openhands/templates/_env.yaml
+++ b/charts/openhands/templates/_env.yaml
@@ -30,6 +30,10 @@
   value: {{ .Values.appConfig.OPENHANDS_CONVERSATION_VALIDATOR_CLS }}
 - name: OPENHANDS_EXPERIMENT_MANAGER_CLS
   value: {{ .Values.appConfig.OPENHANDS_EXPERIMENT_MANAGER_CLS }}
+{{- if ne .Values.dailyConversationLimit nil }}
+- name: OH_DAILY_CONVERSATION_LIMIT
+  value: {{ .Values.dailyConversationLimit | quote }}
+{{- end }}
 - name: OH_RESOLVER_LABEL
   value: {{ get $integrations "resolverLabel" | default "openhands" | quote }}
 - name: LOG_JSON # Configure structured logging for Google Cloud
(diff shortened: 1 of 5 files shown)


aivong-openhands (approved):
**Code review — APPROVE**

Reviewed against the OpenHands-Cloud chart review guidelines. This is a clean, well-scoped addition that follows existing chart conventions.

### What's good
- **Correct conditional guard.** Using `ne .Values.dailyConversationLimit nil` is the right choice over a truthiness check — a plain `if` would silently drop a `0` value. The tests explicitly cover `0` and `-1` being preserved, which is exactly the kind of edge case that's easy to miss.
- **Follows the established env pattern.** The variable is added to `openhands.env.defaults`, so it flows through the `openhands.env` de-duplication wrapper and consumers can still override it via `.Values.env.OH_DAILY_CONVERSATION_LIMIT` without conflicts. This mirrors the existing conditional-env pattern (e.g. `POSTHOG_CLIENT_KEY`).
- **Consumer-friendly value.** `dailyConversationLimit` names the behavior rather than the internal env var, and the values.yaml comment documents that `null` means unlimited. Matches the "name the effect, not the scenario" principle.
- **Schema and tests.** `values.schema.json` uses `["integer", "null"]` and the schema test asserts that a string is rejected. Good coverage.
- **Sensible default for self-hosted users.** `null`/unlimited is a broadly-applicable default and isn't derived from our dev/staging/prod scale.

### Verification
Ran `helm unittest` locally on the new and changed suites — all pass:

```
Tests: 12 passed, 12 total
(daily_conversation_limit_test.yaml + values_schema_test.yaml)
```

### Nit (non-blocking)
Placement is a top-level key rather than under `appConfig` (which houses the `OPENHANDS_*_CLS` settings). A friendly top-level key is a reasonable — arguably better — choice per the consumer-friendly-config guideline, so no change needed; just flagging for consistency awareness.

Coordinates with OpenHands/enterprise#179 and OpenHands/saas-deploy#639 for the app side that consumes `OH_DAILY_CONVERSATION_LIMIT`; that's out of scope for this chart PR.

---

PR #1108, opened 2026-08-18
fix(e2e): replace networkidle waits and fix Promise.race short-circuit

tofarr:
## Description

Two related fixes to the e2e page objects that eliminate fixed stalls and a premature-timeout bug.

**`HomePage` — replace `networkidle` stall with concrete element waits**

`goto()` and `waitForHomeScreen()` previously relied on `waitForLoadState("networkidle")`, which never fires for this SPA because it holds open WebSocket/SSE connections. Every `homePage.goto()` incurred a fixed ~30s stall.

Replaced with `domcontentloaded` plus waits on concrete page elements:
- `home-screen` testid visible
- `launch-new-conversation-button` visible and enabled

These are the actual readiness signals — the page is interactive when the launch button is enabled, not when the network goes quiet.

**`ConversationPage` — fix `Promise.race` short-circuit in `waitForConversationReady()` and `waitForMessageContaining()`**

Both methods raced a readiness/message wait (120s/180s budget) against an error-banner sentinel (5s budget). The error-banner branch resolved to `"timeout"` on expiry via `.catch(() => "timeout")`, which always won the race after 5 seconds — abandoning the readiness branch with ~115s of unused budget. The failure message misleadingly reported the full timeout (e.g. "120000ms") despite only waiting 5s.

The error-banner branch now stays pending on timeout (`.catch(() => {})` — swallowed, never resolves), so it only resolves the race when an error actually appears. The full timeout budget is left to the readiness/message branch, matching the original "fail fast on errors" intent.

## Helm Chart Checklist

- [x] N/A — no Helm chart changes (e2e test helpers only)

## Additional Notes

Observed during testing: the "should be able to start a conversation" spec failed after 5s at the "Setting up skills" stage because the `Promise.race` short-circuit abandoned the readiness wait prematurely. This fix gives the sandbox cold-start the full timeout budget it needs.


hieptl (approved):
Thank you! 🙏

PR #1112, opened 2026-08-18
fix(automation): keep events service available during node drains

aivong-openhands:
## Description
The automation-events deployment runs two replicas with no PodDisruptionBudget and nothing spreading them across nodes, so a single node drain — such as a node-pool rotation — can evict both pods at once and leave `/v1/events/*` with no healthy backend, the failure mode that recently took the route to zero backends and served ~1505 HTTP 503s. This adds a PodDisruptionBudget (`minAvailable: 1`) and hostname topology spread to the events pods, both gated on `events.enabled`, so installs that don't run the events deployment (the chart default) render exactly as before.

## Helm Chart Checklist

## Additional Notes

- **Upgrade is additive** — rendering the chart on `main` against this branch with `events.enabled=true` yields a single delta: the new PodDisruptionBudget and the deployment's `topologySpreadConstraints`. No other object or field changes, and the new values keys all default, so existing `values.yaml` overrides keep working and there are no new required values.

### DIFF
--- a/charts/openhands/charts/automation/templates/events-deployment.yaml
+++ b/charts/openhands/charts/automation/templates/events-deployment.yaml
@@ -20,6 +20,10 @@ spec:
         app: automation-events
     spec:
       serviceAccountName: {{ .Values.serviceAccount.name }}
+      {{- with .Values.events.topologySpreadConstraints }}
+      topologySpreadConstraints:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       {{- with .Values.securityContext }}
       securityContext:
         {{- toYaml . | nindent 8 }}
--- a/charts/openhands/charts/automation/templates/events-pdb.yaml
+++ b/charts/openhands/charts/automation/templates/events-pdb.yaml
@@ -1,0 +1,19 @@
+{{- if and .Values.events.enabled .Values.events.podDisruptionBudget.enabled }}
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  name: automation-events
+  labels:
+    {{- include "automation.labels" . | nindent 4 }}
+    app.kubernetes.io/component: events
+    app: automation-events
+spec:
+  {{- if .Values.events.podDisruptionBudget.minAvailable }}
+  minAvailable: {{ .Values.events.podDisruptionBudget.minAvailable }}
+  {{- else if .Values.events.podDisruptionBudget.maxUnavailable }}
+  maxUnavailable: {{ .Values.events.podDisruptionBudget.maxUnavailable }}
+  {{- end }}
+  selector:
+    matchLabels:
+      app: automation-events
+{{- end }}
--- a/charts/openhands/charts/automation/templates/events-pdb.yaml
+++ b/charts/openhands/charts/automation/templates/events-pdb.yaml
@@ -1,4 +1,10 @@
-{{- if and .Values.events.enabled .Values.events.podDisruptionBudget.enabled }}
+{{- /*
+Only rendered above one replica: at replicas=1 a minAvailable=1 budget makes the
+single pod unevictable, so the node drain this is meant to survive hangs forever
+instead. On a single-node cluster the same stall is possible at any replica count
+(nowhere to reschedule the evicted pod) — turn the budget off there.
+*/}}
+{{- if and .Values.events.enabled .Values.events.podDisruptionBudget.enabled (gt (int .Values.events.deployment.replicas) 1) }}
 apiVersion: policy/v1
 kind: PodDisruptionBudget
 metadata:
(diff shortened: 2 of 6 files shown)


aivong-openhands (commented):
Chart shape is right: additive, gated on `events.enabled`, template style matches the `device-plugin` `{{- with }}` / `toYaml | nindent` convention, and `helm template` with the shipped defaults renders zero new objects. Three things worth fixing before merge.

### 1. `minAvailable: 0` / `maxUnavailable: 0` silently render a budget-less PDB

`events-pdb.yaml` branches on truthiness, and Go templates treat `0` as false, so both branches fall through:

```console
$ helm template t charts/openhands/charts/automation --set events.enabled=true \
    --set events.podDisruptionBudget.maxUnavailable=0 -s templates/events-pdb.yaml
spec:
  selector:
    matchLabels:
      app: automation-events
```

`maxUnavailable: 0` is exactly the "never evict" value runtime-api already uses for sandbox PDBs in this repo, so it's a value a consumer will plausibly set — and they'd get a PodDisruptionBudget expressing no budget at all instead of the strongest one.

Simplest fix is to drop `maxUnavailable` entirely (nothing asked for it, and `enabled: false` already covers "no budget") and render `minAvailable` unconditionally — one knob, one effect, no branch to get wrong:

```yaml
spec:
  minAvailable: {{ .Values.events.podDisruptionBudget.minAvailable }}
```

### 2. `enabled: true` + `minAvailable: 1` can wedge the drain it's meant to survive

The PDB renders regardless of replica count:

```console
$ helm template t charts/openhands/charts/automation --set events.enabled=true \
    --set events.deployment.replicas=1 -s templates/events-pdb.yaml
spec:
  minAvailable: 1
```

At `replicas: 1` the only pod can never be voluntarily evicted, so a node drain hangs indefinitely. Same outcome on a single-node cluster at the default `replicas: 2` — `whenUnsatisfiable: ScheduleAnyway` lets both pods land on the same node, and the second eviction can never satisfy `minAvailable: 1` because there is nowhere to reschedule.

This repo has already paid for that lesson: `charts/infra/templates/sandbox-upgrade-coordinator.yaml` exists solely because `maxUnavailable=0` sandbox PDBs stall k0s autopilot drains, and that coordinator only relaxes PDBs matching the sandbox selector — it would not touch this one.

Blast radius today is small (`events.enabled` defaults false, and `replicated/openhands.yaml` never turns it on), but the guard is cheap:

```yaml
{{- if and .Values.events.enabled .Values.events.podDisruptionBudget.enabled (gt (int .Values.events.deployment.replicas) 1) }}
```

Worth a line in the values comment too, noting a single-node cluster still needs the operator to turn it off.

### 3. New knobs missing from the parent chart's values mirror

`charts/openhands/values.yaml` (~1136-1147) mirrors the `automation.events` block, and that is the file self-hosted consumers actually read and copy. `podDisruptionBudget` and `topologySpreadConstraints` should appear there as well. Deep-merge means it works without this — it is a discoverability gap, not a bug.

### Nit: no unit test

`charts/openhands/charts/automation/tests/` exists and CI runs `helm unittest` on this chart. A short `events_pdb_test.yaml` — renders when enabled, absent at the default, honours a custom `minAvailable` — would have caught #1.

aivong-openhands (commented):
**Verdict: APPROVE** (posting as a comment — GitHub won't let me approve my own PR). One design question, one test gap, nothing blocking. CI is green; I also rendered the chart and ran `helm unittest charts/openhands/charts/automation` locally (12/12 pass).

### What this gets right

- Both new keys are gated on `events.enabled` (default `false`), so a default install renders identically — confirmed by diffing `helm template` against `main`.
- Canonical Kubernetes vocabulary (`podDisruptionBudget.minAvailable`, verbatim `topologySpreadConstraints`) rather than an invented toggle — passes the "name the effect, not the scenario" test.
- Parent `charts/openhands/values.yaml` mirrors the subchart block, matching the existing convention for the rest of the `automation:` section.
- The `replicas > 1` guard is right: at one replica a `minAvailable: 1` budget makes the pod unevictable and hangs the very drain this exists for.
- `ScheduleAnyway` is the correct choice — `DoNotSchedule` with `maxSkew: 1` would leave the second replica unschedulable on a single-node cluster.

### Question: `podDisruptionBudget.enabled: true` as a shipped default

The justification here is a SaaS incident, which is usually the signal that says "override it in our environment values, not in the chart default."

Concretely: on a single-node cluster with `events.enabled=true`, the defaults give 2 replicas + `minAvailable: 1`, and the second eviction can never be satisfied — the drain stalls. That's the same failure class that already forced `charts/infra/templates/sandbox-upgrade-coordinator.yaml` into existence for k0s autopilot, and that coordinator only patches PDBs matching the sandbox selector, so it would not relax this one.

Blast radius today is small — `events.enabled` defaults to `false` and `replicated/openhands.yaml` never enables it — and the values comment does tell single-node operators to turn it off. So this is a question, not a blocker: should the chart ship `enabled: false` and have our environment values set it true? If it stays `true`, it may be worth a follow-up to widen the coordinator's selector so an embedded-cluster upgrade can't be wedged by an app PDB.

### Test gap: the topology-spread half is untested

`events_pdb_test.yaml` covers the PDB from five angles, but the `events-deployment.yaml` change has no assertion. One case would cover it, including the `{{- with }}` behaviour:

```yaml
  - it: omits topology spread when unset
    set: {events.enabled: true, events.topologySpreadConstraints: []}
    asserts:
      - notExists: {path: spec.template.spec.topologySpreadConstraints}
```

### Nit (no action needed)

The guard reads static values, so a later `kubectl scale`/HPA to more than one replica gets no PDB, and `replicas: 3` with `minAvailable: 3` would still stall. Not worth templating around.

jlav (approved):

dylan-openhands (approved):

Issue #203, opened 2025-09-22
[bug]: search is not working for me for repos


jamiechicago312:
when searching for a repo, nothing is available

<img width="1919" height="1065" alt="Image" src="https://github.com/user-attachments/assets/e7962006-5e63-47c2-bc70-8e00ce41558e" />

<img width="540" height="428" alt="Image" src="https://github.com/user-attachments/assets/5ef379ee-620b-4c4d-9e5e-c90309602bac" />

after signing out and signing back it, it worked and my repos were there

<img width="1118" height="999" alt="Image" src="https://github.com/user-attachments/assets/136e2bf2-3847-463e-8df6-b30f92b3368b" />

PR #1114, opened 2026-08-19
fix: refresh changes panel when empty in VSCode integration test

tofarr:
> _This pull request was created by an AI agent (OpenHands) on behalf of the user._

## Description

The `should be able to select repository and use VSCode integration` E2E test intermittently fails at the `file-diff-viewer-outer` assertion: the chat shows the task completed, but the changes panel renders empty (no data loaded). It works when run manually.

### Root cause (legacy frontend)

The changes panel (`use-unified-get-git-changes.ts`) is repopulated **only** by WebSocket `ActionEvent`s invalidating the `file_changes` cache. There is:
- no polling/timer interval,
- no refresh trigger on `AgentState.FINISHED` (the status the test waits on),
- `retry: false` on the query.

So if the final file-edit `ActionEvent` is missed over the WebSocket, or its refetch transiently fails, the cache holds stale (empty) data and **no amount of waiting** will repopulate it within the test's window. The test keys on the "Agent has finished the task" status text, which does not invalidate `file_changes` — hence the intermittent empty panel.

### Fix

Add a safety net in the test: if the README diff hasn't appeared within 10s after task completion, click the editor tab's refresh button (which calls `refetch()` on the `file_changes` query) once and re-wait. A genuinely empty/broken panel still fails the final `expect(readMe).toBeVisible()` assertion — this only recovers the transient stale-cache case.

The refresh button is targeted via the stable `data-testid="changes-refresh-button"`, which was added to `ConversationTabTitle` in enterprise PR [#203](https://github.com/OpenHands/enterprise/pull/203). This depends on that testid being deployed to the environment before the test can use it.

## Helm Chart Checklist

- [x] N/A — no Helm chart changes; this is an E2E test-only change.

## Additional Notes

- Depends on [enterprise PR #203](https://github.com/OpenHands/enterprise/pull/203) (`changes-refresh-button` testid) being merged and deployed.
- Test typechecks (`tsc --noEmit` clean). Not run against a live staging environment (requires cluster + credentials).
- AI disclosure: this PR was created by an AI agent (OpenHands) on behalf of the user.


hieptl (approved):
Thank you! 🙏

PR #1116, opened 2026-08-19
feat: configure Enterprise SSO for Replicated VM deployments

jpelletier1:
## Description

Adds Replicated Admin Console configuration for Enterprise SSO (SAML) in VM deployments, including optional automatic provisioning of the `enterprise_sso` identity provider in the bundled Keycloak realm.

## What changed

- Adds installer fields for the Enterprise SSO toggle, display name, and HTTPS IdP metadata URL.
- Uses one canonical Helm values path, `enterpriseSSO`, for both login-provider advertising and Keycloak provisioning.
- Sends Keycloak 26.3 the required JSON `ImportConfig` request and consumes its flat configuration-map response.
- Upserts the SAML provider with `validateSignature=true`, trust-email, and the hardcoded `identity_provider=enterprise_sso:saml` mapper.
- Propagates transport, parsing, and API failures through the best-effort reconciliation wrapper instead of reporting false success.
- Preserves `.Values.env` precedence for Enterprise SSO init-container settings without duplicate environment entries.
- Documents the required manual Keycloak mapper in both Helm and Replicated guidance.
- Passes operator-controlled display names and metadata URLs as container environment data and encodes them with `jq`, rather than interpolating them into shell source.
- Validates metadata URLs as HTTPS and disables a chart-managed provider when the Enterprise SSO toggle is turned off.
- Preserves manual Keycloak management when `idpMetadataUrl` is blank.

Provisioning remains best-effort: invalid or unreachable metadata logs a warning without blocking OpenHands startup.

## Helm Chart Checklist

- [x] Defaults remain inert (`enabled: false`, empty metadata URL).
- [x] Manual setup remains available with a blank metadata URL and documents the required mapper contract.
- [x] The chart README documents the canonical values and managed disable behavior.
- [x] Schema and installer validation require HTTPS metadata URLs.

## Verification

- `473 passed` — repository script tests
- `117 passed` — Helm unit tests, including automatic, manual, disabled, and operator-override SSO modes
- Helm lint, Ruff, YAML parsing, and diff checks pass
- Rendered `keycloak-config.sh` passes `sh -n` and `bash -n` in automatic, manual, and disabled modes
- All 22 GitHub checks pass on `b53f613`

_This PR description was updated by an AI agent (OpenHands) on behalf of the user._

### DIFF
--- a/charts/openhands/README.md
+++ b/charts/openhands/README.md
@@ -294,13 +294,15 @@ Enterprise SSO signs users in with a corporate SAML identity provider through th
 2. Update site-values.yaml file:
 
    ```yaml
-   enterpriseSso:
+   enterpriseSSO:
      enabled: true
      displayName: "Company SSO"          # optional, defaults to "Company SSO"
-     idpMetadataUrl: "<your-idp-metadata-url>"
+     idpMetadataUrl: "https://idp.example.com/saml/metadata"
    # When idpMetadataUrl is provided, the chart automatically creates and keeps updated the
    # enterprise_sso SAML identity provider in the bundled Keycloak on every pod start.
    # Leave it empty to configure the provider manually in the Keycloak admin console instead.
+   # When disabling chart-managed SSO, retain idpMetadataUrl for that rollout so the chart
+   # also disables the managed Keycloak provider.
    ```
 
 ### LiteLLM configuration
--- a/charts/openhands/README.md
+++ b/charts/openhands/README.md
@@ -300,9 +300,12 @@ Enterprise SSO signs users in with a corporate SAML identity provider through th
      idpMetadataUrl: "https://idp.example.com/saml/metadata"
    # When idpMetadataUrl is provided, the chart automatically creates and keeps updated the
    # enterprise_sso SAML identity provider in the bundled Keycloak on every pod start.
-   # Leave it empty to configure the provider manually in the Keycloak admin console instead.
-   # When disabling chart-managed SSO, retain idpMetadataUrl for that rollout so the chart
-   # also disables the managed Keycloak provider.
+   # The managed provider validates SAML signatures and trusts the assertion email for
+   # account linking, so only use metadata from an identity provider you trust.
+   # Leave idpMetadataUrl empty to configure the provider manually in the Keycloak admin
+   # console instead. When disabling chart-managed SSO, retain idpMetadataUrl for that
+   # rollout so the chart can distinguish and disable the provider it manages without
+   # modifying a manually managed provider.
    ```
 
 ### LiteLLM configuration
--- a/charts/openhands/README.md
+++ b/charts/openhands/README.md
@@ -282,6 +282,27 @@ Bitbucket Data Center is the self-hosted version of Bitbucket. The setup is diff
      host: <your-bitbucket-data-center-host>
    ```
 
+#### Enterprise SSO (SAML)
+
+Enterprise SSO signs users in with a corporate SAML identity provider through the bundled Keycloak.
+
+1. Register Keycloak with your identity provider using these SAML values:
+
+   - ACS URL `https://auth.openhands.example.com/realms/allhands/broker/enterprise_sso/endpoint`
+   - Entity ID `https://auth.openhands.example.com/realms/allhands`
+
+2. Update site-values.yaml file:
+
+   ```yaml
+   enterpriseSso:
+     enabled: true
+     displayName: "Company SSO"          # optional, defaults to "Company SSO"
+     idpMetadataUrl: "<your-idp-metadata-url>"
+   # When idpMetadataUrl is provided, the chart automatically creates and keeps updated the
+   # enterprise_sso SAML identity provider in the bundled Keycloak on every pod start.
+   # Leave it empty to configure the provider manually in the Keycloak admin console instead.
+   ```
+
 ### LiteLLM configuration
 
 > [!IMPORTANT]
--- a/charts/openhands/README.md
+++ b/charts/openhands/README.md
@@ -308,6 +308,14 @@ Enterprise SSO signs users in with a corporate SAML identity provider through th
    # modifying a manually managed provider.
    ```
 
+   For manual setup, also add this identity-provider mapper to `enterprise_sso`:
+
+   - Name: `identity-provider`
+   - Mapper type: `hardcoded-attribute-idp-mapper`
+   - Attribute: `identity_provider`
+   - Value: `enterprise_sso:saml`
+   - Sync mode: `FORCE`
+
 ### LiteLLM configuration
 
 > [!IMPORTANT]
--- a/charts/openhands/README.md
+++ b/charts/openhands/README.md
@@ -300,12 +300,11 @@ Enterprise SSO signs users in with a corporate SAML identity provider through th
      idpMetadataUrl: "https://idp.example.com/saml/metadata"
    # When idpMetadataUrl is provided, the chart automatically creates and keeps updated the
    # enterprise_sso SAML identity provider in the bundled Keycloak on every pod start.
-   # The managed provider validates SAML signatures and trusts the assertion email for
-   # account linking, so only use metadata from an identity provider you trust.
+   # The managed provider validates SAML signatures, trusts the assertion email for account
+   # linking, and stores an ownership marker in Keycloak. Turning enabled off disables only
+   # a provider with that marker, even if idpMetadataUrl is cleared in the same rollout.
    # Leave idpMetadataUrl empty to configure the provider manually in the Keycloak admin
-   # console instead. When disabling chart-managed SSO, retain idpMetadataUrl for that
-   # rollout so the chart can distinguish and disable the provider it manages without
-   # modifying a manually managed provider.
+   # console instead; the chart does not disable providers without its ownership marker.
    ```
 
    For manual setup, also add this identity-provider mapper to `enterprise_sso`:
--- a/charts/openhands/templates/deployment.yaml
+++ b/charts/openhands/templates/deployment.yaml
@@ -74,6 +74,12 @@ spec:
         image: '{{.Values.image.repository}}:{{.Values.image.tag | default .Chart.AppVersion }}'
         env:
         {{- include "openhands.env" . | nindent 8 }}
+        {{- if and .Values.enterpriseSSO.enabled .Values.enterpriseSSO.idpMetadataUrl }}
+        - name: ENTERPRISE_SSO_DISPLAY_NAME
+          value: {{ .Values.enterpriseSSO.displayName | quote }}
+        - name: ENTERPRISE_SSO_IDP_METADATA_URL
+          value: {{ .Values.enterpriseSSO.idpMetadataUrl | quote }}
+        {{- end }}
         volumeMounts:
         - name: keycloak-config-script
           mountPath: /scripts
--- a/charts/openhands/templates/deployment.yaml
+++ b/charts/openhands/templates/deployment.yaml
@@ -76,7 +76,7 @@ spec:
         {{- include "openhands.env" . | nindent 8 }}
         {{- if and .Values.enterpriseSSO.enabled .Values.enterpriseSSO.idpMetadataUrl }}
         - name: ENTERPRISE_SSO_DISPLAY_NAME
-          value: {{ .Values.enterpriseSSO.displayName | quote }}
+          value: {{ .Values.enterpriseSSO.displayName | default "Company SSO" | quote }}
         - name: ENTERPRISE_SSO_IDP_METADATA_URL
           value: {{ .Values.enterpriseSSO.idpMetadataUrl | quote }}
         {{- end }}
--- a/charts/openhands/templates/deployment.yaml
+++ b/charts/openhands/templates/deployment.yaml
@@ -75,9 +75,10 @@ spec:
         env:
         {{- include "openhands.env" . | nindent 8 }}
         {{- if and .Values.enterpriseSSO.enabled .Values.enterpriseSSO.idpMetadataUrl }}
+        {{- /* Keep these init-only; openhands.env already emits .Values.env overrides. */}}
         {{- if not (hasKey (.Values.env | default dict) "ENTERPRISE_SSO_DISPLAY_NAME") }}
         - name: ENTERPRISE_SSO_DISPLAY_NAME
-          value: {{ .Values.enterpriseSSO.displayName | default "Company SSO" | quote }}
+          value: {{ .Values.enterpriseSSO.displayName | quote }}
         {{- end }}
         {{- if not (hasKey (.Values.env | default dict) "ENTERPRISE_SSO_IDP_METADATA_URL") }}
         - name: ENTERPRISE_SSO_IDP_METADATA_URL
--- a/charts/openhands/templates/deployment.yaml
+++ b/charts/openhands/templates/deployment.yaml
@@ -75,10 +75,14 @@ spec:
         env:
         {{- include "openhands.env" . | nindent 8 }}
         {{- if and .Values.enterpriseSSO.enabled .Values.enterpriseSSO.idpMetadataUrl }}
+        {{- if not (hasKey (.Values.env | default dict) "ENTERPRISE_SSO_DISPLAY_NAME") }}
         - name: ENTERPRISE_SSO_DISPLAY_NAME
           value: {{ .Values.enterpriseSSO.displayName | default "Company SSO" | quote }}
+        {{- end }}
+        {{- if not (hasKey (.Values.env | default dict) "ENTERPRISE_SSO_IDP_METADATA_URL") }}
         - name: ENTERPRISE_SSO_IDP_METADATA_URL
           value: {{ .Values.enterpriseSSO.idpMetadataUrl | quote }}
+        {{- end }}
         {{- end }}
         volumeMounts:
         - name: keycloak-config-script
--- a/charts/openhands/templates/keycloak-config-script.yaml
+++ b/charts/openhands/templates/keycloak-config-script.yaml
@@ -19,6 +19,32 @@ data:
         exit 1
       fi
     }
+
+    refresh_access_token() {
+      ACCESS_TOKEN=$(curl -sS -X POST "$KEYCLOAK_SERVER_URL/realms/master/protocol/openid-connect/token" \
+        -H "Content-Type: application/x-www-form-urlencoded" \
+        -d "client_id=admin-cli" \
+        -d "grant_type=password" \
+        -d "username=admin" \
+        -d "password=$KEYCLOAK_ADMIN_PASSWORD" | jq -r '.access_token // empty')
+      if [ -z "$ACCESS_TOKEN" ]; then
+        echo "ERROR: could not refresh the Keycloak admin access token." >&2
+        return 1
+      fi
+    }
+
+    keycloak_get_optional() {
+      URL=$1
+      STATUS=$(curl -sS -o /tmp/keycloak-get-response.json -w '%{http_code}' "$URL" \
+        -H "Authorization: Bearer $ACCESS_TOKEN")
+      RESPONSE=$(cat /tmp/keycloak-get-response.json)
+      case "$STATUS" in
+        200) return 0 ;;
+        404) RESPONSE=""; return 0 ;;
+        *) echo "ERROR: Keycloak GET $URL returned HTTP $STATUS: $RESPONSE" >&2; return 1 ;;
+      esac
+    }
+
     echo "Waiting for Keycloak to be ready..."
     until curl --output /dev/null --silent --head --fail $KEYCLOAK_SERVER_URL; do
         echo '.'
@@ -240,6 +266,7 @@ data:
     # failure here (e.g. an unreachable metadata URL) warns but never blocks the
     # app from starting.
     (
+      refresh_access_token
       KC_REALM="$KEYCLOAK_SERVER_URL/admin/realms/$KEYCLOAK_REALM_NAME"
       AUTH="-H \"Authorization: Bearer $ACCESS_TOKEN\""
       CT="-H \"Content-Type: application/json\""
@@ -247,7 +274,6 @@ data:
       echo "Importing enterprise SSO IdP metadata from the configured URL..."
       IMPORT_REQUEST=$(jq -n --arg from_url "$ENTERPRISE_SSO_IDP_METADATA_URL" '{
         providerId: "saml",
-        alias: "enterprise_sso",
         fromUrl: $from_url
       }')
       IMPORT_RESPONSE=$(curl -sS -X POST "$KC_REALM/identity-provider/import-config" \
@@ -255,9 +281,9 @@ data:
         -H "Content-Type: application/json" \
         --data "$IMPORT_REQUEST")
       if ! echo "$IMPORT_RESPONSE" | jq -e \
-        'type == "object" and has("idpEntityId") and has("singleSignOnServiceUrl")' \
+        'type == "object" and has("idpEntityId") and has("singleSignOnServiceUrl") and has("signingCertificate")' \
         >/dev/null 2>&1; then
-        echo "ERROR: could not import IdP metadata: $IMPORT_RESPONSE" >&2
+        echo "ERROR: IdP metadata must include an entity ID, SSO service URL, and signing certificate: $IMPORT_RESPONSE" >&2
         echo "The enterprise_sso identity provider was not configured. Fix the metadata URL and redeploy to retry." >&2
         exit 1
       fi
@@ -268,7 +294,7 @@ data:
           alias: "enterprise_sso",
           providerId: "saml",
           enabled: true,
-          displayName: (if $display == "" then "Company SSO" else $display end),
+          displayName: $display,
           updateProfileFirstLoginMode: "on",
           trustEmail: true,
           storeToken: false,
@@ -279,8 +305,8 @@ data:
           config: (. + {"syncMode": "IMPORT", "validateSignature": "true"})
         }' > /tmp/idp-enterprise-sso.json
 
-      EXISTING_IDP=$(curl -s "$KC_REALM/identity-provider/instances/enterprise_sso" \
-        -H "Authorization: Bearer $ACCESS_TOKEN")
+      keycloak_get_optional "$KC_REALM/identity-provider/instances/enterprise_sso"
+      EXISTING_IDP=$RESPONSE
       if echo "$EXISTING_IDP" | jq -e '.internalId or .id' >/dev/null 2>&1; then
         keycloak_api_call "curl -s -X PUT \"$KC_REALM/identity-provider/instances/enterprise_sso\" $AUTH $CT --data \"@/tmp/idp-enterprise-sso.json\""
         echo "  Updated identity provider: enterprise_sso"
@@ -299,8 +325,8 @@ data:
           syncMode: "FORCE"
         }
       }' > /tmp/mapper-enterprise-sso.json
-      EXISTING_MAPPER_ID=$(curl -s "$KC_REALM/identity-provider/instances/enterprise_sso/mappers" \
-        -H "Authorization: Bearer $ACCESS_TOKEN" | jq -r \
+      keycloak_api_call "curl -s \"$KC_REALM/identity-provider/instances/enterprise_sso/mappers\" $AUTH"
+      EXISTING_MAPPER_ID=$(echo "$RESPONSE" | jq -r \
         '.[] | objects | select(.name == "identity-provider") | .id // empty')
       if [ -n "$EXISTING_MAPPER_ID" ]; then
         jq --arg id "$EXISTING_MAPPER_ID" '. + {id: $id}' /tmp/mapper-enterprise-sso.json > /tmp/mapper-enterprise-sso-update.json
@@ -317,11 +343,12 @@ data:
     # when the feature toggle is turned off so direct kc_idp_hint requests cannot
     # bypass the hidden login button.
     (
+      refresh_access_token
       KC_REALM="$KEYCLOAK_SERVER_URL/admin/realms/$KEYCLOAK_REALM_NAME"
       AUTH="-H \"Authorization: Bearer $ACCESS_TOKEN\""
       CT="-H \"Content-Type: application/json\""
-      EXISTING_IDP=$(curl -s "$KC_REALM/identity-provider/instances/enterprise_sso" \
-        -H "Authorization: Bearer $ACCESS_TOKEN")
+      keycloak_get_optional "$KC_REALM/identity-provider/instances/enterprise_sso"
+      EXISTING_IDP=$RESPONSE
       if echo "$EXISTING_IDP" | jq -e '.internalId or .id' >/dev/null 2>&1; then
         echo "$EXISTING_IDP" | jq '.enabled = false' > /tmp/idp-enterprise-sso-disabled.json
         keycloak_api_call "curl -s -X PUT \"$KC_REALM/identity-provider/instances/enterprise_sso\" $AUTH $CT --data \"@/tmp/idp-enterprise-sso-disabled.json\""
--- a/charts/openhands/templates/keycloak-config-script.yaml
+++ b/charts/openhands/templates/keycloak-config-script.yaml
@@ -232,7 +232,7 @@ data:
       echo "Updated allhands realm configuration."
     fi
 
-    {{- if and .Values.enterpriseSso.enabled .Values.enterpriseSso.idpMetadataUrl }}
+    {{- if and .Values.enterpriseSSO.enabled .Values.enterpriseSSO.idpMetadataUrl }}
     # Enterprise SSO: upsert the enterprise_sso SAML identity provider from the
     # operator-supplied IdP metadata URL, plus the identity_provider user
     # attribute mapper the app relies on to detect SAML logins. Runs after the
@@ -244,21 +244,27 @@ data:
       AUTH="-H \"Authorization: Bearer $ACCESS_TOKEN\""
       CT="-H \"Content-Type: application/json\""
 
-      echo "Importing enterprise SSO IdP metadata from {{ .Values.enterpriseSso.idpMetadataUrl }}..."
-      IMPORT_RESPONSE=$(curl -s -X POST "$KC_REALM/identity-provider/import-config" \
+      echo "Importing enterprise SSO IdP metadata from the configured URL..."
+      IMPORT_REQUEST=$(jq -n --arg from_url "$ENTERPRISE_SSO_IDP_METADATA_URL" '{
+        providerId: "saml",
+        alias: "enterprise_sso",
+        fromUrl: $from_url
+      }')
+      IMPORT_RESPONSE=$(curl -sS -X POST "$KC_REALM/identity-provider/import-config" \
         -H "Authorization: Bearer $ACCESS_TOKEN" \
-        -H "Content-Type: application/x-www-form-urlencoded" \
-        --data-urlencode "fromUrl={{ .Values.enterpriseSso.idpMetadataUrl }}" \
-        -d "providerId=saml")
-      if ! echo "$IMPORT_RESPONSE" | jq -e '.config' >/dev/null 2>&1; then
+        -H "Content-Type: application/json" \
+        --data "$IMPORT_REQUEST")
+      if ! echo "$IMPORT_RESPONSE" | jq -e \
+        'type == "object" and has("idpEntityId") and has("singleSignOnServiceUrl")' \
+        >/dev/null 2>&1; then
         echo "ERROR: could not import IdP metadata: $IMPORT_RESPONSE" >&2
         echo "The enterprise_sso identity provider was not configured. Fix the metadata URL and redeploy to retry." >&2
         exit 1
       fi
 
       echo "$IMPORT_RESPONSE" | jq \
-        --arg display "{{ .Values.enterpriseSso.displayName }}" \
-        '.config as $cfg | {
+        --arg display "$ENTERPRISE_SSO_DISPLAY_NAME" \
+        '{
           alias: "enterprise_sso",
           providerId: "saml",
           enabled: true,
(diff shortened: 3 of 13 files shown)


aivong-openhands (commented):
Reviewed against the OpenHands-Cloud Helm chart guidelines. The Keycloak auto-provisioning logic itself is solid, but there's an important configuration-naming problem that will bite self-hosted consumers.

## Important - two near-identical values keys (`enterpriseSso` vs `enterpriseSSO`)

This PR adds a **new** key `enterpriseSso` (lowercase `sso`) at `charts/openhands/values.yaml:117`, but the chart already ships `enterpriseSSO` (capital `SSO`) at `charts/openhands/values.yaml:224` (added in #325). YAML keys are case-sensitive, so these are two independent toggles that differ only by capitalization:

- `enterpriseSSO.enabled` (existing) -> `templates/_env.yaml` uses `.Values.enterpriseSSO.enabled` to add `"enterprise_sso"` to `OH_WEB_CLIENT_PROVIDERS_CONFIGURED`.
- `enterpriseSso.enabled` + `idpMetadataUrl` (new) -> gates the Keycloak SAML IdP auto-provisioning in `keycloak-config-script.yaml`.

Having `enterpriseSso` and `enterpriseSSO` side by side in the same `values.yaml` is a serious foot-gun: a consumer who sets one will almost certainly assume they set the other. This runs against the "consumer-friendly configuration / one knob, one effect" principle. I'd strongly suggest consolidating onto the existing `enterpriseSSO` key (e.g. add `displayName` and `idpMetadataUrl` under it) so a single toggle drives both the web-client provider advertisement and the Keycloak auto-config.

## Important - consumer split-brain for raw-Helm installs

Because of the two keys, enabling SSO end-to-end via the raw chart currently requires setting **both** `enterpriseSSO.enabled: true` (so the provider is advertised to the web client) **and** `enterpriseSso.enabled: true` + `idpMetadataUrl` (so Keycloak is configured). The new `README.md` section only documents the lowercase `enterpriseSso` key, so a consumer who follows it will get Keycloak provisioned but the provider may never be advertised to the web client. Please make the documented path a single, complete toggle.

Related: `replicated/openhands.yaml` wires the new lowercase `enterpriseSso` block and `ENABLE_ENTERPRISE_SSO`, but never sets the capital `enterpriseSSO`, so `OH_WEB_CLIENT_PROVIDERS_CONFIGURED` won't include `enterprise_sso` on Replicated installs.

## Question - `ENABLE_ENTERPRISE_SSO` vs `OH_WEB_CLIENT_PROVIDERS_CONFIGURED`

The button appears to be driven by `ENABLE_ENTERPRISE_SSO` (presence-checked, set only in the Replicated layer), while `OH_WEB_CLIENT_PROVIDERS_CONFIGURED` is driven by the capital `enterpriseSSO.enabled`. Are both needed? Note that for pure-Helm consumers nothing sets `ENABLE_ENTERPRISE_SSO`, so it would help to clarify which mechanism is authoritative and ensure the raw chart covers it.

## Looks good (no action needed)

- Failure-safety is correct: the provisioning runs in a subshell terminated by `|| echo "WARNING..."`, so an unreachable/invalid metadata URL warns without blocking pod startup, and the internal `exit 1` only exits the subshell.
- `validateSignatures: "true"`, `trustEmail` documented, and `hideOnLogin: true` (valid as a top-level field on the bundled Keycloak - bitnami chart 24.7.5) are all sensible defaults.
- The upsert style (existence check -> PUT/POST, `keycloak_api_call`) matches the existing realm/IdP/mapper loop vocabulary, and a separate block is justified since the SAML metadata is imported dynamically and can't live in the static realm template.
- The Replicated env mapping that renders `true` vs an empty string (rather than the literal `false`) correctly matches the app's presence-check semantics and mirrors the existing conditional-literal pattern in this file.

aivong-openhands (commented):
Reviewed at `349b9243`.

Verified locally before writing this:

- `helm unittest` on the new `enterprise_sso_test.yaml` plus the changed `values_schema_test.yaml`: 13 passed.
- `pytest scripts/test_keycloak_realm_template.py`: 17 passed, 1 pre-existing failure (`test_keycloak_identity_provider_socket_timeout`, needs built subcharts, unrelated to this PR).
- Rendered `keycloak-config.sh` for both the auto-config and the disable branch and ran `sh -n` and `bash -n` on each. Both clean.
- Ran the extracted identity-provider `jq` filter through real `jq` against a metadata import response with no signing certificate.
- Confirmed the bundled Keycloak is 26.3.0 (bitnami chart 24.7.5), so `hideOnLogin` as a top-level field is correct for this version.

Nice work on passing the operator inputs as container env and encoding them with `jq --arg`. That is the right call, and the jq-filter test is the one new test I would keep as is.

Three things I think are worth changing before this merges.

## 1. `ENABLE_ENTERPRISE_SSO` is only set on the Replicated path

`replicated/openhands.yaml` is the only place it is set. `charts/openhands/templates/_env.yaml` adds `enterprise_sso` to `OH_WEB_CLIENT_PROVIDERS_CONFIGURED`, but nothing sets `ENABLE_ENTERPRISE_SSO`. The enterprise server also builds its own `providers_configured` list from `ENABLE_ENTERPRISE_SSO`, and that path does not read the `OH_WEB_CLIENT_*` form. So a plain Helm install that sets `enterpriseSSO.enabled: true` can get the identity provider created in Keycloak and still show no login button.

The chart's existing convention for this lives in `_env.yaml`: `ENABLE_JIRA`, `ENABLE_JIRA_DC`, and `ENABLE_LINEAR` are all `{{- if .Values.<x>.enabled }}` blocks with `value: "true"`. Following that here fixes the self-hosted case and lets you drop the `repl{{ if }}...repl{{ end }}` form.

One related detail: the current form renders `value: ""` when the toggle is off, not an absent variable. I confirmed the env dedup wrapper emits the empty entry. It is harmless today because the app strips the value before testing it, but the comment's "presence-checked" wording does not match that.

## 2. `validateSignature` is forced to `"true"` even with no signing certificate

The filter ends with `config: (. + {"syncMode": "IMPORT", "validateSignature": "true"})`, and the import guard only checks `idpEntityId` and `singleSignOnServiceUrl`. Running the filter against a certificate-less import response gives:

```json
"config": {
  "idpEntityId": "https://idp/e",
  "singleSignOnServiceUrl": "https://idp/sso",
  "postBindingResponse": "true",
  "validateSignature": "true",
  "syncMode": "IMPORT"
}
```

No `signingCertificate`. Keycloak sets `validateSignature` itself based on what it finds in the descriptor, and this overrides that to `true` with nothing to validate against. The init container then logs `enterprise_sso SAML identity provider configured.` and the pod goes Ready, but every login fails at signature verification. Adding `has("signingCertificate")` to the guard would catch it.

Worth considering instead: set `useMetadataDescriptorUrl: "true"` and `metadataDescriptorUrl` and let Keycloak fetch the descriptor itself. That removes the import call and its response validation entirely. It also fixes a second problem. Today the certificate is pinned at pod-start time, so an identity provider certificate rotation breaks SSO until someone restarts the pod.

## 3. No guard for a metadata URL with no chart-provisioned Keycloak

`keycloak-config-script.yaml` renders only when `openhands.keycloakProvisionRealm` is true, which needs `keycloak.enabled` or `keycloak.provisionRealm`. With both false and an external Keycloak, `idpMetadataUrl` passes the schema and silently does nothing. `templates/validations.yaml` is the established place for this, and the guard is a couple of lines.

## Smaller points

**The disable path keys off a retained `idpMetadataUrl`.** On Replicated that works, since KOTS keeps a hidden item's stored value, and there is already a comment in `replicated/openhands.yaml` saying so. For a chart consumer the natural action is to set `enabled: false` and delete the URL, which skips the branch. The provider stays enabled and `kc_idp_hint=enterprise_sso` keeps working, which is the bypass the branch exists to prevent. The README asks operators to remember to retain the URL for one rollout, and that is easy to get wrong. What is the reason for gating on `idpMetadataUrl` rather than reconciling whenever `enterpriseSSO.enabled` is false? That needs no marker value at all.

**Failures are invisible, and the reused token makes them more likely.** The subshell swallows everything, the pod goes Ready, and the only signal is one init-container log line. The block also reuses the `ACCESS_TOKEN` minted before the whole realm sync of clients, identity providers, and mappers. Master `admin-cli` tokens are short-lived, so a slow realm update turns this into a 401 that fails the import guard, and the result is a healthy-looking deploy with no identity provider. Re-minting the token at the top of the block is cheap. The disable branch has the same issue and is quieter still: a 401 body has no `internalId`, so it takes the "does not exist" path and prints nothing.

**`trustEmail: true` is hardcoded.** That means the SAML assertion's email is trusted for auto-linking to existing accounts, including accounts created through GitHub OAuth, with no verification step. That is a reasonable default for corporate SSO, but it is a trust decision made on the operator's behalf and it is not mentioned in the README or the installer help text.

**The `"Company SSO"` fallback lives in the jq filter** while `values.yaml` has `displayName: ""`. Putting the default in `values.yaml`, or using `| default "Company SSO"` in `deployment.yaml`, makes it visible to chart consumers and gives one source of truth. `replicated/config.yaml` could then just carry `default: "Company SSO"`.

**Some of the new tests assert the diff back to itself.** The exact `values.yaml` key order and formatting, `"enterpriseSso" not in ...`, `"validateSignatures" not in ...`, and the exact regex string in `config.yaml` all guard typos that were never on `main`, and they break on a harmless reformat. Meanwhile nothing asserts that the `ENTERPRISE_SSO_*` env vars land on the `keycloak-config` init container. The pytest test greps template text, and the helm-unittest suite only checks `containers[0]`. One assert on `initContainers[?(@.name=="keycloak-config")].env` would catch a real regression, and a certificate-less metadata case for the jq filter would be worth more than the string greps.

**Two tiny ones.** `IMPORT_REQUEST` sends `alias`, which `import-config` ignores. And the `values.schema.json` pattern `^$|^https://` is looser than the installer regex, since it accepts a bare `https://` and embedded whitespace.

charts/openhands/templates/keycloak-config-script.yaml
```diff
@@ -231,4 +257,103 @@ data:
...
+      if [ -n "$EXISTING_MAPPER_ID" ]; then
+        jq --arg id "$EXISTING_MAPPER_ID" '. + {id: $id}' /tmp/mapper-enterprise-sso.json > /tmp/mapper-enterprise-sso-update.json
+        keycloak_api_call "curl -s -X PUT \"$KC_REALM/identity-provider/instances/enterprise_sso/mappers/$EXISTING_MAPPER_ID\" $AUTH $CT --data \"@/tmp/mapper-enterprise-sso-update.json\""
+        echo "  Updated mapper: enterprise_sso/identity-provider"
+      else
+        keycloak_api_call "curl -s -X POST \"$KC_REALM/identity-provider/instances/enterprise_sso/mappers\" $AUTH $CT --data \"@/tmp/mapper-enterprise-sso.json\""
+        echo "  Created mapper: enterprise_sso/identity-provider"
+      fi
+      echo "enterprise_sso SAML identity provider configured."
+    ) || echo "WARNING: enterprise_sso auto-configuration failed (details above). Fix the SAML metadata URL and redeploy to retry; continuing startup." >&2
```
aivong-openhands, line 386: **Important — the best-effort wrapper disables the failure propagation it relies on.** Because this whole subshell is the left side of `||`, `set -e` does not stop on intermediate failures inside it: `refresh_access_token` and `keycloak_get_optional` can return nonzero and execution continues. `keycloak_api_call` compounds this by using `export RESPONSE=$(eval $COMMAND)`, whose status is the `export`, so a curl transport failure is also masked. I reproduced the same control flow with `set -e; (export RESPONSE=$(false); echo configured) || echo warned`, which prints `configured`. The auto path can therefore print `enterprise_sso ... configured` after failed writes, and the disable path can silently skip its warning after failed auth/GET. Please explicitly propagate failures (for example, test each helper/curl and `exit 1` from the subshell), preserve command-substitution status, and add an executable failing-API test.

charts/openhands/templates/deployment.yaml
```diff
@@ -74,6 +74,12 @@ spec:
         image: '{{.Values.image.repository}}:{{.Values.image.tag | default .Chart.AppVersion }}'
         env:
         {{- include "openhands.env" . | nindent 8 }}
+        {{- if and .Values.enterpriseSSO.enabled .Values.enterpriseSSO.idpMetadataUrl }}
+        - name: ENTERPRISE_SSO_DISPLAY_NAME
+          value: {{ .Values.enterpriseSSO.displayName | default "Company SSO" | quote }}
+        - name: ENTERPRISE_SSO_IDP_METADATA_URL
+          value: {{ .Values.enterpriseSSO.idpMetadataUrl | quote }}
```
aivong-openhands, line 85: **Important — these entries bypass the chart's environment de-duplication contract.** `openhands.env` filters defaults against `.Values.env` and appends operator overrides last, but these two generated variables are appended after that completed list. A consumer setting `env.ENTERPRISE_SSO_IDP_METADATA_URL` or `env.ENTERPRISE_SSO_DISPLAY_NAME` gets duplicate names, with the generated value last; this both defeats the documented override pattern and can recreate strategic-merge upgrade failures. Please route these init-container defaults through the same de-duplication logic (or explicitly filter colliding keys before appending), and test that an override renders exactly one entry with the intended value.

charts/openhands/README.md
```diff
@@ -282,6 +273,32 @@ Bitbucket Data Center is the self-hosted version of Bitbucket. The setup is diff
...
+   ```yaml
+   enterpriseSSO:
+     enabled: true
+     displayName: "Company SSO"          # optional, defaults to "Company SSO"
+     idpMetadataUrl: "https://idp.example.com/saml/metadata"
+   # When idpMetadataUrl is provided, the chart automatically creates and keeps updated the
+   # enterprise_sso SAML identity provider in the bundled Keycloak on every pod start.
+   # The managed provider validates SAML signatures and trusts the assertion email for
+   # account linking, so only use metadata from an identity provider you trust.
+   # Leave idpMetadataUrl empty to configure the provider manually in the Keycloak admin
```
aivong-openhands, line 306: **Important — the advertised manual path omits a required mapper.** The provisioning script says OpenHands relies on the `identity_provider` mapper and creates `identity_provider=enterprise_sso:saml`; every existing broker in the realm template has the same hardcoded-attribute pattern. These instructions only tell an operator to create the `enterprise_sso` IdP, so following them does not produce the contract automatic mode provides. Please document the exact mapper (name `identity-provider`, type `hardcoded-attribute-idp-mapper`, attribute `identity_provider`, value `enterprise_sso:saml`, sync mode `FORCE`) in both Helm and Replicated manual-setup guidance, or reconcile that mapper independently whenever SSO is enabled while leaving the IdP itself manually managed.

aivong-openhands (commented):
🔴 **Needs improvement** — the overall design is reasonable, but three important failure/operability gaps need attention.

> **Current status:** The inline findings were revalidated on head `a99f064`; the merge from `main` did not change the Enterprise SSO implementation.

**[CRITICAL ISSUES]**
- The best-effort shell wrapper can suppress failed Keycloak operations and report success.
- Init-container Enterprise SSO variables bypass the chart's explicit environment de-duplication/override contract.
- The documented manual setup path omits the mapper this change says OpenHands relies on.

**[RISK ASSESSMENT]**
Defaults are inert and CI is green, but this changes authentication provisioning. The identified paths can leave SSO unconfigured or partially configured while deployment appears healthy.

**VERDICT:**
❌ **Needs rework**: Address the three inline issues before merging.

**KEY INSIGHT:**
Authentication provisioning must fail observably and preserve the chart's established override contract, even when startup remains best-effort.

---

>
>
>

---

aivong-openhands (commented):
🔴 **Needs improvement**

Re-reviewed at `a99f064` after the merge from `main`. The merge adds no changes to the effective Enterprise SSO PR diff, and the three existing review threads remain current and unresolved:

1. [Keycloak failures can be masked while provisioning reports success](https://github.com/OpenHands/OpenHands-Cloud/pull/1116#discussion_r3857959081).
2. [Enterprise SSO init env vars bypass de-duplication and operator override precedence](https://github.com/OpenHands/OpenHands-Cloud/pull/1116#discussion_r3857959085).
3. [Manual setup guidance omits the required `identity_provider=enterprise_sso:saml` mapper](https://github.com/OpenHands/OpenHands-Cloud/pull/1116#discussion_r3857959087).

**[RISK ASSESSMENT]**
Defaults are inert and all current CI checks pass, but this changes authentication provisioning. The unresolved paths can leave SSO unconfigured or partially configured while the deployment appears healthy.

**VERDICT:**
❌ **Needs rework**: Address the three linked threads before merging.

**KEY INSIGHT:**
Best-effort startup must not turn failed authentication reconciliation into apparent success.

---

>
>
>

---

charts/openhands/README.md
aivong-openhands: Agreed on the gap. The manual path does need the mapper documented, in both this README and the Replicated help text.

One adjustment before we write `enterprise_sso:saml` into the docs. That value has an app-side problem today, and the fix does not belong on the mapper.

Two places in the app read the `identity_provider` attribute:

1. **The login handler** splits on the last `:` and treats a `saml` suffix as the signal to skip the OAuth broker token fetch. SAML providers have no OAuth tokens to retrieve, so the suffix is load-bearing here.
2. **The user-authorization lookup** passes the attribute straight through to a query that compares the provider string for exact equality.

The second one never sees the split. So a login carrying `enterprise_sso:saml` does not match an authorization rule scoped to `enterprise_sso`. Only rules with a null provider apply. An operator who scopes an email allowlist to this provider would get their SSO users blocked at login.

Since the suffix is required by the login handler, changing the mapper value to a bare `enterprise_sso` is not an option. Every other broker in the realm template uses a bare value, but none of them are SAML, so none of them hit this. The fix is to normalize the suffix on the authorization lookup path too, so both readers agree on what the provider is.

Suggested order:

1. Normalize the provider string on the authorization lookup path, so `enterprise_sso:saml` matches an `enterprise_sso` rule.
2. Then document the mapper here and in the Replicated help text, using the value the chart already writes.

That order keeps the value in one place in the docs. Documenting it first means it is in three places by the time the app-side change lands.

charts/openhands/templates/keycloak-config-script.yaml
saurya: Fixed in c1ba37d. Keycloak helper functions now preserve transport and parsing failures, and every fallible operation in both best-effort Enterprise SSO subshells explicitly exits on failure. The executable regression invokes the real helper with a failing command and verifies a nonzero result. _This reply was generated by an AI agent (OpenHands) on behalf of the user._

charts/openhands/templates/deployment.yaml
saurya: Fixed in c1ba37d. Generated Enterprise SSO init-container values are now omitted when the same key is present in `.Values.env`, preserving the existing override precedence and rendering one entry with the operator value. Helm and rendered-container regressions cover both variables. _This reply was generated by an AI agent (OpenHands) on behalf of the user._

charts/openhands/README.md
saurya: Fixed in c1ba37d. The Helm README and Replicated installer guidance now document the complete manual mapper contract: `identity-provider`, `hardcoded-attribute-idp-mapper`, `identity_provider`, `enterprise_sso:saml`, and `FORCE`. The provider-suffix authorization normalization noted in the follow-up remains an app-side concern outside this deployment chart change. _This reply was generated by an AI agent (OpenHands) on behalf of the user._

charts/openhands/templates/keycloak-config-script.yaml
```diff
@@ -12,13 +12,59 @@ data:
...
+      fi
+      export RESPONSE
+      if ! ERROR=$(echo "$RESPONSE" | jq -r 'try if type == "array" then (.[0].error // .[0].errorMessage) else (.error // .errorMessage) end'); then
+        echo "ERROR: could not parse Keycloak API response: $RESPONSE" >&2
+        return 1
+      fi
       if [ -n "$ERROR" ] && [ "null" != "$ERROR" ]; then
         echo "Error from Keycloak: $RESPONSE"
-        exit 1
+        return 1
```
aivong-openhands, line 26: **Important — an empty response body still reads as success.** Verified locally: `keycloak_api_call "printf ''"` returns 0. The HTML-error-page case is now caught by the new parse guard, but because every call site uses `curl -s` without `-f` or `-w '%{http_code}'`, the HTTP status is thrown away — and a successful `PUT /identity-provider/instances/{alias}` returns 204 with an empty body, which is byte-identical to a 403 that returns nothing.

Concretely, the four SSO write calls below can print `Created identity provider: enterprise_sso` / `Updated mapper: enterprise_sso/identity-provider` when Keycloak wrote nothing. Because the SSO block is intentionally best-effort, that failure never reaches the operator: the pod starts, the app advertises the login button via `ENABLE_ENTERPRISE_SSO`, and sign-in dead-ends with no error anywhere.

`keycloak_get_optional`, added a few lines down in this same PR, already captures the status with `-w '%{http_code}'` and branches on it. Reusing that shape for the write calls (or teaching this helper to take a status) would close the gap without inventing a new pattern.

charts/openhands/templates/keycloak-config-script.yaml
```diff
@@ -231,4 +277,104 @@ data:
...
+        jq --arg id "$EXISTING_MAPPER_ID" '. + {id: $id}' /tmp/mapper-enterprise-sso.json > /tmp/mapper-enterprise-sso-update.json || exit 1
+        keycloak_api_call "curl -s -X PUT \"$KC_REALM/identity-provider/instances/enterprise_sso/mappers/$EXISTING_MAPPER_ID\" $AUTH $CT --data \"@/tmp/mapper-enterprise-sso-update.json\"" || exit 1
+        echo "  Updated mapper: enterprise_sso/identity-provider"
+      else
+        keycloak_api_call "curl -s -X POST \"$KC_REALM/identity-provider/instances/enterprise_sso/mappers\" $AUTH $CT --data \"@/tmp/mapper-enterprise-sso.json\"" || exit 1
+        echo "  Created mapper: enterprise_sso/identity-provider"
+      fi
+      echo "enterprise_sso SAML identity provider configured."
+    ) || echo "WARNING: enterprise_sso auto-configuration failed (details above). Fix the SAML metadata URL and redeploy to retry; continuing startup." >&2
+    {{- else if .Values.enterpriseSSO.idpMetadataUrl }}
```
aivong-openhands: **Important — `idpMetadataUrl` is overloaded as an ownership marker.** This branch infers "the chart manages this provider" purely from the metadata URL still being set, which forces the README to tell operators to retain a URL they no longer want used. A key named for *where to fetch metadata* now also means *who owns the provider* — two effects on one knob, and the second isn't predictable from the name.

The failure mode is concrete: clear the URL and flip the toggle off in the same rollout — the intuitive action, and the one the KOTS UI encourages since `when:` hides the field when the toggle is off — and neither branch renders. The provider stays `enabled: true` in Keycloak and remains reachable through a direct `kc_idp_hint=enterprise_sso` request, which is precisely what this block exists to prevent.

An explicit marker would be both safer and self-describing: stamp something like `"openhands.managed": "true"` into the provider's `config` when creating it, then key the disable path off the provider's own state rather than off values that the operator may legitimately clear. It also survives the KOTS question below.

Separately: can you confirm `ConfigOption "enterprise_sso_idp_metadata_url"` still returns the stored value once `when:` hides the field? If KOTS yields empty for hidden items, this branch never renders in Replicated installs — the deployment target this PR is written for.

charts/openhands/templates/deployment.yaml
```diff
@@ -74,6 +74,16 @@ spec:
...
         env:
         {{- include "openhands.env" . | nindent 8 }}
+        {{- if and .Values.enterpriseSSO.enabled .Values.enterpriseSSO.idpMetadataUrl }}
+        {{- if not (hasKey (.Values.env | default dict) "ENTERPRISE_SSO_DISPLAY_NAME") }}
+        - name: ENTERPRISE_SSO_DISPLAY_NAME
+          value: {{ .Values.enterpriseSSO.displayName | default "Company SSO" | quote }}
+        {{- end }}
+        {{- if not (hasKey (.Values.env | default dict) "ENTERPRISE_SSO_IDP_METADATA_URL") }}
+        - name: ENTERPRISE_SSO_IDP_METADATA_URL
+          value: {{ .Values.enterpriseSSO.idpMetadataUrl | quote }}
```
aivong-openhands, line 85: Minor: these `hasKey` guards re-implement the `.Values.env` precedence contract that `openhands.env` already owns, so there are now two copies of that rule and they can drift. The rendered output is correct today — the override test exercises real `helm template` output and I confirmed it — so this is a maintainability note, not a bug.

If keeping them outside `openhands.env.defaults` is deliberate (presumably so SSO env doesn't leak into the app and litellm containers, which don't need it), a one-line comment saying so would save the next reader the trip through `_env.yaml`.

aivong-openhands (commented):
Reviewed at `c1ba37d` against the OpenHands-Cloud Helm chart guidelines. All 22 checks are green and I re-verified the shell/jq behaviour locally rather than reading it off the diff.

The overall shape is good: one canonical `enterpriseSSO` values key, HTTPS enforced in both the JSON schema and the KOTS regex, operator input passed as container env and encoded with `jq --arg` (I confirmed a display name of `Company "Platform" $(touch /tmp/should-not-run)` round-trips as data), inert defaults, and a render guard that fails fast when `idpMetadataUrl` is set without chart-managed realm provisioning. The earlier review rounds clearly moved this a long way.

Two **important** items remain. Neither blocks the design, but both are the kind of thing that turns into a support ticket rather than a failed deploy.

**[IMPORTANT]**

1. **`keycloak_api_call` still reports success for an empty response body**, so the SSO write paths can log "Created identity provider: enterprise_sso" when nothing was written. Verified locally: `keycloak_api_call "printf ''"` returns 0. Because every call uses `curl -s` without `-f` or `%{http_code}`, HTTP status is discarded, and a `PUT` that succeeds (204, empty body) is indistinguishable from one that failed with an empty body. The HTML-error-page case *is* now caught by the new parse guard — good — so this is a narrow but real residual gap. `keycloak_get_optional`, added in this same PR, already has the right shape; routing the four SSO `POST`/`PUT` calls through a status-checking helper would close it cheaply and consistently. This matters more here than elsewhere in the script because the SSO block is deliberately best-effort: a masked failure produces a green deploy plus a login button that dead-ends at Keycloak, with no signal anywhere except init-container logs.

2. **`idpMetadataUrl` is doing two jobs: it supplies the metadata *and* marks the provider as chart-managed.** The disable branch keys off `.Values.enterpriseSSO.idpMetadataUrl` alone, so the README has to instruct operators to keep a metadata URL populated while turning SSO off — a value whose name promises "where to fetch metadata" now silently means "the chart owns this provider." That is the `one knob, one effect` rule in the repo guidelines, and the effect isn't predictable from the key name. It also has a concrete failure mode: if an operator clears the URL in the same rollout that flips the toggle off (the intuitive action, and what the KOTS UI nudges toward since the field is hidden by `when:`), neither branch renders and the provider stays enabled in Keycloak — reachable via a direct `kc_idp_hint=enterprise_sso` request, which is exactly what the disable path exists to prevent. An explicit ownership marker (e.g. a `managed-by` entry written into the provider's `config` at create time, or a separate values key) would make the intent legible and survive the URL being cleared. Related question I could not settle from the repo alone: does `ConfigOption "enterprise_sso_idp_metadata_url"` still return the stored value once `when:` hides the field? If KOTS returns empty for hidden items, the disable path never renders in Replicated installs — the primary target of this PR. Worth confirming before merge.

**[MINOR]**

3. `"Company SSO"` is defaulted in three places (`values.yaml`, the `| default` in `deployment.yaml`, and `replicated/config.yaml`). Since `values.yaml` already sets it, the template fallback only fires when someone explicitly sets it to empty. Dropping the `| default` would leave one authoritative chart default.

4. The two env entries in `deployment.yaml` re-implement the `.Values.env` precedence contract with `hasKey` guards instead of flowing through `openhands.env`. It renders correctly today — I confirmed the override test exercises the real `helm template` output — but it is a second copy of a cross-cutting rule that will drift from the helper. If keeping them out of `openhands.env.defaults` is deliberate (to avoid leaking SSO env into the app and litellm containers), a one-line comment saying so would help the next person.

**[RISK]** Moderate but well-contained. Defaults are inert, so existing installs and OSS values render unchanged, and the render guard prevents the half-configured combination. The residual exposure is operational rather than structural: a provisioning failure is best-effort by design and, given item 1, may not be visible at all; and item 2's disable path can silently no-op. Both surface as "SSO is broken/still on" rather than a failed rollout, which is the harder class to debug.

Happy to approve once items 1 and 2 are addressed or explicitly dismissed with reasoning.

charts/openhands/values.yaml
```diff
@@ -221,8 +221,13 @@ bitbucket:
   auth:
     existingSecret: bitbucket-app
 
+# Enterprise SSO (SAML). Enabling this advertises the login provider. When an
+# HTTPS metadata URL is also set, the keycloak-config init container manages the
+# enterprise_sso identity provider on every pod start.
 enterpriseSSO:
   enabled: false
+  displayName: "Company SSO"
```
aivong-openhands, line 229: Minor: `"Company SSO"` is now defaulted here, again via `| default "Company SSO"` in `deployment.yaml`, and a third time in `replicated/config.yaml`. Since this file always supplies a value, the template fallback only fires when someone explicitly sets it to empty string — at which point falling back is arguably the wrong behaviour anyway. Dropping the `| default` would leave one authoritative chart default.

charts/openhands/templates/keycloak-config-script.yaml
saurya: Fixed in b53f613. All five Enterprise SSO provider and mapper writes now use a status-aware helper that accepts empty 2xx responses but rejects empty 403 and other non-2xx responses. A real local HTTP regression verifies both 204 success and bodyless 403 failure. _This reply was generated by an AI agent (OpenHands) on behalf of the user._

charts/openhands/templates/keycloak-config-script.yaml
saurya: Fixed in b53f613. Auto-provisioned providers now store openhandsManaged=true in Keycloak config. Toggle-off reconciliation runs independently of idpMetadataUrl and disables only a provider carrying that marker, so clearing or hiding the KOTS field cannot skip disable and manual providers remain untouched. _This reply was generated by an AI agent (OpenHands) on behalf of the user._

charts/openhands/templates/deployment.yaml
saurya: Addressed in b53f613. The guards remain intentionally local to the Keycloak init container so provisioning-only values do not leak into app and LiteLLM containers; an inline comment now records that rationale and notes that openhands.env already emits operator overrides. _This reply was generated by an AI agent (OpenHands) on behalf of the user._

charts/openhands/values.yaml
saurya: Fixed in b53f613. The deployment template fallback was removed, leaving values.yaml as the authoritative chart default. Replicated retains its separate installer-field default so the Admin Console emits the intended chart value. _This reply was generated by an AI agent (OpenHands) on behalf of the user._

aivong-openhands (commented):
Reviewed at `7c02209` against the OpenHands-Cloud Helm chart guidelines. I rendered the chart and executed the generated script against a stub Keycloak rather than reading behaviour off the diff.

**What I verified locally**

- `helm unittest` on `enterprise_sso_test.yaml`, `validations_test.yaml`, `values_schema_test.yaml`: 31 passed.
- `pytest scripts/test_keycloak_realm_template.py`: 24 passed.
- Rendered `keycloak-config.sh` in automatic, manual, and disabled modes; all pass `sh -n` and `bash -n`.
- Ran the rendered enterprise-SSO block end to end against a stub Keycloak. It POSTs the provider and the mapper, and the stored provider carries `validateSignature=true`, `syncMode=IMPORT`, `openhandsManaged=true`, and `attribute.value=enterprise_sso:saml`. Re-ran the toggle-off block against that same stub state and it correctly flipped `enabled` to `false`.

The earlier round of feedback was genuinely addressed: transport and HTTP status failures now propagate, `openhandsManaged` replaced the overloaded `idpMetadataUrl` ownership inference, and the manual mapper contract is documented in both the README and the installer help text. The Keycloak provisioning logic is correct on the happy path.

Two findings below are worth attention before merge. Neither is a blocker for the provisioning logic itself; both are reachable states an operator can land in through the installer UI.

---

**Important — toggling Enterprise SSO off in the installer renders an un-deployable chart.**

`validations.yaml` fails on `idpMetadataUrl` being non-empty regardless of `enterpriseSSO.enabled`. But `replicated/config.yaml` marks `enterprise_sso_idp_metadata_url` with `when: enterprise_sso_enabled == "1"`, and hiding a KOTS config item does not clear the value the operator already typed. So the natural "turn Enterprise SSO back off" path leaves `enabled: false` with a stale URL still set.

That combination is fine while `keycloak.enabled=true` (the Replicated default), but it fails to render on any split-Keycloak deployment. Verified:

```
$ helm template t charts/openhands --set enabled=true \
    --set keycloak.enabled=false --set keycloak.provisionRealm=false \
    --set enterpriseSSO.enabled=false \
    --set-string enterpriseSSO.idpMetadataUrl=https://idp.example.com/saml/metadata
Error: enterpriseSSO.idpMetadataUrl requires chart-managed Keycloak realm provisioning...
```

The operator disabled the feature and the deploy now fails on a value they cannot see in the UI. Guarding on the URL only when the feature is on — `and .Values.enabled .Values.enterpriseSSO.enabled .Values.enterpriseSSO.idpMetadataUrl (not (include ...))` — matches the intent ("you asked for auto-provisioning but there is no realm to provision into") and makes the toggle-off path inert, which is what the disable branch already assumes.

---

**Important — the disable path is one-way; re-enabling never restores the provider.**

The toggle-off branch only sets `.enabled = false` on the managed provider, which is the right conservative choice. But the re-enable path is `and enabled idpMetadataUrl`, and per the finding above the metadata URL is the value most likely to be missing after a toggle-off/toggle-on cycle. In that state:

- `enabled: true`, no URL → `ENABLE_ENTERPRISE_SSO` is set and `enterprise_sso` is advertised in `OH_WEB_CLIENT_PROVIDERS_CONFIGURED`,
- but neither branch of the script renders, so the disabled provider in Keycloak stays disabled.

I confirmed both halves: rendering with `enabled=true, idpMetadataUrl=""` emits the login-provider env vars but produces no provisioning and no reconciliation block, and my stub still held `enabled: false` after the disable run. The result is a login button that leads to a disabled IdP, with no message anywhere explaining why. Re-enabling a provider that already carries `openhandsManaged=true` is the symmetric counterpart to the disable branch and closes the loop.

---

The remaining notes are minor and I would not hold the PR for them.

charts/openhands/templates/keycloak-config-script.yaml
```diff
@@ -231,4 +299,104 @@ data:
...
+        name: "identity-provider",
+        identityProviderAlias: "enterprise_sso",
+        identityProviderMapper: "hardcoded-attribute-idp-mapper",
+        config: {
+          attribute: "identity_provider",
+          "attribute.value": "enterprise_sso:saml",
+          syncMode: "FORCE"
+        }
+      }' > /tmp/mapper-enterprise-sso.json || exit 1
+      keycloak_api_call "curl -s \"$KC_REALM/identity-provider/instances/enterprise_sso/mappers\" $AUTH" || exit 1
```
aivong-openhands, line 374: Minor — the mapper existence check reads a bare `RESPONSE` and treats every non-array shape as "no mapper", so it falls through to POST. The provider and mapper writes below correctly use the status-aware `keycloak_write`, but this read still goes through `keycloak_api_call`, which only inspects the body for an `error` key and never sees the HTTP status.

Exercised the helper directly with the rendered script's own filter:

```
200 empty body             -> POST -> create
200 whitespace body        -> POST -> create
object instead of array    -> POST -> create
empty array                -> POST -> create
array with the mapper      -> PUT (id=abc)
```

The first three are the interesting rows: a proxy returning an empty 200, or Keycloak answering with an object because the provider POST above silently didn't land, both read as "mapper absent" and re-POST. Keycloak permits duplicate mapper names on the same alias, so a flapping proxy accumulates mappers rather than converging — which undercuts the idempotency the block's own comment promises.

`keycloak_get_optional` already returns the status-checked body you want here. Swapping this call for it and rejecting a non-array `RESPONSE` before the `select` would make the read as safe as the writes. Note that `keycloak_get_optional` maps 404 to an empty `RESPONSE`, so the non-array guard needs to distinguish "absent" from "unparseable".

charts/openhands/templates/deployment.yaml
```diff
@@ -74,6 +74,17 @@ spec:
         image: '{{.Values.image.repository}}:{{.Values.image.tag | default .Chart.AppVersion }}'
         env:
         {{- include "openhands.env" . | nindent 8 }}
+        {{- if and .Values.enterpriseSSO.enabled .Values.enterpriseSSO.idpMetadataUrl }}
+        {{- /* Keep these init-only; openhands.env already emits .Values.env overrides. */}}
+        {{- if not (hasKey (.Values.env | default dict) "ENTERPRISE_SSO_DISPLAY_NAME") }}
```
aivong-openhands, line 79: Minor, and I'm fine with the current answer — just flagging the consequence. These `hasKey` guards give `.Values.env` precedence for the init container, which is correct and matches `openhands.env`. The side effect is that `enterpriseSSO.displayName` is only ever emitted when `idpMetadataUrl` is also set.

```
$ helm template ... --set enterpriseSSO.enabled=true \
    --set-string enterpriseSSO.displayName="Acme SSO" \
    --set-string enterpriseSSO.idpMetadataUrl=""
# "Acme SSO" appears 0 times in the rendered output
```

So in manual mode an operator can set `displayName` in values (or type it into the installer field, which `config.yaml` shows whenever the toggle is on, independent of the URL) and it is silently discarded — the name comes from whatever they typed into the Keycloak admin console instead. That's the correct runtime behaviour, since the chart isn't managing the provider, but nothing tells them. A sentence in the `values.yaml` comment saying `displayName` applies only when `idpMetadataUrl` is set would cover it.

jlav (approved):

PR #1122, opened 2026-08-20
feat: add org-management e2e suite with super-admin REST provisioning

tofarr:
## Description

Introduces a new e2e spec suite (`006-org-management.spec.ts`) that exercises the organization-management REST surface directly (outside the browser) using a super-admin API key. This lays the groundwork for org-management UI checks that will follow.

Flow (runs only for the `new-user` role, skipped via `runUser` at run time):

1. `POST /api/organizations` with `{ name: "e2e test org <timestamp>", contact_name: "Testy Tester", contact_email: "e2e_test@openhands.org" }` and capture the returned `org.id`.
2. `POST /api/organizations/provision-user` (with `X-Org-Id: <org.id>`) to provision the New User (by `KEYCLOAK_NEW_USER_EMAIL`) into the new org as a `member`. The endpoint is idempotent, so re-runs only re-link membership and return the existing API key.

Supporting changes:
- Adds the `SUPER_ADMIN_API_KEY` env var (must be an **unbound** superadmin key so the server resolves the target org per-request from `X-Org-Id`, since the superadmin is not a member of the orgs it creates).
- Adds `superAdminApiKey()` and `newUserEmail()` config helpers in `utils/config.ts`.
- Adds a REST client (`utils/org-management.ts`) wrapping `createOrg` and `provisionUser`.
- Documents the new role/variable in `README.md` and the `.env` placeholder.

## Screenshots

image

## Helm Chart Checklist

- [x] N/A — no Helm chart changes; this is e2e test harness only.

## Additional Notes

- The `SUPER_ADMIN_API_KEY` placeholder in `.env` is left empty; the real unbound superadmin key must be supplied before running the suite. The `.env` file is gitignored.
- The spec is listed under both `:returning` and `:new-user` Playwright projects (per the existing topology) and skips itself at runtime for the returning role. If preferred, I can instead add a `testIgnore` to the returning projects so it isn't listed at all for that role.
- UI checks for the org-management flow will be added in a follow-up once the REST provisioning is confirmed working.


hieptl (approved):
Thank you! 🙏

PR #1126, opened 2026-08-20
fix(e2e): PLTF-3461 honor AUTH_BASE_URL for Keycloak admin URL

### LINKED ISSUE #1125: e2e harness: keycloak-cleanup derives wrong auth host for app.-subdomain targets (honor AUTH_BASE_URL)
## Summary

The e2e harness derives the Keycloak admin URL by prefixing `auth.` to the **full** `BASE_URL` host. That works when the app is served at an environment apex, but produces the wrong host when the app is served under an `app.` subdomain, so `keycloak-cleanup` fails at the TLS handshake and the New-User path can't run.

Example pattern (hostnames generalised):

- App (`BASE_URL`): `https://app.<env-domain>`
- Derived Keycloak: `https://auth.app.<env-domain>` ❌
- Correct Keycloak: `https://auth.<env-domain>`

The derived host isn't on the environment's wildcard cert (`*.<env-domain>`), so cleanup fails:

```
[keycloak-cleanup] delete new user from keycloak
  TypeError: fetch failed
  [cause]: Error: Hostname/IP does not match certificate's altnames:
    Host: auth.app.<env-domain> is not in the cert's altnames: DNS:*.<env-domain>
```

## Root cause

`e2e_tests/utils/config.ts`:

```ts
function keycloakUrlFromBaseUrl(): string {
  const baseUrl = process.env.BASE_URL;
  if (!baseUrl) throw new Error("BASE_URL is required to derive the Keycloak URL.");
  const url = new URL(baseUrl);
  return `${url.protocol}//auth.${url.host}`;   // prefixes "auth." to the FULL host
}
```

`keycloakAdminConfig()` uses this value, and there is no way to override it. The assumption that the Keycloak host is `auth.<app-host>` only holds when the app is served at the environment apex. When the app is served at an `app.` subdomain, the correct auth host drops the `app.` label rather than prepending `auth.` to it.

## Scope / impact

- Affects **only** `keycloak-cleanup` (New-User reset). Login (`setup:returning`, `setup:new-user`) is unaffected because it navigates to `BASE_URL` and the app redirects to its own Keycloak — it never uses this derivation.
- Observed with a parameterised runner that runs the same suite against multiple targets: apex-style targets pass end-to-end (7/7); an `app.`-subdomain target passes login + home checks but fails `keycloak-cleanup` on the cert mismatch above, which blocks the New-User path.

## Proposed fix

Honor an explicit auth URL when provided, falling back to the current derivation. A caller can resolve the target's real auth host (e.g. from the app's published web-client config) and pass it into the container as `AUTH_BASE_URL`; today the harness ignores it.

```ts
function keycloakUrlFromBaseUrl(): string {
  const explicit = process.env.AUTH_BASE_URL;
  if (explicit) return explicit.replace(/\/$/, "");   // honor the resolved auth host
  const baseUrl = process.env.BASE_URL;
  if (!baseUrl) throw new Error("BASE_URL is required to derive the Keycloak URL.");
  const url = new URL(baseUrl);
  return `${url.protocol}//auth.${url.host}`;          // fallback: legacy derivation
}
```

Notes:

- **Back-compatible.** For apex-style targets, `AUTH_BASE_URL == auth.<host>`, identical to the derived value, so existing behaviour is unchanged. It only changes the `app.`-subdomain case, where the two differ.
- Consider documenting `AUTH_BASE_URL` in the config header alongside `BASE_URL`, and (optionally) validating it is `https`.
- Alternative name if preferred: `KEYCLOAK_URL`. `AUTH_BASE_URL` is suggested because callers may already set it.

## Acceptance criteria

- With `AUTH_BASE_URL` set to the correct auth host, `keycloak-cleanup` authenticates and deletes the New User on an `app.`-subdomain target, and the full New-User path runs.
- Apex-style targets remain green with no change (no `AUTH_BASE_URL` required, or set equal to the derived host).

## Once merged

The runner bumps its pinned test revision to the fixed commit; the affected run should then go fully green.

---
_This issue was created by an AI agent (OpenHands) on behalf of a maintainer._

aivong-openhands:
## Why

`keycloak-cleanup` prepends `auth.` to the full `BASE_URL` host, so targets served at `app.<env-domain>` resolve to `auth.app.<env-domain>`, which is outside the environment certificate and blocks the New-User path. The harness now prefers a trimmed, validated HTTP(S) `AUTH_BASE_URL`; leaving it unset or blank preserves the existing derivation for apex-style targets.

Fixes #1125

## Validation

- **Unit tests** — `npm run test:unit` passed all six cases covering unset and blank fallback, URL normalization, supported protocols, malformed input, and unsupported protocols.

### DIFF
--- a/e2e_tests/package.json
+++ b/e2e_tests/package.json
@@ -6,6 +6,7 @@
   "type": "module",
   "scripts": {
     "test": "playwright test",
+    "test:unit": "playwright test --config unit-tests/playwright.config.ts",
     "test:chromium": "playwright test --project=chromium:returning --project=chromium:new-user",
     "test:returning": "playwright test --project=chromium:returning --project=firefox:returning --project=webkit:returning",
     "test:new-user": "playwright test --project=chromium:new-user --project=firefox:new-user --project=webkit:new-user",
--- a/e2e_tests/unit-tests/keycloak-config.spec.ts
+++ b/e2e_tests/unit-tests/keycloak-config.spec.ts
@@ -1,0 +1,90 @@
+import { expect, test } from "@playwright/test";
+
+import { keycloakAdminConfig } from "../utils/config";
+
+const configEnvKeys = [
+  "AUTH_BASE_URL",
+  "BASE_URL",
+  "KEYCLOAK_ADMIN_USERNAME",
+  "KEYCLOAK_ADMIN_PASSWORD",
+  "KEYCLOAK_NEW_USER_EMAIL",
+] as const;
+
+function withConfigEnv(
+  overrides: Partial<Record<(typeof configEnvKeys)[number], string>>,
+  run: () => void,
+): void {
+  const original = Object.fromEntries(
+    configEnvKeys.map((key) => [key, process.env[key]]),
+  );
+
+  try {
+    configEnvKeys.forEach((key) => delete process.env[key]);
+    Object.assign(process.env, {
+      KEYCLOAK_ADMIN_USERNAME: "admin",
+      KEYCLOAK_ADMIN_PASSWORD: "password",
+      KEYCLOAK_NEW_USER_EMAIL: "new-user@example.com",
+      ...overrides,
+    });
+    run();
+  } finally {
+    configEnvKeys.forEach((key) => {
+      const value = original[key];
+      if (value === undefined) {
+        delete process.env[key];
+      } else {
+        process.env[key] = value;
+      }
+    });
+  }
+}
+
+test("uses the derived Keycloak URL when AUTH_BASE_URL is empty", () => {
+  withConfigEnv(
+    {
+      AUTH_BASE_URL: "   ",
+      BASE_URL: "https://staging.all-hands.dev",
+    },
+    () => {
+      expect(keycloakAdminConfig().keycloakUrl).toBe(
+        "https://auth.staging.all-hands.dev",
+      );
+    },
+  );
+});
+
+test("uses the derived Keycloak URL when AUTH_BASE_URL is unset", () => {
+  withConfigEnv({ BASE_URL: "https://staging.all-hands.dev" }, () => {
+    expect(keycloakAdminConfig().keycloakUrl).toBe(
+      "https://auth.staging.all-hands.dev",
+    );
+  });
+});
+
+test("normalizes an explicit AUTH_BASE_URL", () => {
+  withConfigEnv({ AUTH_BASE_URL: "  https://auth.example.com/  " }, () => {
+    expect(keycloakAdminConfig().keycloakUrl).toBe("https://auth.example.com");
+  });
+});
+
+test("accepts an explicit HTTP AUTH_BASE_URL", () => {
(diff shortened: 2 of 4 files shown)


aivong-openhands (commented):
### Code review — `/codereview`

**Verdict: APPROVE-equivalent** ✅ _(posted as a COMMENT because the GitHub token belongs to the PR author, and GitHub disallows approving your own PR)_

Scope is a single Playwright e2e harness file (`e2e_tests/utils/config.ts`); no Helm charts are touched, so the chart-specific checks (storage/state, env-var de-duplication, `values.yaml` toggles, `Chart.yaml`) don't apply here. The Helm checklist is correctly marked N/A.

**What's good**
- Moves the auth host from a fragile derived convention (`auth.` + host) toward an **explicit, consumer-provided value** (`AUTH_BASE_URL`) — exactly the "explicit over implicit" direction the guidelines favor. The derivation stays as a fallback, so apex-style targets are unchanged and only the broken subdomain case (`app.<env>`) is fixed.
- Well documented: `AUTH_BASE_URL` is described both in the env-var header block and in the `keycloakUrlFromBaseUrl()` docstring, including *why* the old derivation is wrong for subdomain targets.
- Back-compatible, as the description claims: for apex targets the explicit value equals the derived one.

**Minor nitpicks (non-blocking)**
1. `if (explicit)` means an empty `AUTH_BASE_URL=""` silently falls through to the derivation. That's a reasonable "treat empty as unset" choice, but worth being intentional about.
2. The explicit branch only trims a trailing slash, whereas the derived branch parses via `new URL()`. A malformed `AUTH_BASE_URL` would surface later at request time rather than failing fast here — wrapping it in `new URL(...)` (or validating the protocol) would give a clearer config-time error. Optional.

Neither rises to "important," so this is good to merge.

---
_This review was generated by an AI agent (OpenHands) on behalf of the user, using the repo's `code-review` skill._

aivong-openhands (commented):
### Code review — `/codereview`

**Verdict: APPROVE-equivalent** ✅ _(posted as a COMMENT because the GitHub token belongs to the PR author, and GitHub disallows approving your own PR)_

Reviewed the current head `90b5b74`. No important issues found.

The follow-up resolves both points from the previous review:

- Whitespace-only `AUTH_BASE_URL` is intentionally treated as unset, preserving the legacy `BASE_URL` derivation.
- Explicit values are trimmed, parsed with `new URL()`, restricted to HTTP(S), normalized, and rejected early with a clear configuration error when invalid.
- Focused tests cover unset/blank fallback, normalization, HTTP and HTTPS acceptance, malformed input, and unsupported protocols.
- The unit-test Playwright config is isolated under `unit-tests/`, and the main E2E config remains scoped to `./tests`.
- All reported PR checks are passing.

This remains backwards-compatible and follows the repo guideline to prefer explicit consumer-provided configuration over fragile hostname conventions. Good to merge.

---
_This review was generated by Codex on behalf of the user, using the repository's `code-review` skill._

saurya (approved):

PR #1132, opened 2026-08-20
ci: fix unparseable expression in deploy-replicated, lint workflows in CI

dylan-openhands:
## Human Description

A few of the replicated deploy actions had parsing errors that were valid YAML but invalid Github Action format. Did a pass of all action files in the repo to patch a few warnings and fix the mistakes i introduced.

Also added a linter step using `actionlint` for catching this in the future.

- `deploy-replicated.yml` — `timeout-minutes: 30`, a literal. `scripts/replicated_deploy.sh:11-12` enforces its own deadline from `TIMEOUT_MINUTES`, so the job cap is only a backstop; both callers use the default `"25"`, so this preserves current behavior. A caller passing `timeout_minutes: 60` would now be killed at 30 — there's no way to derive it without arithmetic, so the constraint is a comment.
- `lint-workflows.yml` — new. Runs actionlint on any PR touching `.github/workflows/**`, pinned to v1.7.12. This class of bug was statically detectable the whole time.
- `deploy-gate.yml` — `$(echo $PATHS)` relied on word-splitting to flatten newlines to spaces; now `${PATHS//$'\n'/ }`.
- `scan-docker-images.yml` — quote `$GITHUB_OUTPUT`.

The last two are pre-existing shellcheck findings, fixed so the new lint job starts from green.

Kept it a standalone workflow rather than a job in `pr-checks.yml`, which would have meant widening that workflow's path filter and running the chart/template suite on every workflow-only PR.

## Helm Chart Checklist

N/A — no charts touched.

## Additional Notes

`actionlint` 1.7.12 is clean across the whole repo (exit 0), not just the changed files. Verified `${PATHS//$'\n'/ }` produces output identical to `$(echo $PATHS)` for a multi-line value.

The check won't block merges until its context is added to the `main` ruleset — same gap `deploy-gate.yml` has today.

### DIFF
--- a/.github/workflows/lint-workflows.yml
+++ b/.github/workflows/lint-workflows.yml
@@ -1,0 +1,22 @@
+name: Lint Workflows
+
+# Linter for github actions specifically. Surfaces rules YAML wouldn't notice
+# like attempting arithmetic in action files
+on:
+  pull_request:
+    paths:
+      - '.github/workflows/**'
+  workflow_dispatch:
+
+jobs:
+  actionlint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Run actionlint
+        env:
+          ACTIONLINT_VERSION: '1.7.12'
+        run: |
+          bash <(curl -sSf "https://raw.githubusercontent.com/rhysd/actionlint/v${ACTIONLINT_VERSION}/scripts/download-actionlint.bash") "$ACTIONLINT_VERSION"
+          ./actionlint -color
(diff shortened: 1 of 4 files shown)


.github/workflows/lint-workflows.yml
```diff
@@ -0,0 +1,22 @@
...
+# like attempting arithmetic in action files
+on:
+  pull_request:
+    paths:
+      - '.github/workflows/**'
+  workflow_dispatch:
+
+jobs:
+  actionlint:
+    runs-on: ubuntu-latest
```
aivong-openhands, line 13: nit - I recommend a fixed ubuntu image

aivong-openhands (approved):

PR #1139, opened 2026-08-21
feat(replicated): PLTF-3461 configure duplicate email checks

aivong-openhands:
## Why

Replicated installations need an explicit way to disable duplicate base-email protection when plus-address identities are intentional. This adds a typed `duplicateEmailCheck` chart value and a Replicated **Prevent Duplicate Base Emails** checkbox; it defaults to `true` to preserve existing behavior, while an explicit `env.DUPLICATE_EMAIL_CHECK` override still takes precedence without rendering duplicate entries. Disabling the option permits identities whose plus-address variants normalize to the same base email and should only be used where that behavior is intentional.

## Validation

- **Chart rendering** — Verified the default and disabled values render `DUPLICATE_EMAIL_CHECK` as `true` and `false`, respectively.
- **Override precedence** — Verified an explicit environment override wins and renders exactly one `DUPLICATE_EMAIL_CHECK` entry.
- **Replicated contract** — Verified the checkbox defaults enabled and maps to the chart value rather than directly to the application environment.

### DIFF
--- a/charts/openhands/templates/_env.yaml
+++ b/charts/openhands/templates/_env.yaml
@@ -34,6 +34,8 @@
 - name: OH_DAILY_CONVERSATION_LIMIT
   value: {{ .Values.dailyConversationLimit | quote }}
 {{- end }}
+- name: DUPLICATE_EMAIL_CHECK
+  value: {{ .Values.duplicateEmailCheck | quote }}
 - name: OH_RESOLVER_LABEL
   value: {{ get $integrations "resolverLabel" | default "openhands" | quote }}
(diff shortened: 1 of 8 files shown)


aivong-openhands (commented):
Reviewed using the repository code-review guidelines. No important-level findings: the typed chart value, schema validation, Replicated mapping, and env-override de-duplication follow established patterns, and all current CI checks pass. The resulting disposition is APPROVE, but GitHub does not allow the PR author account to approve its own PR.

dylan-openhands (approved):

PR #995, opened 2026-07-30
feat(previews): add Enterprise Replicated preview workflow

saurya:
## Summary

Adds an opt-in workflow for publishing OpenHands Enterprise PR images as Replicated releases and installing them on Embedded Cluster preview VMs. GCP is the preferred provider; AWS remains available as a legacy test path.

This revision incorporates the review feedback by removing per-preview GCP certificate issuance, using flat testing-domain hosts covered by one shared wildcard certificate, enabling GitHub authentication, installing k9s, removing license expiry drift, adding age-based cleanup, and merging current `main`.

## GCP preview architecture

- Project: `staging-092324`
- Region / zone: `us-central1` / `us-central1-a`
- Network / subnetwork: `staging-core-app`
- DNS zone: `staging-all-hands-testing-dot-dev`
- Shared suffix: `staging.all-hands-testing.dev`
- Application: `app-pr-<PR>.staging.all-hands-testing.dev`
- Authentication: `auth-pr-<PR>.staging.all-hands-testing.dev`
- Admin console: `admin-pr-<PR>.staging.all-hands-testing.dev:30000`
- Runtime API: `runtime-api-pr-<PR>.staging.all-hands-testing.dev`
- Sandboxes: `runtime-pr-<PR>.staging.all-hands-testing.dev/<runtime-id>`

The PR number is kept in the same DNS label as the service name. One certificate for `*.staging.all-hands-testing.dev` can therefore be stored in GitHub secrets and shared across all preview VMs. GCP Terraform no longer includes an ACME provider, account, or certificate resource.

## Authentication

Real VM previews enable GitHub authentication through a shared GitHub App. The workflow renders the OAuth client, App metadata, webhook secret, and private key into Replicated ConfigValues. The App must include each active preview's OAuth callback URL.

## Lifecycle

- Dry run remains the default.
- Real GCP previews require remote Terraform state so later workflow runs can destroy the full resource set.
- Replicated development customers no longer expire, avoiding license drift when a preview is extended.
- `.github/workflows/cleanup-enterprise-replicated-previews.yml` runs daily and dispatches full destroy jobs for GCP preview VMs older than seven days.
- Manual cleanup runs default to report-only mode and support a custom age.
- k9s is installed by the GCP VM startup script.

## Required GCP secrets

- `REPLICATED_API_TOKEN`
- `GCP_WORKLOAD_IDENTITY_PROVIDER`
- `GCP_SERVICE_ACCOUNT`
[...]

PR #1149, opened 2026-08-24
ci: block the openhands release PR on a red unstable E2E [ref PLTF-3540]

lilagrc:
Promotion from unstable to beta happens by merging the openhands release-please PR, and nothing checks whether unstable is actually green first.

Adds `release-gate.yml`, a check scoped to that one PR.

### Reading the result from Argo, not from GitHub

Most unstable runs will be scheduled by a CronWorkflow in saas-deploy ([#995](https://github.com/OpenHands/saas-deploy/pull/995)), and **a scheduled run has no GitHub run at all** — so a gate inspecting job names would be blind to the majority of runs and would gate on the rarer deploy-triggered subset. One `list` call over `openhands-e2e-unstable-*` covers both trigger paths, because scheduled (`-scheduled-<epoch>`) and deploy-bound (`-gh-<run id>-<attempt>`) submits share the prefix. It also survives a job being renamed, which name-matching does not.

### Blocks on "no run", not just on red

Argo has no archive database here, so a finished Workflow lives only as long as its 24h TTL. The staleness window is **20h**, deliberately inside that: a wider window would block on runs that merely aged out. A missing signal is treated as a failure — a badge renders its last conclusion forever, so "the suite stopped running" has to be a blocking state or it's invisible.

### Override is attributable and single-use

`override-e2e-gate` is honoured only when its newest `labeled` event **postdates the failing signal** and its applier has write access. Re-labelling is required for each new red signal, and the gate posts a comment recording who overrode what.

### Scoping

The release PR is identified by its same-repository head branch (`release-please--branches--main--components--openhands`), not by a label anyone can remove. Non-matching PRs skip the job, which satisfies a required check without a `paths:` filter.

A 15-minute `schedule` and a `workflow_run` trigger re-run the gate for an open release PR, so a scheduled recovery clears it without anyone pushing. The refresh job exits immediately when no release PR is open.

### Depends on

- saas-deploy [#996](https://github.com/OpenHands/saas-deploy/pull/996) for `ARGO_WORKFLOWS_GATE_TOKEN`. Set it **repo-level, not scoped to `e2e-replicated`** — the gate job runs on `refs/pull/N/merge`, which that environment does not admit, so an environment-scoped secret reads as empty and blocks every release PR on "Argo unreachable". Safe at repo level: the token is read-only (`list`, `get` on workflows in one namespace).
- saas-deploy [#995](https://github.com/OpenHands/saas-deploy/pull/995) for the schedule, so a quiet week still produces a signal.

**Not enforcing on merge.** Add the `release-gate` context to the main ruleset only after watching it against real release PRs.

Rebased onto main post-#1228 and squashed: the earlier six commits included a wait-and-page implementation that a later commit removed, and `e2e-replicated.yml` now takes main's version unchanged. `scripts/test_deploy_replicated_e2e_trigger.py` 22 pass. `actionlint` clean.

### DIFF
--- a/.github/workflows/release-gate.yml
+++ b/.github/workflows/release-gate.yml
@@ -1,0 +1,225 @@
+name: Release gate (unstable E2E)
+
+# Blocks the openhands release-please PR when the latest unstable E2E is not green.
+# Enforce by adding the `release-gate` context to the main ruleset.
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize, labeled, unlabeled]
+  workflow_run:
+    workflows: ["Release Replicated to Unstable"]
+    types: [completed]
+  # Scheduled Argo runs fire no GitHub event, so nothing else would re-evaluate
+  # a blocked release PR when the signal recovers.
+  schedule:
+    - cron: "*/15 * * * *"
+
+permissions: {}
+
+env:
+  RELEASE_PR_BRANCH: "release-please--branches--main--components--openhands"
+  OVERRIDE_LABEL: "override-e2e-gate"
+  ARGO_NAMESPACE: "openhands-e2e"
+  ARGO_BASE: "https://workflows.dev.all-hands.dev"
+  E2E_INSTANCE: "unstable"
+  # Must stay under Argo's 24h ttlStrategy, or aged-out runs read as "no run".
+  E2E_MAX_AGE_HOURS: "20"
+
+jobs:
+  release-gate:
+    name: release-gate
+    if: >-
+      github.event_name == 'pull_request' &&
+      github.event.pull_request.base.ref == 'main' &&
+      github.event.pull_request.head.ref == 'release-please--branches--main--components--openhands' &&
+      github.event.pull_request.head.repo.full_name == github.repository
+    runs-on: ubuntu-latest
+    # No `environment:` - e2e-replicated rejects refs/pull/N/merge, so the gate
+    # token is repo-level. Safe: it is read-only, no submit, no Secrets.
+    permissions:
+      actions: read
+      contents: read
+      pull-requests: write
+    steps:
+      - name: Evaluate release gate
+        env:
+          GH_TOKEN: ${{ github.token }}
+          REPO: ${{ github.repository }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+          ARGO_GATE_TOKEN: ${{ secrets.ARGO_WORKFLOWS_GATE_TOKEN }}
+        run: |
+          set -euo pipefail
+
+          {
+            echo "### Release gate (unstable E2E)"
+            echo ""
+          } >> "$GITHUB_STEP_SUMMARY"
+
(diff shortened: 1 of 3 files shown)


mamoodi (approved):

jlav (approved):

.github/workflows/release-gate.yml
```diff
@@ -0,0 +1,225 @@
...
+      actions: write
+      contents: read
+      pull-requests: read
+    steps:
+      - name: Rerun the open release PR gate
+        env:
+          GH_TOKEN: ${{ github.token }}
+          REPO: ${{ github.repository }}
+        run: |
+          set -euo pipefail
```
aivong-openhands, line 188: Could these be replaced with `bash: shell`?

.github/workflows/release-gate.yml
```diff
@@ -0,0 +1,225 @@
...
+
+jobs:
+  release-gate:
+    name: release-gate
+    if: >-
+      github.event_name == 'pull_request' &&
+      github.event.pull_request.base.ref == 'main' &&
+      github.event.pull_request.head.ref == 'release-please--branches--main--components--openhands' &&
+      github.event.pull_request.head.repo.full_name == github.repository
+    runs-on: ubuntu-latest
```
aivong-openhands, line 36:
```suggestion
    runs-on: ubuntu-24.04
```

Can use a fixed image

.github/workflows/release-gate.yml
```diff
@@ -0,0 +1,225 @@
...
+              "_Posted automatically by an AI-assisted workflow on behalf of the OpenHands-Cloud maintainers._")
+            gh pr comment "$PR_NUMBER" --repo "$REPO" --body "$body"
+          fi
+
+          echo "Override honored for the signal completed at ${run_completed}."
+
+  refresh-release-gate:
+    name: Refresh release gate
+    if: github.event_name == 'workflow_run' || github.event_name == 'schedule'
+    runs-on: ubuntu-latest
```
aivong-openhands, line 177:
```suggestion
    runs-on: ubuntu-24.04
```

aivong-openhands (approved):
Just minor recommendations, LGTM!

.github/workflows/release-gate.yml
dylan-openhands: Good call. The reason that it _wasn't_ used is because i've found from experience that there's a bunch of silent failure possibilities with default `shell` invocations. I added this instead at the top as a default:
```
defaults:
  run:
    # Some reasonable default args to any bash invocation usage
    shell: bash --noprofile --norc -euo pipefail {0}
```

Using `shell: bash` alone doesn't usually give the `-euo pipefail` pieces. The `pipefail` part, in particular, is the biggest source of my own headaches in the past. Using a piped command where the first part of the pipe fails but the end succeeds and thus the whole command is considered a success.

Anyways thanks for the catch and suggestion

aivong-openhands: TIL this is a neat find, thanks!

PR #1156, opened 2026-08-25
fix: Disable agent-canvas telemetry by default for self-hosted

lilagrc:
## Summary

Defaults the `agent-canvas` subchart to disable PostHog product telemetry in the pre-built canvas frontend served under `/canvas`. This is the deployment side of Linear **OHE-3135** (Disable PostHog telemetry by default for self-hosted installs).

OHE is sold as an on-prem offering with no telemetry, but the shared published `agent-canvas` image ships with telemetry **on by default**. Once OHE customers started viewing the canvas at `/canvas`, PostHog events began flowing. This chart targets self-hosted deployments, so telemetry should be **off by default**.

## How it works — env var, not a CLI flag (safe rollout by design)

When `staticServer.disableTelemetry: true` (the new default), the chart sets **`AGENT_CANVAS_DISABLE_TELEMETRY=1`** on the container. `static-server.mjs` reads it and injects `window.__AGENT_CANVAS_DO_NOT_TRACK__`, which the precompiled bundle honors to disable all telemetry (including the anonymous install event) — no `VITE_DO_NOT_TRACK` baked into the image required.

**Why an env var and not a `--disable-telemetry` CLI flag:** `static-server.mjs` parses CLI args with a strict allowlist and throws on any unknown flag, so a chart that passed a *new* flag to an *older* image would **crash-loop the container**. An unknown **env var**, by contrast, is simply ignored:

- A **new** image reads `AGENT_CANVAS_DISABLE_TELEMETRY` and disables telemetry.
- An **older** image that predates the option ignores it and starts normally — telemetry just stays at the image's current default. No crash, no regression.

This removes the image-tag/chart rollout-ordering hazard entirely. **The chart and the image can roll out in any order**, and this PR does **not** need to bump `image.tag` to be safe. The worst case on version skew is "telemetry stays on until the image catches up" — never "canvas pod down."

## Changes

- **`charts/openhands/charts/agent-canvas/templates/deployment.yaml`**: when `staticServer.disableTelemetry` is true, adds `AGENT_CANVAS_DISABLE_TELEMETRY=1` to the container `env`. The env block now renders when telemetry is disabled OR `.Values.env` is set, and composes both. The previous `--disable-telemetry` CLI arg is removed.
- **`charts/openhands/charts/agent-canvas/values.yaml`**: `staticServer.disableTelemetry: true` default, documented with the env-var mechanism and the forward-compatibility rationale.
- **`charts/openhands/values.yaml`**: surfaces the same default in the umbrella `agent-canvas` block for discoverability.
- **`charts/openhands/charts/agent-canvas/README.md`**: documents `staticServer.disableTelemetry` (default, env-var mechanism, image requirement, SaaS override) next to the existing `lockToCloud` section.

The runtime plumbing (env var → `window.__AGENT_CANVAS_DO_NOT_TRACK__`) is added in the companion PR **OpenHands/OpenHands#16908** (`disable-canvas-telemetry-self-hosted`).

## SaaS override

A SaaS deployment that wants telemetry enabled overrides `staticServer.disableTelemetry: false` (or does not use this subchart). With `false`, neither the env var nor any flag is rendered.

## Validation

`helm lint charts/openhands/charts/agent-canvas` passes. `helm template` across the relevant value combinations:

| Values | Result |
|---|---|
| default (`disableTelemetry: true`) | `env:` contains `AGENT_CANVAS_DISABLE_TELEMETRY=1`; no `--disable-telemetry` arg |
| `--set staticServer.disableTelemetry=false` | no telemetry env var, no flag |
| `true` + `--set env.FOO=bar` | both `AGENT_CANVAS_DISABLE_TELEMETRY=1` and `FOO=bar` render |
| `false` + `--set env.FOO=bar` | only `FOO=bar` renders |

```console
$ helm lint charts/openhands/charts/agent-canvas
1 chart(s) linted, 0 chart(s) failed

$ helm template ac charts/openhands/charts/agent-canvas | grep -A1 AGENT_CANVAS_DISABLE_TELEMETRY
            - name: AGENT_CANVAS_DISABLE_TELEMETRY
              value: "1"

$ helm template ac charts/openhands/charts/agent-canvas \
    --set staticServer.disableTelemetry=false | grep -c AGENT_CANVAS_DISABLE_TELEMETRY
0
```

---

_This PR was created by an AI agent (OpenHands) on behalf of the team._

@lilagrc can click here to [continue refining the PR](https://app.all-hands.dev/conversations/b072b08a-3120-4a55-a200-0519075ad2c8)


hieptl (approved):
Thank you! 🙏

PR #1167, opened 2026-08-26
ci: check the agent-server tag against the enterprise SDK pin

jlav:
## Description

The enterprise server imports `openhands-agent-server` (published from `software-agent-sdk`) as a client, and talks to the agent-server image over HTTP in every sandbox. Client and server are the same codebase, so a version split between them is a protocol split.

The charts pin both halves independently. `image.tag` names the enterprise-server release, `global.agentServerImage.tag` names the sandbox image, and nothing links the two. Bumping either one alone ships that split silently.

So this adds a blocking check on release-please PRs. It reads the enterprise-server pin, resolves it on the `OpenHands/enterprise` remote, reads the `openhands-agent-server` version that release pinned in its `pyproject.toml`, and fails when the chart's agent-server tag doesn't match.

One thing to know before merging: `main` is already out of sync. It pins enterprise-server `1.56.0`, which was built against SDK `1.43.1`, but the charts run `1.41.0-python`. This check fails on the next release PR until those tags move.

## Helm Chart Checklist

## Additional Notes

The `pull_request` trigger is deliberately unfiltered by path. An `openhands-secrets` or `crd-check` release PR touches nothing under `charts/openhands`, but it still publishes the release line this invariant covers. Keeping the workflow always-triggered also means it can be made a required check without leaving every other PR pending forever, since the job's `if` gate reports as skipped rather than never running at all.

The check covers all three agent-server pins, so it picks up `charts/image-loader/values.yaml` and the runtime-api subchart too. Those are separate files with no shared default, so they can drift from each other as well as from the SDK.

`OpenHands/enterprise` is public, so the job's own read-only `GITHUB_TOKEN` is enough to read its tags and `pyproject.toml`. It's passed only to lift the API rate limit, and there's no app token to provision.

Tests live in `scripts/test_check_agent_server_sync.py` and stub the remote, so they don't touch the network. One of them resolves the pin table against the real chart files, so a `values.yaml` restructure that moves a pin fails loudly instead of quietly turning the check into a no-op.

### DIFF
--- a/.github/workflows/check-agent-server-sync.yml
+++ b/.github/workflows/check-agent-server-sync.yml
@@ -1,0 +1,69 @@
+name: Check agent-server sync
+
+# Blocking check on release-please PRs: the agent-server image tag the charts
+# pin must match the software-agent-sdk version the pinned enterprise-server
+# release was built against.
+#
+# The enterprise server imports openhands-agent-server (published from
+# OpenHands/software-agent-sdk) as a client and talks to the agent-server image
+# over HTTP in every sandbox. Client and server are the same codebase, so a
+# version split is a protocol split. Nothing links the two pins, so a bump to
+# either one alone silently ships that split -- this catches it at the release
+# PR, the last point before the chart is published.
+#
+# See scripts/check_agent_server_sync.py for the pins it reads.
+
+on:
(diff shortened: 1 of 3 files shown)


aivong-openhands (approved):
This is going to be super helpful, thank you!

PR #1176, opened 2026-08-31
test: OHE-3120 : add 005 canvas conversations spec

tofarr:
## Description

Adds `005-conversations-canvas.spec.ts`, which mirrors `003-conversations-legacy.spec.ts` against the new Agent Canvas UI (the OpenHands/OpenHands frontend bundle hosted at `/canvas`). The canvas frontend shares a few test IDs with the legacy UI but differs in enough selectors that a parallel set of page objects is required rather than reusing the legacy ones.

**New page objects:**
- `CanvasHomePage` — home screen at `/canvas`; starts conversations by typing into `chat-input` and submitting (no dedicated launch button), selects repositories via `git-repo-dropdown` + `repo-launch-button`, and navigates recent conversations via `conversation-panel` → `conversation-card`.
- `CanvasConversationPage` — conversation interface; surfaces readiness/completion via `chat-status-indicator` text ("Ready" / "Done") instead of the legacy long-form strings ("Waiting for task" / "Agent has finished the task"), renders messages as `assistant-message`/`user-message`, and sends via `submit-button`.

**Tests ported from 003 (serial, per-role):**
1. Start a conversation via chat input and reverse a string
2. Navigate to a running conversation (click first `conversation-card`)
3. Tavily search (asserts `Micheál Martin` with accent-tolerant regex)

**Omitted from 003:** The VSCode integration test. The canvas UI does not expose a `conversation-tab-vscode` test ID; VSCode is reached via `drawer-vscode-link` whose iframe structure differs from the legacy editor tab. That flow should be validated against a live canvas DOM via `npx playwright codegen` before being ported, to avoid asserting on unconfirmed selectors.

The spec is automatically picked up by the existing Playwright project topology (`.*\.spec\.ts$`) for all browser/role combinations — no config changes needed.

## Helm Chart Checklist

_(No Helm chart changes — this PR only adds e2e test files.)_

## Additional Notes

- **Auth validation needed:** The canvas static server runs in `--auth-required` mode. Whether the storage-state produced by the existing `setup:returning`/`setup:new-user` projects (which authenticate against the Cloud app's Keycloak) satisfies the canvas auth gate depends on the deployment (shared Keycloak realm `allhands` + same parent domain → likely yes; otherwise a "Manage Backends" step may be required). This should be validated against a live `/canvas` target before the suite is run in CI.
- **Selectors derived from source, not a live DOM.** The canvas selectors were read from the OpenHands/OpenHands frontend source. The four flows most likely to need adjustment after a live codegen pass are: home chat launcher submit, repo+branch launch, the `conversation-card` click, and (if/when ported) the VSCode drawer iframe.


hieptl (approved):
Thank you! 🙏

PR #1178, opened 2026-08-31
feat(runtime-api): sync split cleanup CronJob from runtime-api#729

tofarr:
## Problem

PR [OpenHands/runtime-api#729](https://github.com/OpenHands/runtime-api/pull/729) (merged in release [v0.9.0](https://github.com/OpenHands/runtime-api/releases/tag/v0.9.0)) split the single cleanup CronJob into four per-phase CronJobs (`reaper`, `k8s-garbage`, `snapshotter`, `retention`) for independent failure isolation.

The runtime-api chart was published as OCI artifact `0.2.12`, but the **vendored copy** of the runtime-api subchart in `charts/openhands/charts/runtime-api/` was never updated — PR #1177 only bumped the `image.tag` (`0.8.2` → `0.9.0`) in `values.yaml`. As a result, no published `openhands` umbrella version (including `0.57.0`) contains the split. Deployed clusters still see the old single `…-cleanup` CronJob.

## Changes

This PR syncs the three changed files from the runtime-api source into the vendored subchart:

### `cleanup-cronjob.yaml`
Rewritten from a single CronJob to a `range`-driven template rendering four CronJobs, each with its own `schedule`, `activeDeadlineSeconds`, and `python cleanup.py <phase>` command.

**Preserves the vendored chart's existing features** that are not in the runtime-api source chart:
- `runtime-api.affinity` include
- `ingressBase.enabled` volumes/volumeMounts

The archive `emptyDir` volume/mount is scoped to the reaper phase only (the only phase that archives workspaces).

### `_helpers.tpl`
Adds the `runtime-api.cronjobName` helper (truncates the fullname to stay under K8s' 63-char name limit with the `-cleanup-<phase>` suffix).

### `values.yaml`
- Adds `cleanup.jobs.{reaper,k8s-garbage,snapshotter,retention}` with per-job `enabled`/`schedule`/`activeDeadlineSeconds`.
- Adds `runtimeFileArchive.tmpDir`/`tmpSizeLimit` (referenced by the new template, with `default` fallbacks for backwards compat).
- Preserves existing defaults (`dead_seconds: 5184000`, etc.).

## Schedules

| Job | Phase(s) | Schedule | Deadline |
|---|---|---|---|
| `cleanup-reaper` | pause/stop/remove idle runtimes | `*/5 * * * *` | 900s |
| `cleanup-k8s-garbage` | reap pods then stuck PVCs (serial) | `2-59/5 * * * *` | 600s |
| `cleanup-snapshotter` | snapshot + delete PVCs | `*/10 * * * *` | 1200s |
| `cleanup-retention` | delete dead runtimes + expired fuse workspaces | `0 */1 * * *` | 1800s |

The two K8s-list-heavy jobs (reaper `*/5`, k8s-garbage `2-59/5`) are staggered to avoid simultaneous full namespace pod/PVC lists against the apiserver.

## Validation

- `helm lint` on the runtime-api subchart: **passed** (0 errors)
- `helm template` renders all 4 CronJobs with correct names, schedules, deadlines, and commands
- Archive `emptyDir` volume/mount confirmed scoped to reaper phase only
- `ingressBase` volumes confirmed present in all phases when enabled

## After merge

The next `openhands` umbrella release (e.g. `0.58.0`) will bundle this updated subchart. `saas-deploy` must then bump its `Chart.yaml` pin to that version and re-deploy.

---

### DIFF
--- a/charts/openhands/charts/runtime-api/templates/_helpers.tpl
+++ b/charts/openhands/charts/runtime-api/templates/_helpers.tpl
@@ -10,6 +10,13 @@ Create a default fully qualified app name.
 We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
 If release name contains chart name it will be used as a full name.
 */}}
+{{- define "runtime-api.cronjobName" -}}
+{{- $suffix := printf "-%s" .suffix -}}
+{{- $maxLen := int (sub 52 (len $suffix)) -}}
+{{- $fullname := include "runtime-api.fullname" .root -}}
+{{- printf "%s%s" ($fullname | trunc $maxLen | trimSuffix "-") $suffix -}}
+{{- end -}}
+
 {{- define "runtime-api.fullname" -}}
 {{- if .Values.fullnameOverride }}
 {{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" }}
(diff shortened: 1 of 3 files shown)


hieptl (approved):
Thank you! 🙏

PR #1181, opened 2026-08-31
fix: rename warm-runtime default config to v1_current to match the app default spec lookup

ak684:
## Summary

The app looks up its default sandbox spec by `OH_SANDBOX_SPEC_DEFAULT_SPEC_NAME=v1_current`, while the installer-managed warm-runtime entry rendered as `default`. With multiple configs the lookup missed and fell back to list order.

This aligns the rendered runtime name without breaking the existing Helm values contract:

- Keep `configsByName.default` as the stable map key in the runtime-api subchart, umbrella values, and KOTS helmValues.
- Add `name: v1_current` to that entry.
- Update the map normalizer so an explicit entry name wins; map keys are still promoted when no explicit name is present.
- Preserve Helm-direct partial overrides such as `configsByName.default.count=4`.
- Add a cross-layer invariant test tying the app env default to the umbrella, subchart, and Replicated warm-runtime entry names.

The app, reconciler, list API, and docs now see `v1_current`, while existing customer values continue merging into the complete installer entry.

## Upgrade impact

- Helm-direct installs do not need to rename existing `configsByName.default.*` overrides.
- The config entry inside `warm-runtimes.json` changes to `v1_current`; the ConfigMap object remains `warm-runtimes-config`.
- Warm pod claiming is content-based (image, command, working directory, and environment), so the name-only change does not churn otherwise matching pods.

## Testing

- Runtime-api subchart Helm unit tests: 13 passed.
- Umbrella Helm unit tests: 109 passed.
- Full script suite: 479 passed.
- Runtime-api Helm lint: clean.
- Direct render with `configsByName.default.count=4`: valid JSON, one complete `v1_current` entry, count 4.

### DIFF
--- a/charts/openhands/charts/runtime-api/templates/warm-runtimes-configmap.yaml
+++ b/charts/openhands/charts/runtime-api/templates/warm-runtimes-configmap.yaml
@@ -10,7 +10,11 @@
 {{- if not $configs }}
 {{- $configs = list }}
 {{- range $name, $config := .Values.warmRuntimes.configsByName }}
-{{- $configs = append $configs (merge (dict "name" $name) $config) }}
+{{- $entry := deepCopy $config }}
+{{- if not (hasKey $entry "name") }}
+{{- $_ := set $entry "name" $name }}
+{{- end }}
+{{- $configs = append $configs $entry }}
 {{- end }}
 {{- end }}
 {{- /* Default image for any entry that omits one: the global agent-server image. */ -}}
--- a/charts/openhands/charts/runtime-api/tests/warm_runtimes_configmap_test.yaml
+++ b/charts/openhands/charts/runtime-api/tests/warm_runtimes_configmap_test.yaml
@@ -21,3 +21,21 @@ tests:
       - notMatchRegex:
           path: data["warm-runtimes.json"]
           pattern: fuse_s3_mount
+
+  - it: preserves legacy default-key overrides while rendering v1_current
+    set:
+      warmRuntimes.enabled: true
+      warmRuntimes.configsByName.default.count: 4
+    asserts:
+      - matchRegex:
+          path: data["warm-runtimes.json"]
+          pattern: '"name": "v1_current"'
+      - matchRegex:
+          path: data["warm-runtimes.json"]
+          pattern: '"working_dir": "/workspace/project"'
+      - matchRegex:
+          path: data["warm-runtimes.json"]
+          pattern: '"count": 4'
+      - notMatchRegex:
+          path: data["warm-runtimes.json"]
+          pattern: '"name": "default"'
--- a/charts/openhands/charts/runtime-api/values.yaml
+++ b/charts/openhands/charts/runtime-api/values.yaml
@@ -283,16 +283,16 @@ warmRuntimes:
   # Legacy list form. Kept for backwards compatibility: when non-empty it takes
   # precedence over configsByName. Prefer configsByName for new configuration.
   configs: []
-  # Canonical map form. Each key is the config's name (promoted to "name" in the
-  # rendered JSON). Maps merge cleanly across values layers, so consumers can
-  # override one field of one entry without re-declaring every entry.
+  # Canonical map form. Each key becomes the config name unless the entry sets
+  # `name` explicitly. Maps merge cleanly across values layers.
   # Set `fuse_s3_mount: true` on an entry for fusey warm pods: dormant startup,
   # /workspace FUSE-mounted from object storage instead of a PVC. Needs the
   # FUSEY_* env on runtime-api and /dev/fuse on the sandbox nodes (see the
   # device-plugin subchart).
   configsByName:
-    # Entry name must match the app's OH_SANDBOX_SPEC_DEFAULT_SPEC_NAME.
-    v1_current:
+    # Keep the map key stable for existing Helm overrides.
+    default:
+      name: v1_current
       # image omitted -> defaults to global.agentServerImage (repo:tag).
       working_dir: "/workspace/project"
       environment:
--- a/charts/openhands/charts/runtime-api/values.yaml
+++ b/charts/openhands/charts/runtime-api/values.yaml
@@ -291,7 +291,8 @@ warmRuntimes:
   # FUSEY_* env on runtime-api and /dev/fuse on the sandbox nodes (see the
   # device-plugin subchart).
   configsByName:
-    default:
+    # Entry name must match the app's OH_SANDBOX_SPEC_DEFAULT_SPEC_NAME.
+    v1_current:
       # image omitted -> defaults to global.agentServerImage (repo:tag).
       working_dir: "/workspace/project"
       environment:
--- a/charts/openhands/values.yaml
+++ b/charts/openhands/values.yaml
@@ -939,9 +939,9 @@ runtime-api:
     # Map form keyed by config name; see runtime-api subchart values for the
     # legacy `configs` list and precedence rules.
     configsByName:
-      # Entry name must match OH_SANDBOX_SPEC_DEFAULT_SPEC_NAME (_env.yaml) so the
-      # app's default sandbox spec lookup resolves by name instead of list order.
-      v1_current:
+      # Keep the map key stable for existing Helm overrides.
+      default:
+        name: v1_current
         # image omitted -> defaults to global.agentServerImage (repo:tag).
         # Must match the app's sandbox spec working_dir, or no warm pod is ever claimable.
         working_dir: "/workspace/project"
--- a/charts/openhands/values.yaml
+++ b/charts/openhands/values.yaml
@@ -939,7 +939,9 @@ runtime-api:
     # Map form keyed by config name; see runtime-api subchart values for the
     # legacy `configs` list and precedence rules.
     configsByName:
-      default:
+      # Entry name must match OH_SANDBOX_SPEC_DEFAULT_SPEC_NAME (_env.yaml) so the
+      # app's default sandbox spec lookup resolves by name instead of list order.
+      v1_current:
         # image omitted -> defaults to global.agentServerImage (repo:tag).
         # Must match the app's sandbox spec working_dir, or no warm pod is ever claimable.
         working_dir: "/workspace/project"
--- a/replicated/openhands.yaml
+++ b/replicated/openhands.yaml
@@ -379,7 +379,9 @@ spec:
         enabled: true
         count: repl{{ ConfigOption "sandbox_warm_runtime_count" }}
         configsByName:
-          default:
+          # Entry name must match the app's OH_SANDBOX_SPEC_DEFAULT_SPEC_NAME
+          # (and the chart values' entry, or the layers stop merging).
+          v1_current:
             # image omitted -> inherits global.agentServerImage. Proxy by
             # default; when custom_sandbox_image_enabled / HasLocalRegistry are
             # active, the optionalValues blocks below override the global instead.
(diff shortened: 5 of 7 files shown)


aivong-openhands (commented):
## Code review — `fix: rename warm-runtime default config to v1_current`

Reviewed against `.agents/skills/code-review.md`. **No blocking issues** — the change is correct and I verified the mechanism end to end. Two suggestions below, one of which I think is worth doing before merge.

### What I verified

- `OH_SANDBOX_SPEC_DEFAULT_SPEC_NAME: v1_current` in `charts/openhands/templates/_env.yaml` is consumed by `DynamicRemoteSandboxSpecService.get_default_sandbox_spec()` (OpenHands/sandbox-server), which does `self._name_to_spec.get(self.default_spec_name)` and falls back to `specs[0]`. The name really was unresolvable before this PR, so the fix is real.
- `v1_current` is also the name runtime-api uses in its own hardcoded `_get_default_config()`, and it is the name `saas-deploy` uses in the legacy `configs` list for dev/staging/prod. This PR makes the chart/Replicated layer agree with everything else rather than inventing a fourth spelling.
- Rendered the configmap on this branch: single `v1_current` entry, content identical to before. Warm pods are matched by content in `warm_runtimes.py` (`warm_runtime_matches` on image/command/working_dir/env), so the "no pool churn on upgrade" claim in the description holds.
- No in-repo consumer overrides `configsByName.default`, and `saas-deploy` dev/staging/prod all use the legacy `configs` list form (which wins over `configsByName`), so this is a no-op for our environments — consistent with "defaults serve the self-hosted user".

### Suggestion 1 (worth doing): nothing enforces the cross-layer invariant

The name `v1_current` now has to stay in lockstep across four places — `_env.yaml`, `charts/openhands/values.yaml`, the runtime-api subchart values, and `replicated/openhands.yaml` — and the only thing holding it together is three comments. Comments do not travel with the key, and this exact class of "two layers silently disagree and the effect is invisible until runtime" is what the PR is fixing.

`charts/openhands/tests/sandbox_hostname_layout_test.yaml` is the existing precedent for this shape of test: it compiles the parent `deployment.yaml` alongside `charts/runtime-api/templates/...` and asserts the two sides agree. A small suite doing the same here — assert the rendered `warm-runtimes.json` contains a config whose `name` equals the `OH_SANDBOX_SPEC_DEFAULT_SPEC_NAME` value on the app deployment — would make the next rename fail in CI instead of silently re-introducing the bug. The current `warm_runtimes_configmap_test.yaml` only covers `fuse_s3_mount`, so nothing catches drift here today.

### Suggestion 2 (optional): a stale `default` override renders invalid JSON, not an error

The upgrade-impact note is honest about Helm-direct installs needing to rename their override key, but the failure mode is worse than "a partial second entry". On this branch:

```
helm template rt charts/openhands/charts/runtime-api \
  --set warmRuntimes.enabled=true --set warmRuntimes.count=1 \
  --set warmRuntimes.configsByName.default.count=2 \
  -s templates/warm-runtimes-configmap.yaml
```

renders

```json
{ "name": "default", "working_dir": , "command": null, ... }
```

which is **not parseable JSON**. The ConfigMap applies cleanly, and the breakage only surfaces later in the warm-runtimes CronJob and in the app's `/api/warm-runtime-configs` fetch. (With `working_dir` also set it parses, but you get a live orphan entry with `"command": null`.)

This weakness pre-dates the PR — any partially-specified entry does it — but this PR is what makes a real consumer likely to hit it. `charts/openhands/templates/validations.yaml` already establishes the pattern of failing the render with a clear message rather than shipping a broken object; a `fail` in `warm-runtimes-configmap.yaml` when an entry omits `working_dir` or `command` would turn this into "rename your key" at upgrade time. Reasonable to split into a follow-up if you would rather keep this PR a pure rename.

### Nit

The description says the config list "has no defined order". Helm's `range` over a map iterates keys in sorted order, so the pre-PR fallback to `specs[0]` was deterministic — it just happened to land on `default` because `d` sorts first. The argument for the change is unaffected (an entry named e.g. `custom` would have silently stolen the default), but the "non-deterministic" framing overstates it slightly.

---

_This review was generated by an AI agent (OpenHands), following the repository's `.agents/skills/code-review.md`._

aivong-openhands (commented):
## Code review

Reviewed against the repo's [code-review skill](https://github.com/OpenHands/OpenHands-Cloud/blob/main/.agents/skills/code-review.md).

The core fix is right and the diff is small. Aligning the rendered entry name with `OH_SANDBOX_SPEC_DEFAULT_SPEC_NAME` is clearly the correct target, and the `deepCopy` before `set` is the right call — the previous `merge (dict "name" $name) $config` would have been the obvious-but-wrong way to do this. A few things I'd like to talk through before this goes in.

---

### 1. The literal `v1_current` now lives in five files with nothing linking them (important)

After this PR, `v1_current` is hand-written in:

- `charts/openhands/templates/_env.yaml:8` — `OH_SANDBOX_SPEC_DEFAULT_SPEC_NAME`
- `charts/openhands/charts/runtime-api/values.yaml`
- `charts/openhands/values.yaml`
- `replicated/openhands.yaml`
- `scripts/update_openhands_charts/conftest.py` (x2 fixtures)

The bug being fixed here *is* a silent drift between two of those strings. The fix resolves today's drift but widens the surface from two copies to five, and the failure mode on the next drift is unchanged: lookup misses, silent fallback to list order, and everything looks healthy. Nothing in CI catches it.

This repo already treats exactly this class of cross-file invariant as worth guarding — `check-agent-server-sync.yml` exists because "nothing links the two pins, so a bump to either one alone silently ships that split." Same shape of problem here.

Cheapest thing that would have caught the original bug: a helm-unittest assertion on the umbrella chart that the rendered `warm-runtimes.json` entry name equals the `OH_SANDBOX_SPEC_DEFAULT_SPEC_NAME` the app deployment renders. If that's awkward to express across the two templates, even a comment at `_env.yaml:8` pointing at the warm-runtime config (and vice versa) is better than nothing.

Related: `OH_SANDBOX_SPEC_DEFAULT_SPEC_NAME` is a hardcoded literal in `_env.yaml`, but a consumer *can* override it through the `.Values.env` de-dup path. Anyone who does that silently re-breaks warm claiming unless they also know to change `configsByName.<entry>.name`. Worth a comment on at least one side.

### 2. Test coverage sits on the copy that doesn't ship

The new regression test is a good one, and asserting `notMatchRegex '"name": "default"'` is the right negative to pin. But it runs in the runtime-api subchart suite against the *subchart's* `values.yaml`.

The copies that actually reach users are `charts/openhands/values.yaml` (Helm installs) and `replicated/openhands.yaml` (KOTS). Neither is asserted anywhere. Someone can edit the umbrella `name: v1_current` back out and the whole suite still goes green — which is the same hole the original bug fell through. `charts/openhands/tests/sandbox_hostname_layout_test.yaml` already reaches into `charts/runtime-api/templates/`, so there's a working pattern for an umbrella-level assertion.

### 3. Map key no longer describes what it renders

Principle 2 in the review guidelines asks that a consumer be able to predict what a key does from its name alone, without internal context. The subchart values previously documented `configsByName` as "Each key is the config's name (promoted to `name` in the rendered JSON)" — a rule you could read off the key. After this PR, the key is `default`, the rendered name is `v1_current`, and the only thing reconciling them is the comment `# Keep the map key stable for existing Helm overrides` — which is repo-internal context that doesn't travel into a consumer's values file.

I understand the tradeoff you're making and the upgrade-impact section makes the reasoning explicit, so this may well be the right call. But I'd like to check the premise: how real is the compat surface? `configsByName` looks recent, ships exactly one entry, and the only in-repo consumer (`local-kind/kind-resources.yaml`) just sets `enabled: false`. If no external site values are pinning `configsByName.default.*` yet, renaming the key to `v1_current` outright is strictly simpler — key and rendered name agree again, the values doc stays true, and the `hasKey` branch in the template isn't needed at all. If there *are* known external consumers, say so in the description and I'll drop this.

### 4. Question on reconciler behaviour during the rename

The description says claiming is content-based so matching pods won't churn. That covers claiming — what about the pool itself? On upgrade the reconciler sees the `default` entry disappear and a `v1_current` entry appear. Does it scale the old pool down cleanly, or is there a window with orphaned `default` pods that nothing owns? Not blocking, but worth a sentence in the description since this ships to self-hosted upgrades.

### 5. Nit

"The rendered ConfigMap name changes to `v1_current`" in the upgrade-impact section — the ConfigMap name is unchanged (`configMapName: warm-runtimes-config`); it's the config *entry* name inside `warm-runtimes.json` that changes. Worth correcting so the upgrade note doesn't send someone hunting for a renamed object.

---

Not blocking on 3–5. I'd like to see a guard for 1 and an umbrella-level assertion for 2 before merge, since without them the next occurrence of this exact bug is just as invisible as this one was.

---

aivong-openhands (approved):
## Code review

Reviewed against the repo's [code-review skill](https://github.com/OpenHands/OpenHands-Cloud/blob/main/.agents/skills/code-review.md). Approving — the two things I flagged as important on the previous pass are now covered, and nothing new rises above nit level.

### The cross-layer guard closes both gaps

`scripts/test_warm_runtime_default_name.py` is the right shape for this. I ran it on the branch (passes) and it pins all four copies of the name to a single set:

- `OH_SANDBOX_SPEC_DEFAULT_SPEC_NAME` in `charts/openhands/templates/_env.yaml`
- `charts/openhands/values.yaml` (Helm installs)
- `charts/openhands/charts/runtime-api/values.yaml`
- `replicated/openhands.yaml` (KOTS)

That also covers point 2 from my last review: the two copies that actually ship to users are now asserted, not just the subchart's. Putting it in `scripts/` rather than helm-unittest means it runs under `test-scripts.yml`, which triggers on `charts/**` and `replicated/**` — so editing the umbrella `name:` back out fails CI even though the test file itself didn't change. `replicated/openhands.yaml` parses cleanly under `yaml.safe_load` despite the `repl{{ }}` templating, so the assertion is real and not silently skipping a layer.

### Template change verified

Rendered every branch of the normalizer with helm 3.18.4:

| values | rendered `name` |
|---|---|
| defaults | `v1_current` |
| `configsByName.default.count=4` | `v1_current`, `count: 4`, rest of entry intact |
| `configsByName.default.name=mine` | `mine` (explicit wins) |
| new key `custom`, no `name` | `custom` (key still promoted) |

The `deepCopy` before `set` is doing real work — `set` mutates in place, so without it the normalizer would write back into `.Values`. Worth keeping as-is.

### Nits (not blocking)

**Stale comment on the normalizer.** The header block in `warm-runtimes-configmap.yaml` still says the map form works by "promoting each key to the entry's `name`", which is now only the fallback. The values files got updated for this; the template header didn't. One line, so the next reader doesn't have to derive the precedence from the `hasKey` branch.

**Regex is slightly brittle.** `value: ([^\s]+)` in the new test captures the raw token, so if anyone quotes the value in `_env.yaml` (`value: "v1_current"`) the test fails on the quotes rather than on real drift. Stripping quotes off the capture would make the failure mode honest.

**Pre-existing, now more reachable.** A partially-specified entry still renders unparseable JSON — `--set warmRuntimes.configsByName.custom.count=2` alone produces `"working_dir": ,`. Same class: a consumer who keys their own entry `v1_current` gets two entries with the same name and no error. Neither is this PR's doing and both are fine as follow-ups; I'm not asking for them here.

### On the key/name mismatch

I raised last pass that `configsByName.default` rendering as `v1_current` reads oddly against principle 2. You've kept the key for merge compatibility and said so explicitly in both the values comments and the upgrade-impact section, and the invariant test now makes the coupling machine-checked rather than comment-checked. Reasonable place to land — the readability cost is bounded and the compat story is documented.

CI is green across the board (helm-unittest, test-scripts, all four lint-and-test matrix legs, appversion drift, secret checksum).

---

PR #1182, opened 2026-08-31
fix: give the Runtime API admin password a real KOTS field with a generated default

ak684:
## Summary

The `admin_password` item was hidden in a disabled config group and had no generated value. Stock Replicated installs therefore rendered an empty Secret, manual Secret patches were overwritten on deploy, and password changes did not restart runtime-api even though the value is consumed through `secretKeyRef`.

## Changes

- Move `admin_password` into the visible Sandbox Configuration group as a masked Runtime API Admin Password.
- Seed unset installs with `value: '{{repl RandomString 32}}'`. With a non-readonly password item, KOTS persists the generated value and user-set values win on later deploys.
- Add runtime-api `podAnnotations` support.
- Add `checksum/admin-password` to the runtime-api pod template so only password changes alter that checksum and roll the deployment.
- Add Helm unit coverage proving annotations render and remain absent by default.

`admin_password` remains in the existing `admin-password` Secret template and in the global secrets checksum. OpenHands/runtime-api#730 independently fails closed if an existing install still has an empty stored value.

## Testing

- Runtime-api Helm unit tests: 14 passed.
- Runtime-api Helm lint: clean.
- `scripts/check_secret_checksum.py`: all 43 password fields covered.
- Replicated manifest validation remains covered by PR CI.

### DIFF
--- a/charts/openhands/charts/runtime-api/templates/deployment.yaml
+++ b/charts/openhands/charts/runtime-api/templates/deployment.yaml
@@ -19,6 +19,10 @@ spec:
       {{- include "runtime-api.selectorLabels" . | nindent 6 }}
   template:
     metadata:
+      {{- with .Values.podAnnotations }}
+      annotations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       labels:
         {{- include "runtime-api.selectorLabels" . | nindent 8 }}
         {{- /* Distinguishes API pods from the warm-runtimes CronJob pods
--- a/replicated/config.yaml
+++ b/replicated/config.yaml
@@ -608,11 +608,6 @@ spec:
       description: Configure passwords and secrets for various services
       when: false
       items:
-        - name: admin_password
-          title: Admin Password
-          help_text: Admin password for Runtime API
-          type: password
-          hidden: true
         - name: postgres_username
           title: PostgreSQL Username
           help_text: Username for embedded PostgreSQL database
@@ -1276,6 +1271,11 @@ spec:
             regex:
               pattern: ^(0|[1-9][0-9]*)$
               message: Must be a non-negative integer
+        - name: admin_password
+          title: Runtime API Admin Password
+          help_text: Password for the Runtime API admin endpoints, used to manage warm runtime configurations (for example, running multiple custom sandbox images side by side). A random password is generated at install; to use the admin API, set your own password here and deploy. Changing this password restarts the runtime-api service.
+          type: password
+          value: '{{repl RandomString 32}}'
         - name: sandbox_max_num_sandboxes
           title: Max Concurrent Sandboxes
           help_text: >-
--- a/replicated/openhands.yaml
+++ b/replicated/openhands.yaml
@@ -293,6 +293,10 @@ spec:
       replicaCount: 1
       autoscaling:
         enabled: false
+      # ADMIN_PASSWORD comes from a secretKeyRef, so a password change needs a
+      # rollout to take effect (same pattern as checksum/litellm-credentials).
+      podAnnotations:
+        checksum/admin-password: 'repl{{ sha256sum (ConfigOption "admin_password") }}'
       image:
         repository: 'images.r9.all-hands.dev/proxy/{{repl LicenseFieldValue "appSlug"}}/ghcr.io/openhands/runtime-api'
         # Do not override the image tag here. Pin to a stable version in the chart.
--- a/charts/openhands/charts/runtime-api/tests/pod_annotations_test.yaml
+++ b/charts/openhands/charts/runtime-api/tests/pod_annotations_test.yaml
@@ -1,0 +1,17 @@
+suite: pod annotations
+templates:
+  - templates/deployment.yaml
+tests:
+  - it: renders configured pod annotations
+    set:
+      podAnnotations:
+        checksum/admin-password: password-hash
+    asserts:
+      - equal:
+          path: spec.template.metadata.annotations["checksum/admin-password"]
+          value: password-hash
+
+  - it: omits pod annotations by default
+    asserts:
+      - notExists:
+          path: spec.template.metadata.annotations
--- a/charts/openhands/charts/runtime-api/values.yaml
+++ b/charts/openhands/charts/runtime-api/values.yaml
@@ -10,6 +10,10 @@ fullnameOverride: ""
 
 replicaCount: 1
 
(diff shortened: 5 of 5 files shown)


replicated/openhands.yaml
```diff
@@ -293,6 +293,10 @@ spec:
       replicaCount: 1
       autoscaling:
         enabled: false
+      # ADMIN_PASSWORD comes from a secretKeyRef, so a password change needs a
+      # rollout to take effect (same pattern as checksum/litellm-credentials).
+      podAnnotations:
+        checksum/admin-password: 'repl{{ sha256sum (ConfigOption "admin_password") }}'
```
aivong-openhands, line 299: Question, not a blocker: this rolls runtime-api when `admin_password` changes, but `ADMIN_PASSWORD` is not the only secret-backed env on this Deployment — `DEFAULT_API_KEY` (`defaultAPIKeySecretName`, deployment.yaml:83) and `DB_PASS` (`_env.yaml:126`) are `secretKeyRef` too and have exactly the same stale-pod problem.

Would reusing the existing `secretsChecksum` value here be better than a bespoke single-field hash? Two reasons it appeals:

1. It covers all three secrets at once rather than one.
2. `scripts/check_secret_checksum.py` already guards `secretsChecksum` against drift when someone adds a password field. A one-off annotation is invisible to that guard, so the next secret runtime-api consumes silently repeats this bug.

The trade-off is noisier rollouts (runtime-api restarts when any of the 43 secrets change), which is the trade-off the openhands Deployment already accepts. If you chose the narrow annotation deliberately to avoid that, a one-line note saying so would stop the next person from "fixing" it — and the other two env vars would still be worth a follow-up.

charts/openhands/charts/runtime-api/templates/deployment.yaml
```diff
@@ -19,6 +19,10 @@ spec:
       {{- include "runtime-api.selectorLabels" . | nindent 6 }}
   template:
     metadata:
+      {{- with .Values.podAnnotations }}
+      annotations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
```
aivong-openhands, line 25: The PR body says this was verified by rendering the template by hand. `charts/openhands/charts/runtime-api/tests/` already exists (`labels_test.yaml`, `scheduling_test.yaml`) and is wired into the `helm-unittest` workflow, so a small case would make that verification permanent — set `podAnnotations.checksum/admin-password` and assert it lands at `spec.template.metadata.annotations`.

Worth it mainly because this annotation is load-bearing for the fix: if a refactor drops the `with` block, password changes silently stop taking effect again, which is the exact failure mode the PR exists to close.

replicated/config.yaml
```diff
@@ -1276,6 +1271,11 @@ spec:
             regex:
               pattern: ^(0|[1-9][0-9]*)$
               message: Must be a non-negative integer
+        - name: admin_password
+          title: Runtime API Admin Password
+          help_text: Password for the Runtime API admin endpoints, used to manage warm runtime configurations (for example, running multiple custom sandbox images side by side). A random password is generated at install; to use the admin API, set your own password here and deploy. Changing this password restarts the runtime-api service.
+          type: password
+          value: '{{repl RandomString 32}}'
```
aivong-openhands, line 1278: Nit on placement: this lands between `sandbox_warm_runtime_count` and `sandbox_max_num_sandboxes`, splitting an otherwise contiguous run of sandbox sizing/count fields. The adjacency-to-warm-runtime-count rationale makes sense, but so would the end of the group, after `custom_sandbox_image_*` — the admin API's headline use (multiple custom sandbox images side by side, per your own help text) sits closer to those. Either is fine; flagging only because Config screen ordering is what operators actually navigate by.

aivong-openhands (approved):
## Code review — `fix: give the Runtime API admin password a real KOTS field with a generated default`

**Approving.** The diagnosis is right and the fix is the minimal one: `admin_password` was the only `type: password` item in the hidden `security_secrets` group without a `RandomString` value, so every stock install rendered the `admin-password` Secret empty. Combined with a runtime-api that accepted an empty password on an ingress that is publicly exposed at `computed_runtime_api_hostname`, that is worth fixing on the install side regardless of what OpenHands/runtime-api#730 does on the service side.

Things I checked that hold up:

- **Follows the established pattern.** The new field mirrors `litellm_admin_password` exactly: visible group, `type: password`, a `RandomString 32` value, no `required`, so it can never block a deploy. The pod annotation mirrors `checksum/litellm-credentials` at `replicated/openhands.yaml:220`.
- **Checksum coverage stays green.** `admin_password` was already in `secretsChecksum`, and `scripts/check_secret_checksum.py` parses `config.yaml` structurally, so moving the field between groups does not change what it sees.
- **`podAnnotations: {}` with `{{- with }}`** is the standard Helm shape and is inert for every consumer who does not set it. The subchart genuinely had no annotation support before, so the addition is necessary rather than incidental.
- **Defaults serve the self-hosted user** (principle 7): a generated per-install password is the right shipped default, and nothing here is tuned to our dev/staging/prod.
- **The new values key names an effect, not a scenario** (principle 2): `podAnnotations` is established Helm vocabulary, and the comment explains why you would set it without the key depending on the comment to be legible.

Three non-blocking notes inline: the narrow scope of the new checksum, a missing `helm unittest` case, and field placement within the Sandbox group. None are reasons to hold this up.

### Risk assessment

**Low-to-moderate, and lower than the status quo.**

- *Blast radius*: one new pod annotation on the runtime-api Deployment (rolling the pod once on a deploy that changes the password) and one relocated config item. No data path, no storage backend, no dependency toggle.
- *Behavior change on upgrade*: an install whose stored value is empty may or may not get a generated password, depending on how KOTS treats stored-empty. The PR body reasons through both outcomes and both are acceptable — but worth confirming empirically after the first upgrade (read the rendered `admin-password` Secret and check its length) rather than leaving it as an inference, since "generated on upgrade" also means runtime-api rolls once on that upgrade.
- *Operators who hand-patched the Secret*: their patch was already overwritten on every deploy; now it is overwritten with a real value instead of an empty one. Worth a release-note line, because the remedy (enter the password in the Config screen) is easy but not discoverable from the symptom.
- *Secret handling*: the annotation stores a sha256, not the password. Correct.

---
_This review was generated by an AI agent (OpenHands) on behalf of the requesting user, following the repository's `.agents/skills/code-review.md` guidelines._

PR #1183, opened 2026-08-31
feat: enable warm-runtime config overlay mode for Replicated

ak684:
## Summary

Enable `WARM_RUNTIME_CONFIG_OVERLAY=1` in the Replicated/OHE runtime-api helmValues so OpenHands Enterprise uses the effective ConfigMap-plus-database behavior from OpenHands/runtime-api#731.

The shared runtime-api env helper passes the Replicated value to both consumers that resolve warm configs:

- the runtime-api Deployment that serves `GET /api/warm-runtime-configs`; and
- the warm-runtimes reconciler CronJob.

Activation is intentionally scoped to Replicated. The shared runtime-api subchart default remains unset, so this PR does not silently change SaaS or Helm-direct installs. Consumers can still set the variable explicitly to `0` or `1`.

Under overlay semantics, a same-named database row overrides the installer entry, a differently-named row runs beside it, and deleting a same-name override reveals the live installer entry again.

## Rollout

The variable is inert on runtime-api images that predate #731, so this chart change can land before the runtime-api pin is updated. Overlay behavior activates for OHE only when its pinned image contains #731.

## Validation

A fresh R02 Replicated/OHE install using a combined validation image confirmed:

- `v1_current` from the KOTS ConfigMap remained visible after adding `custom-ghcr` through the admin API;
- the effective list reported `source: file` and `source: db` respectively;
- the reconciler maintained one ready pool for each image;
- a same-name `v1_current` database row overrode the file entry; and
- deleting that row immediately restored the ConfigMap entry.

The full browser smoke stopped at an unexpected Bitbucket Data Center auth redirect, so the feature-specific live API/reconciler checks pass but the overall click test is not marked PASS.

## Testing

- Runtime-api subchart Helm unit tests: 15 passed, including API, reconciler, and explicit opt-out propagation.
- Umbrella Helm unit tests: 111 passed.
- Full script suite: 479 passed, including a scope guard that requires Replicated to enable the flag while the shared subchart default remains unset.

### DIFF
--- a/charts/openhands/charts/runtime-api/tests/warm_runtime_overlay_env_test.yaml
+++ b/charts/openhands/charts/runtime-api/tests/warm_runtime_overlay_env_test.yaml
@@ -38,3 +38,8 @@ tests:
           content:
             name: WARM_RUNTIME_CONFIG_OVERLAY
             value: "0"
+      - notContains:
+          path: spec.template.spec.containers[0].env
+          content:
+            name: WARM_RUNTIME_CONFIG_OVERLAY
+            value: "1"
--- a/charts/openhands/charts/runtime-api/tests/warm_runtime_overlay_env_test.yaml
+++ b/charts/openhands/charts/runtime-api/tests/warm_runtime_overlay_env_test.yaml
@@ -3,8 +3,11 @@ templates:
   - templates/deployment.yaml
   - templates/warm-runtimes-cronjob.yaml
 tests:
-  - it: enables overlay mode in the API deployment
+  - it: passes OHE overlay mode into the API deployment with warm pools enabled
     template: templates/deployment.yaml
+    set:
+      warmRuntimes.enabled: true
+      env.WARM_RUNTIME_CONFIG_OVERLAY: "1"
     asserts:
       - contains:
           path: spec.template.spec.containers[0].env
@@ -12,13 +15,26 @@ tests:
             name: WARM_RUNTIME_CONFIG_OVERLAY
             value: "1"
 
-  - it: enables overlay mode in the reconciler CronJob
+  - it: passes OHE overlay mode into the reconciler CronJob
     template: templates/warm-runtimes-cronjob.yaml
     set:
       warmRuntimes.enabled: true
+      env.WARM_RUNTIME_CONFIG_OVERLAY: "1"
     asserts:
       - contains:
           path: spec.jobTemplate.spec.template.spec.containers[0].env
           content:
             name: WARM_RUNTIME_CONFIG_OVERLAY
             value: "1"
+
+  - it: preserves an explicit overlay opt-out
+    template: templates/deployment.yaml
+    set:
+      warmRuntimes.enabled: true
+      env.WARM_RUNTIME_CONFIG_OVERLAY: "0"
+    asserts:
+      - contains:
+          path: spec.template.spec.containers[0].env
+          content:
+            name: WARM_RUNTIME_CONFIG_OVERLAY
+            value: "0"
--- a/charts/openhands/charts/runtime-api/tests/warm_runtime_overlay_env_test.yaml
+++ b/charts/openhands/charts/runtime-api/tests/warm_runtime_overlay_env_test.yaml
@@ -1,0 +1,24 @@
+suite: warm runtime overlay environment
+templates:
+  - templates/deployment.yaml
+  - templates/warm-runtimes-cronjob.yaml
+tests:
+  - it: enables overlay mode in the API deployment
+    template: templates/deployment.yaml
+    asserts:
+      - contains:
+          path: spec.template.spec.containers[0].env
+          content:
+            name: WARM_RUNTIME_CONFIG_OVERLAY
+            value: "1"
+
+  - it: enables overlay mode in the reconciler CronJob
+    template: templates/warm-runtimes-cronjob.yaml
+    set:
+      warmRuntimes.enabled: true
+    asserts:
+      - contains:
+          path: spec.jobTemplate.spec.template.spec.containers[0].env
+          content:
+            name: WARM_RUNTIME_CONFIG_OVERLAY
+            value: "1"
--- a/charts/openhands/charts/runtime-api/tests/warm_runtime_overlay_env_test.yaml
+++ b/charts/openhands/charts/runtime-api/tests/warm_runtime_overlay_env_test.yaml
@@ -43,3 +43,20 @@ tests:
           content:
             name: WARM_RUNTIME_CONFIG_OVERLAY
             value: "1"
+
+  - it: preserves an explicit overlay opt-out in the reconciler CronJob
+    template: templates/warm-runtimes-cronjob.yaml
+    set:
+      warmRuntimes.enabled: true
+      env.WARM_RUNTIME_CONFIG_OVERLAY: "0"
+    asserts:
+      - contains:
+          path: spec.jobTemplate.spec.template.spec.containers[0].env
+          content:
+            name: WARM_RUNTIME_CONFIG_OVERLAY
+            value: "0"
+      - notContains:
+          path: spec.jobTemplate.spec.template.spec.containers[0].env
+          content:
+            name: WARM_RUNTIME_CONFIG_OVERLAY
+            value: "1"
--- a/charts/openhands/charts/runtime-api/values.yaml
+++ b/charts/openhands/charts/runtime-api/values.yaml
@@ -50,6 +50,9 @@ service:
   port: 5000
 
 env:
+  # Database warm-runtime configs overlay the installer-managed ConfigMap by
+  # name instead of replacing it (runtime-api#731). Inert on older images.
+  WARM_RUNTIME_CONFIG_OVERLAY: "1"
   PYTHONDONTWRITEBYTECODE: "1"
   PYTHONUNBUFFERED: "1"
   DEV_MODE: "false"
--- a/charts/openhands/charts/runtime-api/values.yaml
+++ b/charts/openhands/charts/runtime-api/values.yaml
@@ -50,9 +50,6 @@ service:
   port: 5000
 
 env:
-  # Database warm-runtime configs overlay the installer-managed ConfigMap by
-  # name instead of replacing it (runtime-api#731). Inert on older images.
-  WARM_RUNTIME_CONFIG_OVERLAY: "1"
   PYTHONDONTWRITEBYTECODE: "1"
   PYTHONUNBUFFERED: "1"
   DEV_MODE: "false"
--- a/replicated/openhands.yaml
+++ b/replicated/openhands.yaml
@@ -328,6 +328,10 @@ spec:
         PGSSLMODE: '{{repl if ConfigOptionEquals "postgres_type" "external_postgres"}}{{repl ConfigOption "external_postgres_ssl_mode"}}{{repl else}}prefer{{repl end}}'
         DB_USER: '{{repl if ConfigOptionEquals "postgres_type" "external_postgres"}}{{repl ConfigOption "external_postgres_username"}}{{repl else}}postgres{{repl end}}'
         DB_NAME: "runtime_api_db"
+        # Admin-API warm configs overlay installer-managed entries by name:
+        # same-named rows override them and differently-named rows run beside
+        # them. Older runtime-api images ignore this variable.
+        WARM_RUNTIME_CONFIG_OVERLAY: "1"
         RUNTIME_BASE_URL: '{{repl ConfigOption "computed_runtime_base_hostname" }}'
         # Joins the sandbox ID to RUNTIME_BASE_URL. "-" keeps each sandbox one
         # label deep ({id}-runtime.<base>) so a single *.<base> wildcard covers
--- a/scripts/test_replicated_warm_runtime_overlay.py
+++ b/scripts/test_replicated_warm_runtime_overlay.py
@@ -14,6 +14,7 @@ def _yaml(path: str) -> dict:
 def test_replicated_enables_overlay_without_changing_shared_chart_defaults() -> None:
     replicated = _yaml("replicated/openhands.yaml")
     subchart = _yaml("charts/openhands/charts/runtime-api/values.yaml")
+    umbrella = _yaml("charts/openhands/values.yaml")
 
     assert (
         replicated["spec"]["values"]["runtime-api"]["env"][
@@ -22,3 +23,4 @@ def test_replicated_enables_overlay_without_changing_shared_chart_defaults() -> 
         == "1"
     )
     assert "WARM_RUNTIME_CONFIG_OVERLAY" not in subchart["env"]
+    assert "WARM_RUNTIME_CONFIG_OVERLAY" not in umbrella["runtime-api"]["env"]
--- a/scripts/test_replicated_warm_runtime_overlay.py
+++ b/scripts/test_replicated_warm_runtime_overlay.py
@@ -1,0 +1,24 @@
+"""Keep warm-runtime overlay activation scoped to Replicated installs."""
+
+from pathlib import Path
+
+import yaml
+
+REPO_ROOT = Path(__file__).resolve().parents[1]
+
+
+def _yaml(path: str) -> dict:
+    return yaml.safe_load((REPO_ROOT / path).read_text(encoding="utf-8"))
+
+
+def test_replicated_enables_overlay_without_changing_shared_chart_defaults() -> None:
+    replicated = _yaml("replicated/openhands.yaml")
+    subchart = _yaml("charts/openhands/charts/runtime-api/values.yaml")
+
+    assert (
+        replicated["spec"]["values"]["runtime-api"]["env"][
+            "WARM_RUNTIME_CONFIG_OVERLAY"
+        ]
+        == "1"
+    )
+    assert "WARM_RUNTIME_CONFIG_OVERLAY" not in subchart["env"]


aivong-openhands (commented):
## Code review

Reviewed against `.agents/skills/code-review.md`. Rendered the chart on this branch to check the claims, and read OpenHands/runtime-api#731 to check what the flag actually does.

**Verified**

- `helm template charts/openhands`: `WARM_RUNTIME_CONFIG_OVERLAY: "1"` renders on the runtime-api Deployment and on the `warm-runtimes` CronJob pod template (both pull from `runtime-api.env`), so the API and the reconciler resolve configs identically. ✅
- The value stays overridable: `--set runtime-api.env.WARM_RUNTIME_CONFIG_OVERLAY=0` wins, because `_env.yaml` ranges over `.Values.env` rather than appending a hardcoded entry. Consumers can turn it off without forking the chart (principle 6). ✅
- With `warmRuntimes.enabled: false` (the subchart default, and how SaaS runs) no ConfigMap is mounted, so overlay resolution degrades to "DB rows only" — identical to legacy. The flag is a no-op there. ✅
- Placement in `env` follows the existing pattern for app-owned runtime-api vars (`DEV_MODE`, `INCLUDE_START_PROBE`, …); no new bespoke toggle invented, which is the right call (principle 1).
- No chart `version` bump, correct for this repo — release-please derives it from the PR title, and `feat:` moves the minor.

Three things worth a response before merge.

### 1. Upgrade behaviour for operators who already saved a differently-named config

The description covers the same-name case (DB overrides ConfigMap). The case that changes materially is the **different-name** one. Today, an OHE operator who saved any config through `PUT /api/admin/warm-runtime-configs/{name}` — say `custom-image` — has the ConfigMap ignored entirely, so they run exactly one pool. After this flag flips on, they run **two**: their `custom-image` pool plus the resurrected installer-managed entry, each with its own `count`.

That is the intended "compose instead of fight" semantics, and for most people it's the fix. But it is a silent increase in warm pods (and their PVCs/CPU/memory) on a cluster the operator sized for one pool, and per runtime-api#731 they cannot delete the ConfigMap entry — the only escape is saving a same-named DB row with `count: 0`, which requires knowing the installer's entry name (`default` today, `v1_current` after #1181).

Not asking you to change the semantics. Two suggestions:

- Extend the values comment to say what happens to configs that *don't* share a name, since that's the surprising half and the comment is the only documentation that travels with the key into a consumer's values file.
- Make sure this lands in the OHE upgrade notes rather than only in the PR body — this is exactly the class of change a self-hosted operator otherwise finds out about from a resource graph.

### 2. The flag activates on a PR nobody will review with this context

`runtime-api.image.tag` is pinned at `0.9.0` and runtime-api#731 is still open, so this is genuinely inert on merge — that part checks out. But the moment it stops being inert is when the automated `bump-image-tag` workflow opens a `feat(runtime-api): …` tag-bump PR. That PR is a one-line tag change; the behaviour change in §1 ships with it, invisible to whoever approves it.

Worth considering landing this *after* the runtime-api pin crosses the release containing #731, so activation is a deliberate, reviewable step. If you'd rather land it now (reasonable — it really is inert), please note the minimum runtime-api version in the values comment so the tag-bump reviewer has a chance of connecting the two.

### 3. `warmRuntimes.count` semantics under overlay

Minor, and probably fine — flagging it so it's a conscious decision. The reconciler falls back to the file's top-level `count` for any entry without its own (`config.get('count', runtime_configs.get('count', DEFAULT_WARM_RUNTIME_COUNT))`), and #731 preserves that top-level `count` whenever file entries contribute. Since `WarmRuntimeConfig.to_dict()` omits `count` when it's NULL, a DB row saved without an explicit count now inherits the **chart's** `warmRuntimes.count` (1 in `charts/openhands/values.yaml`, the `sandbox_warm_runtime_count` ConfigOption on Replicated) instead of `DEFAULT_WARM_RUNTIME_COUNT` (3). That's arguably better — the operator's configured value beating a library constant — but it does mean an existing DB config's effective pool size can change on upgrade without anyone touching it. Intentional?

---

Nothing here is a blocker: the change is one line, follows the chart's env pattern, stays override-friendly, and is correctly scoped as a chart default that serves the self-hosted user (principle 7) rather than an environment-specific override. Mostly asking that §1 and §2 land somewhere an operator will actually read.

aivong-openhands (commented):
## Code review

Reviewed against `.agents/skills/code-review.md`. Cloned the branch, rendered `charts/openhands/charts/runtime-api` with and without `warmRuntimes.enabled`, ran `helm unittest` (14 passed), and read OpenHands/runtime-api#731 to confirm what the flag gates.

There is already a thorough review from @aivong-openhands on this PR. I won't re-litigate its §1 (different-name upgrade behaviour), §2 (activation lands via an unrelated tag-bump PR), or §3 (`count` inheritance) — I independently reached the same conclusions and those questions are still worth an answer. Below are three things that review didn't cover.

### Verified independently

- `WARM_RUNTIME_CONFIG_OVERLAY: "1"` renders on the runtime-api Deployment and the `warm-runtimes` CronJob pod template. ✅
- The value stays overridable: `--set env.WARM_RUNTIME_CONFIG_OVERLAY=0` renders `value: "0"`, and #731's parser (`in ('1','true','yes')`) reads that as off. `_env.yaml` ranges over `.Values.env` rather than appending a hardcoded entry, so no fork is needed to disable it (principle 6). ✅
- Placement in `env` alongside `DEV_MODE` / `INCLUDE_START_PROBE` follows the existing app-owned-var pattern; no bespoke toggle invented (principle 1). ✅
- `replicated/openhands.yaml` doesn't pin this key in its `runtime-api.env` block, so OHE inherits the chart default rather than carrying an environment-specific override — the right side of principle 7. ✅

---

### 1. The values comment's only pointer is unresolvable for the audience it's written for

```yaml
# Database warm-runtime configs overlay the installer-managed ConfigMap by
# name instead of replacing it (runtime-api#731). Inert on older images.
```

`OpenHands/OpenHands-Cloud` is public; `OpenHands/runtime-api` is private (404 unauthenticated). A self-hosted operator reading this `values.yaml` — the exact person principle 2 says must be able to predict a key's effect without internal context — cannot open `runtime-api#731`. The comment defers its substance to a link only we can follow.

"Inert on older images" has the same problem in a more actionable way: it tells the operator a version floor exists but not what it is. That's the one fact they'd need to decide whether the key does anything in their install.

Suggestion — keep the issue reference for us, but make the comment stand on its own:

```yaml
env:
  # Warm-runtime configs saved via the admin API overlay the installer-managed
  # ConfigMap entries by name instead of replacing the whole set: a saved config
  # with the same name wins, differently-named ones run alongside the installer's.
  # Requires runtime-api >= <VERSION>; ignored by older images.
  WARM_RUNTIME_CONFIG_OVERLAY: "1"
```

The "differently-named ones run alongside" clause is what @aivong-openhands asked for in §1, and it's the half an operator will actually be surprised by. Filling in `<VERSION>` also gives the future tag-bump reviewer the thread to pull, which is §2's ask — without either of them depending on a private link.

### 2. The flag lands on eight workloads, not two

The description says the shared helper feeds "the runtime-api Deployment ... and the warm-runtimes reconciler CronJob." `runtime-api.env` is included by six templates, and `cleanup-cronjob.yaml` ranges over phases, so a default render with `warmRuntimes.enabled=true` puts the var on:

```
1 × deployment.yaml
1 × warm-runtimes-cronjob.yaml
4 × cleanup-cronjob.yaml   (one CronJob per phase)
1 × db-cleanup-cronjob.yaml
1 × pod-status-logger-cronjob.yaml
```

This is harmless — the cleanup and pod-status-logger entrypoints never resolve warm configs, so the var is dead weight there — and factoring it out per-workload would cost more than the noise it saves. No change requested. But the description reads as though placement in `env` was scoped to the two consumers that care, and it wasn't; worth stating plainly so a future reader doesn't infer a targeting mechanism that doesn't exist.

### 3. The test asserts the flag in the one configuration where it can't do anything

The deployment case doesn't `set: warmRuntimes.enabled: true`, so it renders against the subchart default (`warmRuntimes.enabled: false`). In that state no ConfigMap is mounted at `/config`, overlay resolution has no file to overlay, and behaviour is identical to legacy — the flag is provably inert. The CronJob case does set the toggle, but that template only exists when it's enabled, so it had no choice.

So the suite covers "flag present when it does nothing" and never renders the Deployment in the configuration the PR exists to change. Adding `set: warmRuntimes.enabled: true` to the first test is a one-line change that makes it exercise the shipped OHE topology (`replicated/openhands.yaml` sets `warmRuntimes.enabled: true`) instead of the default-off one.

Given that overridability is the entire escape hatch for the §1 upgrade surprise, a third case asserting `env.WARM_RUNTIME_CONFIG_OVERLAY: "0"` survives to the rendered pod would be worth more than either existing case — it's the property most likely to silently regress if someone later moves this out of `.Values.env` into a hardcoded `_env.yaml` entry.

---

Not blocking on the chart mechanics: one line, correct pattern, override-friendly, and correctly scoped as a default that serves the self-hosted user rather than an environment override. §2 and §3 are cheap. §1 is the one I'd like resolved before merge, because the values comment is the only documentation that travels with the key into a consumer's values file, and right now it points somewhere they can't go.

aivong-openhands (commented):
## Code review

Reviewed against `.agents/skills/code-review.md`. Cloned `1f9c74b`, installed helm + helm-unittest, rendered the runtime-api subchart, ran the chart and script suites, and re-read OpenHands/runtime-api#731 to check what the flag actually gates.

Two prior reviews from @aivong-openhands are on this PR. This round is scoped to what changed in `1f9c74b` plus what is still outstanding.

### Verified on this revision

- `helm unittest charts/openhands/charts/runtime-api` → 4 suites, **15 passed**. `helm unittest charts/openhands` → 24 suites, **111 passed**. Both match the PR body. ✅
- `pytest scripts/test_replicated_warm_runtime_overlay.py` → 1 passed. ✅
- Rendering the subchart with `warmRuntimes.enabled=true` puts `WARM_RUNTIME_CONFIG_OVERLAY` on the Deployment and the `warm-runtimes` CronJob, so the API and the reconciler resolve configs identically. ✅
- `charts/openhands/charts/runtime-api/values.yaml` and `charts/openhands/values.yaml` both leave `runtime-api.env.WARM_RUNTIME_CONFIG_OVERLAY` unset, so SaaS and Helm-direct installs are genuinely untouched. ✅
- The KOTS block is a plain `env` key in the base `values`, and every `optionalValues` block that touches `runtime-api.env` uses `recursiveMerge: true`, so nothing clobbers it. ✅
- The previous review's §1 ask landed: the shipped comment now states both the same-name override and the different-name composition halves without leaning on the private `runtime-api#731` link. ✅
- The previous review's §3 ask landed: the API test now renders with `warmRuntimes.enabled: true`, i.e. the topology the PR exists to change. ✅

Three things.

### 1. The opt-out test does not actually guard the property it was added for

This is the case the last review specifically asked for — "the property most likely to silently regress if someone later moves this out of `.Values.env` into a hardcoded `_env.yaml` entry." As written it does not catch that regression.

I simulated exactly that regression by appending a hardcoded entry to the `runtime-api.env` helper:

```yaml
- name: WARM_RUNTIME_CONFIG_OVERLAY
  value: "1"
```

`helm unittest charts/openhands/charts/runtime-api` → **15 passed**, unchanged. The rendered Deployment under `--set env.WARM_RUNTIME_CONFIG_OVERLAY=0` is:

```yaml
- name: WARM_RUNTIME_CONFIG_OVERLAY
  value: "0"
- name: WARM_RUNTIME_CONFIG_OVERLAY
  value: "1"
```

Kubernetes takes the **last** entry, so the operator's opt-out is silently ignored — and `contains:` still passes, because the `"0"` entry is present, just outranked.

`contains` asserts presence, not that the value is the effective one. A `notContains` on the opposing value turns it into a real guard:

```yaml
  - it: preserves an explicit overlay opt-out
    template: templates/deployment.yaml
    set:
      warmRuntimes.enabled: true
      env.WARM_RUNTIME_CONFIG_OVERLAY: "0"
    asserts:
      - contains:
          path: spec.template.spec.containers[0].env
          content:
            name: WARM_RUNTIME_CONFIG_OVERLAY
            value: "0"
      - notContains:
          path: spec.template.spec.containers[0].env
          content:
            name: WARM_RUNTIME_CONFIG_OVERLAY
            value: "1"
```

The same applies to the two positive cases, but it matters most here: override-friendliness (principle 6) is the entire escape hatch for the upgrade surprise raised in the first review's §1, and right now nothing in CI defends it.

### 2. This encodes a dependency on a PR that has not merged

OpenHands/runtime-api#731 is still **open**. `charts/openhands/charts/runtime-api/values.yaml` pins `image.tag: 0.9.0`, so the flag is inert today — that part of the description checks out. But the shipped comment says only:

```
# Older runtime-api images ignore this variable.
```

which tells an operator a version floor exists without saying what it is, because no such version exists yet. Two consequences:

- If #731 lands with a different variable name or different semantics, this key becomes permanently dead config that looks live. Nothing in CI or the scope-guard test would notice — the failure mode of a misspelled env var is silence.
- The moment it stops being inert is a `feat(runtime-api): …` tag-bump PR, which is a one-line diff. The behaviour change ships with it, invisible to whoever approves it. `scripts/test_replicated_warm_runtime_overlay.py` pins activation but not a minimum image version, so there is no tripwire at the bump.

Cleanest sequencing is to land this after #731 merges and fill the version into the comment. If you'd rather land now, consider having the scope guard also assert the pinned `runtime-api` tag, so the tag bump has to acknowledge this key rather than sail past it.

### 3. Scoping to Replicated splits behaviour across two self-hosted paths — intentional?

Principle 7 says defaults should serve the **self-hosted user**, and that environment-specific needs belong in our own environment's overrides. `replicated/openhands.yaml` is not our environment — it is the OHE installer, i.e. self-hosted users. So this move puts one class of self-hosted user on overlay semantics and leaves Helm-direct self-hosted users on the legacy all-or-nothing behaviour that #731 describes as the bug (the first admin-API write hides the installer-managed pool and freezes image/env snapshots across upgrades).

That is defensible — OHE is the install shape you validated on R02, and staging activation is a reasonable way to de-risk. But it is the opposite of the direction the first review read principle 7 as pointing, and it's worth being explicit about which it is:

- deliberate staging, with the chart default following once #731 ships and bakes; or
- a permanent split, in which case Helm-direct users keep a known-buggy resolution path with no comment anywhere in `values.yaml` telling them the knob exists.

Related and cheap: the scope guard asserts `charts/openhands/charts/runtime-api/values.yaml` has no `WARM_RUNTIME_CONFIG_OVERLAY`, but not `charts/openhands/values.yaml`. The umbrella is the layer SaaS actually renders and it carries its own `runtime-api.env` block (13 keys today), so a stray key there would defeat the guard's intent while it still passes. One extra assert closes it.

### Still unanswered from the first review

§3 of the first review — `count` inheritance — has not been addressed in the description or the code. Under overlay, #731 preserves the file's top-level `count`, and `WarmRuntimeConfig.to_dict()` omits `count` when NULL, so a DB row saved without an explicit count starts inheriting `sandbox_warm_runtime_count` (default `"1"` per `replicated/config.yaml`) instead of `DEFAULT_WARM_RUNTIME_COUNT` (3). An existing operator's pool can shrink on upgrade without anyone touching the config. Probably the better behaviour, but it should be a stated decision rather than a side effect, and it belongs in the OHE upgrade notes alongside the different-name composition change.

### Not blocking

- The var renders on 8 workloads, not 2 (`deployment`, `warm-runtimes`, 4× `cleanup` phases, `db-cleanup`, `pod-status-logger`), because `runtime-api.env` is shared. Dead weight on the six that never resolve warm configs, and factoring it out would cost more than it saves. No change requested; just don't let the description imply a targeting mechanism that doesn't exist.
- No chart version bump, correct here — release-please derives it from the PR title.
- The PR body is honest about the browser smoke stopping at the Bitbucket auth redirect rather than rounding it up to PASS. Appreciated.

---

The mechanics are right: correct value shape, correct merge semantics, override-friendly, no bespoke toggle invented, and the R02 validation covers the four behaviours that matter. **§1 is the one I'd fix before merge** — it is a two-line test change, and without it the opt-out path is untested despite being listed as covered. §2 and §3 are questions about sequencing and intent rather than defects.

aivong-openhands (commented):
## Code review

Reviewed against `.agents/skills/code-review.md`. Cloned `18b2c5f`, installed helm + helm-unittest, rendered the subchart, ran the chart suites, and cloned `OpenHands/runtime-api` at #731 to execute its resolver against this chart's rendered ConfigMap rather than reasoning about it from the diff.

Three prior reviews from @aivong-openhands are on this PR. This round is scoped to what changed since `1f9c74b`, plus one finding none of them quantified.

### Verified on this revision

- `helm unittest charts/openhands/charts/runtime-api` → 5 suites, **18 passed**. `helm unittest charts/openhands` → 25 suites, **114 passed**. (Both higher than the PR body's 15/111 — main moved under the branch; not a discrepancy in this PR.) ✅
- `pytest scripts/test_replicated_warm_runtime_overlay.py` → 1 passed on the branch, and **fails with `KeyError: 'WARM_RUNTIME_CONFIG_OVERLAY'` against `main`** — so it is a real guard, not a tautology. ✅
- The previous review's §1 ask landed, and I confirmed it actually bites. I re-ran the exact regression it described — appending a hardcoded `WARM_RUNTIME_CONFIG_OVERLAY: "1"` to the `runtime-api.env` helper — and the suite now **fails** on `asserts[1] notContains`. I also tried a subtler variant (hardcode the entry *and* skip the key in the `.Values.env` range, so the operator's value never renders at all); that fails too. The opt-out path is genuinely defended now. ✅
- Fed the chart's rendered ConfigMap through #731's `EffectiveWarmRuntimeConfig.model_validate` — the validation gate overlay mode adds and legacy never had — and the default entry validates clean. The chart's output is compatible with the flag it turns on. ✅
- Every `optionalValues` block touching `runtime-api.env` uses `recursiveMerge: true` (indices 2, 17, 22), so nothing clobbers the new key. ✅

One finding worth a decision, then two small ones.

---

### 1. On the shipped `sandbox_warm_runtime_count`, activation *shrinks* the warm pool — possibly to zero

The first review's §3 raised `count` inheritance as a question. Executing #731's resolver together with the reconciler's `desired_count` arithmetic against this repo's actual defaults, the common case goes the **opposite** direction from the "silent increase in warm pods" the earlier reviews anticipated.

`SaveWarmRuntimeConfigRequest.count` is `Optional[int] = None` and `to_dict()` omits it when NULL, so an operator who saved a config through the admin API without explicitly setting a count has no per-entry count. Legacy resolution returns `{'configs': [...]}` with no top-level `count`, so the reconciler falls back to `DEFAULT_WARM_RUNTIME_COUNT` (3). Overlay preserves the *file's* top-level `count`, which on Replicated is `sandbox_warm_runtime_count` — **default `"1"`** per `replicated/config.yaml`.

Total warm pods for an operator with one admin-API config saved without an explicit count:

| `sandbox_warm_runtime_count` | before | after |
|---|---|---|
| 0 | 3 | **0** |
| 1 (the shipped default) | 3 | 2 |
| 3 | 3 | 6 |
| 5 | 3 | 10 |

The `0` row is the one I'd want a decision on. `sandbox_warm_runtime_count: 0` is a documented, validated setting whose help text reads "Set to 0 to disable and cold-start each sandbox." An operator who set it to 0 and *then* used the admin API to stand up a custom pool gets 3 warm pods today; after activation they get **zero**, and every conversation cold-starts. Their explicitly-configured pool silently stops existing, and nothing surfaces it — `GET /api/warm-runtime-configs` reports the config with `count` absent either way.

At `1`, the shipped default, they go from 3 pods of their own image to 1 of theirs plus 1 installer-managed. Arguably the better shape, but still a capacity change on a cluster sized for the old one.

Operators who *did* pass an explicit count are unaffected on their own entry (4 → 4 even at `sandbox_warm_runtime_count: 0`) and only pick up the installer entry — the intended composition.

This is a property of #731 rather than of this one-line diff, so I'm not asking you to change the chart. But this PR is what makes it real for OHE, and "your warm pool can go to zero" is materially different from the "extra pools appear" framing in the PR body and the values comment. Worth either recording as an accepted trade-off or fixing upstream — the cleanest fix is probably for #731 to fall back to `DEFAULT_WARM_RUNTIME_COUNT` for DB entries with no count, instead of inheriting the file's top-level one, so an admin-API config's size never depends on an installer knob the operator never associated with it. That also keeps "one knob, one effect."

### 2. The operator-facing note has a concrete home in this repo

Two prior reviews asked that the composition change land somewhere an operator actually reads, without naming a destination. There is one: `replicated/application.yaml` carries a `spec.releaseNotes` block, and it gets edited in exactly this class of PR (`feat: wire Agent Canvas through Replicated/Helm installs`, `feat(sysbox): default sandbox isolation on the embedded cluster`). "If you saved warm-runtime configs through the admin API, your pool composition **and size** change on this upgrade" belongs there. That closes both the earlier ask and §1 in one place.

### 3. `notContains` guards the Deployment but not the CronJob

`warm_runtime_overlay_env_test.yaml` now pairs `contains` with `notContains` on the opt-out case for `deployment.yaml`. The reconciler CronJob case is still `contains`-only — and the reconciler is the workload that actually creates pods. An opt-out that survives on the Deployment but is outranked on the CronJob would leave the list endpoint and the reconciler disagreeing about what's live, which is precisely the split this PR exists to prevent. Same two lines as the fix you already applied.

Related and minor: overlay mode silently drops ConfigMap entries that fail `EffectiveWarmRuntimeConfig` validation, where legacy passed them straight through. I confirmed the chart can render such an entry — a `warmRuntimes.configsByName.<name>` without a `command` renders `"command": null`, which validation rejects. Safer than legacy's behaviour, but silent (a `logger.warning` only), so a typo'd entry now vanishes instead of failing loudly. No change requested; the chart's own defaults are fine.

### Still open from earlier rounds

runtime-api#731 has not merged, so the version floor in the comment still can't be filled in, and activation still arrives via a one-line `feat(runtime-api):` tag-bump PR. I won't re-argue it — the earlier reviews covered it well and the sequencing call is yours.

### Not blocking

- No chart `version` bump — correct here, release-please derives it from the PR title.
- The var lands on 8 workloads via the shared `runtime-api.env` helper, not the 2 the description implies. Harmless; already noted upstream.
- The PR body is honest about the browser smoke stopping at the Bitbucket auth redirect instead of rounding it up to PASS. Appreciated.

---

The chart mechanics are right and I'd have approved on them alone: one line, existing env pattern, override-friendly, no bespoke toggle invented, and the opt-out is now genuinely tested. **§1 is why I'm commenting rather than approving** — not because the diff is wrong, but because on the shipped default (`sandbox_warm_runtime_count: 1`, and especially `0`) activation *reduces* an existing operator's warm capacity, the opposite of what the PR body and every prior review anticipated. I'd like that to be a stated decision before it ships. §2 and §3 are cheap.

aivong-openhands (commented):
## Code review — update following the runtime-api#731 merge

runtime-api#731 merged as `779a959` at 22:48 UTC. I re-cloned `OpenHands/runtime-api` at `main` and re-ran my earlier analysis against the **merged** code rather than the PR branch. Updating my previous review accordingly.

### What the merge resolves

The second review's §2 raised a real risk: that #731 could land with a different variable name or different semantics, leaving this key as "permanently dead config that looks live," with the failure mode of a misspelled env var being silence. **That risk is now closed:**

- The variable name is unchanged — `WARM_RUNTIME_CONFIG_OVERLAY` in both `warm_runtimes.overlay_enabled()` and `management_fastapi`. It matches this PR's key exactly. ✅
- The parser is unchanged: `os.environ.get(...).strip().lower() in ('1','true','yes')`, so `"1"` activates and the documented `"0"` opt-out reads as off. ✅
- `git diff 779a959..origin/main -- warm_runtimes.py` is empty — nothing landed on top. The resolver I tested last round is the resolver that shipped. ✅

So the key is live config, not dead config. That was the strongest argument for deferring this PR, and it no longer applies.

### The version floor is now knowable — please fill it in

Two prior reviews asked for a minimum runtime-api version in the comment, and it couldn't be answered because no such version existed. It does now:

- `v0.9.0` (current `Latest`, tagged `cabaa1e`) **predates** #731 — `git tag --contains 779a959` is empty, and `warm_runtimes.py` at `v0.9.0` contains zero occurrences of the variable.
- #731 is in the open release-please PR OpenHands/runtime-api#732, `chore(main): release 0.10.0` (branch head `582de25`, confirmed `779a959` is an ancestor).

**The floor is runtime-api `0.10.0`**, which is merged but **not yet released**. The chart still pins `image.tag: 0.9.0`, so the flag remains genuinely inert on merge — the PR body's rollout claim still holds. Suggest replacing "Older runtime-api images ignore this variable" with the concrete floor:

```yaml
# Admin-API warm configs overlay installer-managed entries by name:
# same-named rows override them and differently-named rows run beside
# them. Requires runtime-api >= 0.10.0; older images ignore this variable.
```

That is the one fact an operator needs to know whether the key does anything in their install, and it gives the future tag-bump reviewer the thread to pull.

### The tripwire concern is now imminent, not hypothetical

Earlier reviews noted that activation would arrive via an automated `feat(runtime-api): bump image tag to <tag>` PR — a one-line diff carrying this behaviour change invisibly. With #731 merged and 0.10.0 queued, that bump is the **next** runtime-api release, not a distant one. There are no open tag-bump PRs against this repo right now, so the window to add a tripwire is before 0.10.0 publishes.

The earlier suggestion still stands and is now cheap and concrete: have `scripts/test_replicated_warm_runtime_overlay.py` also assert the pinned `runtime-api` `image.tag`, so the bump has to acknowledge this key rather than sail past it. With a known floor of `0.10.0` you can now write that assertion meaningfully.

### My §1 stands — re-verified against merged code

I re-ran the resolver plus the reconciler's `desired_count` arithmetic using merged `main`, not the PR branch. Identical results:

| `sandbox_warm_runtime_count` | total warm pods before | after |
|---|---|---|
| 0 | 3 | **0** |
| 1 (the shipped default) | 3 | 2 |
| 3 | 3 | 6 |
| 5 | 3 | 10 |

For an OHE operator with one admin-API config saved without an explicit `count` (`SaveWarmRuntimeConfigRequest.count` is `Optional[int] = None`, and `to_dict()` omits it when NULL): legacy falls back to `DEFAULT_WARM_RUNTIME_COUNT` (3); overlay inherits the file's top-level `count`, which on Replicated is `sandbox_warm_runtime_count`, default `"1"`.

The merge makes this **more** urgent rather than less. It was a latent property of an unmerged PR; it is now the behaviour of shipped code that this chart change activates, one release away. In particular an operator who set `sandbox_warm_runtime_count: 0` — a documented, validated setting ("Set to 0 to disable and cold-start each sandbox") — and then built a custom pool through the admin API drops from 3 warm pods to **zero**, silently.

If the preferred fix is upstream (fall back to `DEFAULT_WARM_RUNTIME_COUNT` for DB entries with no count, rather than inheriting an installer knob the operator never associated with their config), that is now a follow-up PR against merged code and can proceed in parallel — it does not have to block this one. What I'd still like before this ships is for the outcome to be a stated decision rather than a side effect, and captured in `replicated/application.yaml`'s `releaseNotes` where an operator will see it.

### Unchanged from my previous review

- §2 — the operator-facing note belongs in `replicated/application.yaml` `spec.releaseNotes`, which is edited in exactly this class of PR. Now doubly worth it: that note can cite the concrete `0.10.0` floor.
- §3 — the reconciler CronJob opt-out case is still `contains`-only while the Deployment case has the `notContains` guard. Two lines.

### Net

The merge removes the "dead config" risk and unblocks the sequencing question that dominated earlier rounds — that's a real improvement to this PR's position, and I'd now be comfortable with it landing before 0.10.0 publishes. Remaining asks are unchanged and all cheap: fill in the `0.10.0` floor in the comment, mirror the `notContains` guard onto the CronJob, and record the pool-size change (§1) in the release notes. Still commenting rather than approving on §1 — not on the diff's mechanics, which are right, but because activation reducing an operator's warm capacity to zero on a documented setting should be an explicit decision.

aivong-openhands (approved):

PR #1184, opened 2026-08-31
fix: advertise the runtime API ingress to fuse mounts

ak684:
## Summary

Set OHE's `RUNTIME_API_BASE_URL` to `computed_runtime_api_hostname`, the hostname actually rendered on `openhands-runtime-api-ingress`.

The current value uses `computed_runtime_base_hostname`, which is the wildcard base for sandbox runtimes. runtime-api uses `RUNTIME_API_BASE_URL` only to construct fuse storage-broker callback URLs, so those callbacks are sent to the sandbox ingress instead of the runtime API.

## History

- runtime-api#619 introduced fuse/S3 warm pools and defined this variable as the URL of the Runtime API as seen from runtime pods.
- OpenHands-Cloud#926 added the OHE value on 2026-07-20, but assumed the derived API URL was `https://runtime.<base>` and wired it to the sandbox base from its first commit.
- OpenHands-Cloud#985 added `computed_runtime_api_hostname` for the actual API ingress on 2026-07-31, but mechanically preserved the old `RUNTIME_API_BASE_URL`/runtime-base pairing.

This is therefore a pre-existing fuse integration bug, independent of config-overlay activation in #1183, and can ship or be backported independently.

## Live evidence

On a fresh R02 Replicated install:

- `RUNTIME_API_BASE_URL=https://runtime.replicated-02...`
- the runtime API ingress host was `runtime-api.replicated-02...`
- requests to the configured host returned 404
- the same runtime API route on the ingress host succeeded

PVC-backed warm pools do not use this callback, which is why the main Gap 2 overlay validation still passed. Fuse-backed pools would fail when attempting to use the broker URL.

## Risk

The variable has one functional consumer in runtime-api: fuse broker URL construction during a fuse-backed claim. It is not used for PVC-backed pools, overlay resolution, admin auth, sandbox ingress naming, or the normal API/list path. All supported OHE hostname modes already put `computed_runtime_api_hostname` on the runtime-api Ingress and include it in proxy/TLS wiring.

The Deployment rolls once because an env value changes. The only plausible compatibility edge is an unsupported hand-patched install that routes the sandbox hostname to runtime-api and relies on the incorrect value; the shipped templates do not do that.

## Testing

- Added a Replicated DNS-contract regression test requiring `RUNTIME_API_BASE_URL` to use `computed_runtime_api_hostname`.
- Targeted DNS tests: 7 passed.
- Full script suite: 479 passed.
- Umbrella Helm unit tests: 108 passed.

### DIFF
--- a/replicated/openhands.yaml
+++ b/replicated/openhands.yaml
@@ -334,7 +334,7 @@ spec:
         # it; "." nests it under the runtime base. The path-routing block below
         # overrides this to "/", which the runtime-api reads as path mode.
         RUNTIME_URL_SEPARATOR: '{{repl ConfigOption "computed_runtime_hostname_separator" }}'
-        RUNTIME_API_BASE_URL: 'https://{{repl ConfigOption "computed_runtime_base_hostname" }}'
+        RUNTIME_API_BASE_URL: 'https://{{repl ConfigOption "computed_runtime_api_hostname" }}'
         RUNTIME_IMAGE_PULL_SECRETS: '{{repl if and (ConfigOptionEquals "custom_sandbox_image_enabled" "1") (ne (ConfigOption "custom_sandbox_image_registry_server") "") (ne (ConfigOption "custom_sandbox_image_registry_username") "") (ne (ConfigOption "custom_sandbox_image_registry_password") "") }}sandbox-image-pull-secret{{repl else}}{{repl ImagePullSecretName }}{{repl end}}'
         STORAGE_CLASS: "openebs-hostpath"
         INGRESS_CLASS: "traefik"
--- a/scripts/test_replicated_dns_layout.py
+++ b/scripts/test_replicated_dns_layout.py
@@ -20,6 +20,7 @@ REPLICATED_CONFIG = REPO_ROOT / "replicated" / "config.yaml"
 
 SEPARATOR_OPTION = '{{repl ConfigOption "computed_runtime_hostname_separator" }}'
 BASE_OPTION = '{{repl ConfigOption "computed_runtime_base_hostname" }}'
+API_HOST_OPTION = '{{repl ConfigOption "computed_runtime_api_hostname" }}'
 
 
 @pytest.fixture(scope="module")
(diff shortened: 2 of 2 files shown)


aivong-openhands (approved):
## Review: `fix: advertise the runtime API ingress to fuse mounts`

Verified the diagnosis end-to-end rather than taking the PR body at face value.

**The fix is correct.** In `OpenHands/runtime-api`, `RUNTIME_API_BASE_URL` has exactly one functional consumer — `runtimes.py:1642`:

```python
broker_url = f'{RUNTIME_API_BASE_URL}/runtimes/{workspace_runtime_id}/storage'
```

`utils/constants.py:215` documents it as "URL at which the runtime-API is reachable... so pods can call back to the broker." It is read nowhere else (the only other hits are the import and the empty-value guard). So this is unambiguously the runtime-api's own address, not the sandbox base — and `replicated/openhands.yaml:304` already puts `computed_runtime_api_hostname` on `openhands-runtime-api-ingress`. Before this change the callback targeted the sandbox wildcard host, which has no route for `/runtimes/{id}/storage`, hence the 404 in your R02 evidence.

**Holds across every supported layout.** I checked rather than assuming:

- `computed_runtime_api_hostname` (`config.yaml:152`) resolves non-empty in all three `hostname_mode` branches — custom, wildcard, and the legacy fallback.
- The runtime-api Ingress is unconditional, so the host is always actually served.
- Path-routing mode (`openhands.yaml:1036`) flips `ingressBase.enabled: false` and the sandbox Gateway, but leaves `ingress` alone — so the new value is right there too. Under the old value, path mode was arguably worse, since the base host there is a *shared* sandbox host.
- The host is in the computed `NO_PROXY` list, and it is in the cert-SAN instructions at `config.yaml:292`, so the sandbox pod making the callback can both bypass the proxy and validate TLS against the uploaded `openhands-tls`. Both were already true of the old value, so there is no proxy/TLS regression either way.

**Risk is even lower than the PR states.** Nothing in `replicated/` sets `FUSEY_*` env or `fuse_s3_mount: true` on a `warmRuntimes` entry, so no shipped Replicated install currently reaches this code path at all — the guard at `runtimes.py:1638` never even gets a chance to matter today. This is a latent-bug fix that pre-positions correctly for fuse-backed pools, which supports shipping it independently of #1183 as you propose.

**Test.** Matches the surrounding contract-test style in `test_replicated_dns_layout.py`, and the name describes the behavior (`advertises_its_own_ingress_for_storage_callbacks`) rather than the mechanism. Ran the file locally: 7 passed.

---

### One question, not a blocker

The callback is made by a sandbox pod to a service in the same cluster, but it is now addressed via the public ingress hostname — so the request hairpins out to the load balancer and back. `NO_PROXY` already carries `openhands-runtime-api` as an in-cluster name, so the service-DNS form is clearly available.

Was routing via the ingress a deliberate choice? I can see a good reason for it — sandbox pods can live in a separate `sandbox_namespace`, so a service URL would need an FQDN, and the ingress host is the one name guaranteed correct in every namespace/isolation combination. If that is the reasoning, it is the right call and worth nothing more than a note. I would only push back if the hairpin turns out to be a problem in airgapped installs where the LB path is constrained, and that can be handled later if it ever surfaces.

### Not raising

The one-line env change rolls the Deployment. That is inherent and correctly called out in your Risk section.

---

**Approving.** Single-line correction to an explicit, consumer-visible value, with a regression test that pins the contract — exactly the "explicit over implicit" direction the chart guidelines ask for.

PR #1185, opened 2026-08-31
fix: OHE-3186 : render reaper archive volume under a volumes: key (staging reaper outage)

tofarr:
## Summary

The runtime-api `cleanup-reaper` CronJob has been silently failing on **staging**: no reaper pods were appearing (while `k8s-garbage`, `snapshotter`, and `retention` ran fine), so idle runtimes stopped being reclaimed. This PR fixes the chart template bug that caused it.

## Root cause

In `charts/openhands/charts/runtime-api/templates/cleanup-cronjob.yaml`, the `volumes:` and `volumeMounts:` keys were gated behind `ingressBase.enabled`, but the reaper's `archive-tmp` volume/mount entries were appended under a **separate** `{{- if }}` that did **not** emit the key:

```go
{{- if $.Values.ingressBase.enabled }}      // emits "volumes:" key
volumes:
  - name: ingress-base-config
    ...
{{- end }}
{{- if and (eq $phase "reaper") $.Values.runtimeFileArchive.enabled }}   // emits ONLY the list item
  - name: archive-tmp
    emptyDir:
      sizeLimit: ...
{{- end }}
```

When `ingressBase.enabled` was **false** (the chart default) but `runtimeFileArchive.enabled` was **true** (staging's config), the bare `- name: archive-tmp` list item had no parent key and merged into the previous list — `imagePullSecrets` at the pod level and `env` at the container level — producing an **invalid Pod spec**:

```yaml
imagePullSecrets:
  - name: ghcr-login-secret
  - name: archive-tmp        # emptyDir volume illegally here
    emptyDir:
      sizeLimit: "10Gi"
# (no "volumes:" key exists)
```

…and the container `volumeMounts:` similarly merged into `env`:

```yaml
            - name: RUNTIME_FILE_ARCHIVE_TMPDIR
              value: "/archive-tmp"
            - name: archive-tmp          # a volumeMount, illegally inside env
              mountPath: "/archive-tmp"
```

Kubernetes rejects the spec, so the reaper CronJob could never start a Job. This is exactly the staging symptom observed: `k8s-garbage`, `snapshotter`, and `retention` pods completed, but no reaper pod ever appeared.

## Why production was unaffected

Production runs with `runtimeFileArchive.enabled: false` (chart default), so the buggy `{{- if }}` block was skipped entirely and the reaper rendered cleanly — which is why prod shows a healthy reaper (`...-helm-cleanup-reaper-...` pods, including one `Running`).

## The fix

Emit `volumes:` / `volumeMounts:` whenever **either** source (ingressBase or the reaper archive volume) is needed, and gate the individual entries. A `$archiveVol` local avoids recomputing the condition.

## Verification

- `helm lint charts/openhands` ✅
- `helm unittest charts/openhands/charts/runtime-api` — all 18 tests pass (4 suites), including a new regression suite ✅
- Confirmed the new regression tests **fail** against the pre-fix template (2/7 cases) and **pass** with the fix ✅
- Rendered the umbrella with staging-like values (`ingressBase.enabled=false`, `runtimeFileArchive.enabled=true`): the reaper now produces a valid spec with a proper `volumes:` key, `archive-tmp` emptyDir volume, `archive-tmp` volumeMount, and `RUNTIME_FILE_ARCHIVE_TMPDIR` env var — and no pollution of `imagePullSecrets`/`env` ✅
- Tested all 4 combinations of `(ingressBase.enabled, runtimeFileArchive.enabled)` for both the reaper and non-reaper phases ✅

## After merge

Once this lands and is cut into a new `openhands` chart release, saas-deploy's runtime-api wrappers (staging in particular) should be bumped to pick up the fix. Staging's reaper will start running again without any saas-deploy `values.yaml` change.

### DIFF
--- a/charts/openhands/charts/runtime-api/templates/cleanup-cronjob.yaml
+++ b/charts/openhands/charts/runtime-api/templates/cleanup-cronjob.yaml
@@ -3,6 +3,7 @@
 {{- $jobs := dict "reaper" .Values.cleanup.jobs.reaper "k8s-garbage" (index .Values.cleanup.jobs "k8s-garbage") "snapshotter" .Values.cleanup.jobs.snapshotter "retention" .Values.cleanup.jobs.retention }}
 {{- range $phase, $cfg := $jobs }}
 {{- if $cfg.enabled }}
+{{- $archiveVol := and (eq $phase "reaper") $.Values.runtimeFileArchive.enabled }}
 apiVersion: batch/v1
 kind: CronJob
 metadata:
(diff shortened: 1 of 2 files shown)


hieptl (approved):
Thank you! 🙏

PR #1186, opened 2026-09-01
fix(ci): bypass broken deploy-gate check temporarily

tofarr:
The unstable Replicated deploy on main has been failing intermittently with an empty KOTS deploy response, which causes the required last-deploy check on the main ruleset to block all PRs. Temporarily exit 0 so PRs can merge while the platform team investigates.

Tracking: PLTF-3535
Failing runs: 33497185018, 33452723202, 33428909029

## Helm Chart Checklist


simonrosenberg (approved):

hieptl (approved):
Thank you! 🙏

PR #1208, opened 2026-09-08
feat: diagnose runtime ingress failures in support bundles

ak684:
## Description

Add read-only runtime ingress diagnostics to OHE support bundles. For runtimes left in `starting` for more than 60 seconds, the collector runs from the Runtime API container and records:

- direct Kubernetes Service health
- runtime-hostname DNS resolution
- TLS verification, expiry, and DNS SANs
- the external ingress HTTP status and expected readiness value
- structured, credential-redacted errors for each failed layer

The analyzer reports unreachable stale runtimes instead of showing an all-green bundle. This also collects Traefik logs, which were missing while only ingress-nginx logs were requested.

The complementary Runtime API logging change is OpenHands/runtime-api#740, but this collector is self-contained and does not depend on that PR.

## Helm Chart Checklist

- [x] I have updated the chart's README.md if there are any breaking changes or new required values (none)

## Validation

- `helm unittest charts/openhands` — 133 passed
- embedded collector Python compilation — passed
- `helm lint charts/openhands` — passed

## Additional Notes

The collector selects only non-secret runtime identifiers and status timestamps. It does not emit database credentials, request headers, or response bodies. Proxy credentials in exception strings are redacted.


aivong-openhands (commented):
Found one issue that I think needs attention before approval:

- `runtime-api-ingress-diagnostics` only reads `networking.k8s.io` Ingress objects and treats `runtime-<id>` as an Ingress name. The Replicated embedded-cluster path-routing configuration this support-bundle work appears to target sets `USE_GATEWAY_API=true` / `sandboxGateway.enabled`, and runtime-api creates `gateway.networking.k8s.io` HTTPRoutes instead of Ingresses. In that common path, `ingresses.get(f"runtime-{row[0]}")` will always be missing, so the analyzer reports `runtime ingress not found` and fails even when the runtime route is healthy. The same URL construction also cannot safely append `/alive` to nginx regex paths like `/{runtime_id}/runtime(/|$)(.*)`. Please handle Gateway API/path routing explicitly, or gate this diagnostic to traditional Ingress mode, before relying on this check.

AI disclosure: This PR review was created by an AI agent (OpenHands) on behalf of the user.

aivong-openhands (approved):
🟢 **Taste Rating: Good taste**

The added support-bundle diagnostic is scoped, read-only, and follows the existing troubleshoot pattern by executing inside the runtime-api container and reusing its environment/source-of-truth URL helper. The Gateway API follow-up addresses the main routing compatibility concern for Replicated path routing, and the tests cover rendering, Gateway/Ingress support, Traefik log collection, and disabled runtime-api behavior.

**[RISK ASSESSMENT]**
- [Overall PR] ⚠️ Risk Assessment: 🟢 LOW

This is isolated to Helm support-bundle diagnostics and analyzer output. It does not change deployed runtime behavior, chart defaults, permissions, dependencies, or storage semantics. CI is green, including Helm unit tests and chart linting.

**VERDICT:**
✅ **Worth merging**: Core logic is sound and follows the repo-specific chart review guidelines.

**KEY INSIGHT:**
The diagnostic now compares the direct service path with the same external runtime URL shape runtime-api itself exposes, avoiding a parallel URL-construction path.

AI disclosure: This PR review was created by an AI agent (OpenHands) on behalf of the user.

PR #1211, opened 2026-09-08
chore: Bump agent server to 1.45.0

tofarr:
## Helm Chart Checklist


hieptl (approved):
Thank you! 🙏

PR #1213, opened 2026-09-09
feat: wire the git sync wrapping secret and a /workspace volume

hieptl:
## Description

The automation service now syncs automations to git per organization (OpenHands/automation, [OHE-3153](https://linear.app/all-hands-ai/issue/OHE-3153/git-sync-is-not-available-in-cloud-backends)). Two deployment gaps kept that from working in the cloud; this PR closes them ([OHE-3230](https://linear.app/all-hands-ai/issue/OHE-3230)).

- **Wrapping secret.** `charts/openhands/charts/automation/templates/_env.yaml` sets `AUTOMATION_GIT_SYNC_SECRET` from `.Values.gitSyncSecretFromSecret` (`automation-git-sync-secret` / `git-sync-secret`) with `optional: true`, so pods start before ops provisions the secret. Until it exists the service returns 503 from `PUT /v1/git-sync/config` naming the env var rather than falling back to a per-pod key file that the other replicas could not read. The value is added to the subchart `values.yaml` and mirrored in the umbrella `charts/openhands/values.yaml`.
- **Scratch volume.** `deployment.yaml` and `events-deployment.yaml` now always declare `volumes`/`volumeMounts` with an `emptyDir` named `workspace` mounted at `/workspace` (the service's `AUTOMATION_WORKSPACE_BASE` default, which the image does not create). The CA-bundle volume and mount stay conditional on `caBundle.enabled`. Pods keep no state across restarts; a fresh pod re-clones on its first cycle.
- **Secrets chart.** New `charts/openhands-secrets/templates/automation-git-sync-secret.yaml` rendering `automation-git-sync-secret` from `config.automation_git_sync_secret`, mirroring the webhook secret.

Verification:

```
helm lint charts/openhands/charts/automation
helm template t charts/openhands/charts/automation --set events.enabled=true   # env var with optional: true and /workspace emptyDir on both deployments
helm template t charts/openhands/charts/automation --set caBundle.enabled=true # workspace plus CA bundle volumes/mounts
helm unittest charts/openhands/charts/automation                               # 14 tests passed
helm template s charts/openhands-secrets --set automation.enabled=true --set config.automation_git_sync_secret=abc
```

## Helm Chart Checklist

## Additional Notes

- **Follow-ups for ops (not in this PR):** provision the SOPS-encrypted `automation-git-sync-secret` (key `git-sync-secret`) in `infra/k8s/production/secrets/openhands/` and `saas-deploy/secrets/staging/…` and add it to the secret-generator lists; add `gitSyncSecretFromSecret` to `saas-deploy/app/automation/environments/*/values.yaml`; add the option to the Replicated self-hosted config.
- **Ship order:** deploy together with the automation service change and before the Agent Canvas change ([OHE-3229](https://linear.app/all-hands-ai/issue/OHE-3229)) that unhides Git Sync on cloud backends.
- The events pods get the volume too because they run the same application and background loops as the api pods.
- Chart versions are managed by release-please; no manual bump.


charts/openhands/charts/automation/templates/deployment.yaml
```diff
@@ -31,12 +31,17 @@ spec:
       affinity:
         {{- . | nindent 8 }}
       {{- end }}
-      {{- if .Values.caBundle.enabled }}
       volumes:
+        # Scratch space for git sync checkouts (AUTOMATION_WORKSPACE_BASE
+        # defaults to /workspace, which the image does not create). Pods keep
+        # no state across restarts; a fresh pod re-clones on its first cycle.
+        - name: workspace
+          emptyDir: {}
```
aivong-openhands: **Storage**: this `emptyDir` has no `sizeLimit`, so it can grow until the node's ephemeral storage is exhausted and the kubelet evicts the pod — which drops the whole automation API, not just git sync. The container has `memory`/`cpu` limits but no `ephemeral-storage` request or limit, so nothing else bounds it either.

The precedent is one chart over: `runtime-api`'s cleanup CronJob caps its scratch volume at `runtimeFileArchive.tmpSizeLimit` (default `10Gi`). Same situation here, and arguably harder to bound by inspection — the content is clones of consumer-supplied repos, one directory per org (`{workspace_base}/git-sync/{org_id}`).

```yaml
- name: workspace
  emptyDir:
    sizeLimit: {{ .Values.workspace.sizeLimit | quote }}
```

with a `workspace.sizeLimit` default in `values.yaml`. A bounded `emptyDir` also fails legibly: writes get `ENOSPC` inside the pod instead of the node evicting you.

charts/openhands/charts/automation/templates/deployment.yaml
```diff
@@ -181,8 +186,10 @@ spec:
         - name: REQUESTS_CA_BUNDLE
           value: {{ .Values.caBundle.mountPath }}/{{ .Values.caBundle.key }}
         {{- end }}
-        {{- if .Values.caBundle.enabled }}
         volumeMounts:
+        - name: workspace
+          mountPath: /workspace
```
aivong-openhands: **Explicit over implicit**: `/workspace` is hardcoded here, but the directory the service actually writes to is `AUTOMATION_WORKSPACE_BASE`. This chart never sets that env var, and a consumer can set it through the `.Values.env` passthrough that `automation.env` supports by design. Set it and the mount silently stops covering the write path — clones land on the container filesystem, which is the exact failure this PR is fixing.

Make the chart own the value it depends on: add e.g. `workspace.mountPath: /workspace` to `values.yaml`, render `AUTOMATION_WORKSPACE_BASE` from it in `_env.yaml`, and use it for `mountPath` in both deployments. Then the mount and the write path cannot drift, and the "`AUTOMATION_WORKSPACE_BASE` defaults to /workspace" note in the comment above becomes something the templates enforce rather than an assumption about another repo.

charts/openhands/charts/automation/templates/events-deployment.yaml
```diff
@@ -36,12 +36,17 @@ spec:
       affinity:
         {{- . | nindent 8 }}
       {{- end }}
-      {{- if .Values.caBundle.enabled }}
       volumes:
+        # Scratch space for git sync checkouts (AUTOMATION_WORKSPACE_BASE
+        # defaults to /workspace, which the image does not create). Pods keep
+        # no state across restarts; a fresh pod re-clones on its first cycle.
+        - name: workspace
+          emptyDir: {}
```
aivong-openhands: Same unbounded `emptyDir` as `deployment.yaml` — see the `sizeLimit` comment there. Whatever value you pick should come from one shared `values.yaml` key so the api and events pods cannot drift apart.

aivong-openhands (commented):
## Taste Rating: 🟡 Acceptable

Both gaps this PR identifies are real, and the shapes it chose are right. `gitSyncSecretFromSecret` matches the existing `serviceKeyFromSecret` / `automationWebhookSecretFromSecret` vocabulary, the new `openhands-secrets` template mirrors `automation-webhook-secret.yaml` line for line, the umbrella mirror is present, and ungating `volumes:` / `volumeMounts:` is the correct fix rather than nesting another conditional inside a conditional. I re-verified the rendering claims and they hold:

```
$ helm template t charts/openhands/charts/automation --set events.enabled=true | grep -A4 '^      volumes:'
      volumes:
        - name: workspace
          emptyDir: {}
$ helm template t charts/openhands/charts/automation --set caBundle.enabled=true   # workspace first, then ca-bundle ✓
$ helm lint charts/openhands/charts/automation                                     # 1 linted, 0 failed ✓
```

Three things need attention before this ships. None of them ask you to change the approach.

### 1. The `emptyDir` is unbounded, and this repo already learned that lesson

`charts/openhands/charts/runtime-api/templates/cleanup-cronjob.yaml` gives its scratch `emptyDir` a `sizeLimit` driven by `runtimeFileArchive.tmpSizeLimit` (default `10Gi`). This one gets `emptyDir: {}`.

What lands there is git clones of *consumer-supplied* repositories, one directory per org (`{workspace_base}/git-sync/{org_id}`, per OpenHands/automation#429). The total is a function of how many orgs enable sync and how large their repos are — nothing the chart can bound by construction. The automation pods declare `memory` and `cpu` limits but no `ephemeral-storage` request or limit, so an unbounded write path fills the node's ephemeral storage and the kubelet evicts the pod. That takes down the whole automation API, not just git sync. Inline comment below.

### 2. `/workspace` is hardcoded, decoupled from the value that actually decides it

The mount path is a literal `/workspace` in both templates, while the directory the service writes to is `AUTOMATION_WORKSPACE_BASE` — which this chart never sets, and which a consumer *can* set through the `.Values.env` passthrough that `automation.env` explicitly supports. Set it and the mount no longer covers the write path; the failure is silent until a clone lands on the container filesystem. Inline comment below.

### 3. Self-hosted gets a secret that reads as configured and isn't

`charts/openhands-secrets` now renders `automation-git-sync-secret` with `git-sync-secret: ""` whenever `automation.enabled`:

```
$ helm template s charts/openhands-secrets --set automation.enabled=true
...
  name: automation-git-sync-secret
data:
  git-sync-secret: ""
```

The template shape is right — that is exactly what `automation-service-key.yaml` and `automation-webhook-secret.yaml` do. The gap is that `replicated/config.yaml`, `replicated/secrets.yaml`, and the `secretsChecksum` list in `replicated/openhands.yaml` were not extended alongside it. So a Replicated install with automations on gets `AUTOMATION_GIT_SYNC_SECRET=""`, which is falsy: the service falls through `git_sync_secret` → `kv_secret` (also unset by this chart) → `is_local_mode` is false → refuses with a 503 naming an env var the operator has no config option for.

Guideline 7 says defaults serve the self-hosted user. The description lists this as an ops follow-up, but unlike the SOPS secrets — which genuinely live in another repo — this is three edits *in this repo*, in the same three files the webhook secret already touches, and the established `value: '{{repl RandomString 32}}'` pattern generates-once-and-persists, which is exactly the semantics a wrapping key needs. Landing it here keeps the self-hosted default coherent.

**[TESTING GAPS]**

`charts/openhands/charts/runtime-api/tests/cleanup_archive_volumes_test.yaml` is a regression test for precisely the bug class this PR touches: a volume list item emitted without its `volumes:` key, silently merged into `imagePullSecrets`, yielding an invalid Pod spec that took out the staging reaper. That test exists because someone paid for it.

This PR ungates `volumes:` and `volumeMounts:` on two deployments and adds no test. The `helm unittest … 14 tests passed` in the description is the three pre-existing suites (`events_pdb`, `events_topology`, `storage_env`) — none of them assert anything about volumes, so that number is unchanged by this PR and proves nothing about it. A short suite asserting, on both `deployment.yaml` and `events-deployment.yaml`, that `spec.template.spec.volumes[0].name == workspace` with `caBundle.enabled=false`, that the mount is a real `volumeMounts` entry and not an `env` entry, and that both volumes render in order with `caBundle.enabled=true`, would lock this down.

---

**[RISK ASSESSMENT]**
- [Overall PR] ⚠️ Risk Assessment: 🟡 **MEDIUM**

Pod-spec surgery on both deployments of a live service, in the same template region that caused a prior staging outage, with no regression test. Mitigating factors are real: rendering is verified, `optional: true` means a missing secret cannot block pod startup, the values changes are purely additive so existing values files keep working, and `automation.enabled` defaults to `false`. The residual risk is operational rather than structural — unbounded node-local scratch on a data-bearing path, and a self-hosted configuration that looks wired and isn't.

**VERDICT:**
❌ **Needs rework** — the design is sound and I would approve it with items 1 and 3 addressed; the test is what stops the next volumes regression.

**KEY INSIGHT:**
The PR is right that this workspace data is disposable, but disposable bounds *durability*, not *size* — and the one chart in this repo that already mounts consumer-sized scratch into a pod caps it for exactly that reason.

---

_This review was generated by an AI agent (OpenHands) on behalf of @enyst, running `.agents/skills/code-review.md` from `main`._

---

>
>
>

aivong-openhands (approved):
🟢 **Taste Rating:** Good taste — the change follows established Helm chart patterns, keeps the workspace env var and mount sourced from one values block, preserves override-friendly env de-duplication, and adds targeted helm-unittest coverage for the rendering regression shape.

**[RISK ASSESSMENT]**
This touches production deployment wiring, secret management, and Replicated packaging for automation git sync. The implementation is additive and simple, the new secret reference is optional for rollout sequencing, there are no dependency changes, and the PR checks cover Helm lint/render/unit-test paths.

**VERDICT:**
✅ **Worth merging**: Core chart logic is sound; I found no important issues to block approval.

**KEY INSIGHT:**
The important design choice is tying `AUTOMATION_WORKSPACE_BASE` and the `emptyDir` mount to the same `workspace.mountPath` value, which prevents runtime path drift while keeping the configuration consumer-friendly.

_This review was generated by an AI agent (OpenHands) on behalf of the requester._

PR #1226, opened 2026-09-10
fix(openhands): add budget support bundle diagnostics

ak684:
## Summary

- add an `openhands-budget-diagnostics` support-bundle collector for OHE-3260
- capture sanitized Enterprise budget settings, per-member overrides, schema version, and recent budget maintenance results
- flag missing legacy cycle baselines and spend snapshots
- query live LiteLLM team/member financial state and report desired-versus-applied caps, roster drift, and missing expected private member budgets
- distinguish LiteLLM roster-only members on the shared team budget from members with private budget rows
- never query or emit managed/custom LLM keys or encrypted key columns

## Why

Today a customer support bundle cannot establish whether an org has the budget upgrade/reconciliation states tracked in OHE-3253 through OHE-3256. This gives Support an initial diagnostic slice that can identify missing baselines, failed maintenance, and stale LiteLLM caps without direct customer database access.

This intentionally does not attempt repair and does not fully diagnose OHE-3252 managed-key ownership; that needs a separately reviewed key-safe diagnostic or targeted operator inspection.

## Validation

- all Helm unit tests: 137 passed
- collector Python extracted from rendered bundle and compiled successfully
- full chart render and Helm lint passed
- `git diff --check` passed
- ran the collector read-only against the internal beta instance on Enterprise 1.59.1
  - correctly identified 13 LiteLLM roster members
  - identified 3 matching private member caps and 10 known members missing both cycle baselines and expected private budgets
  - reported the failed last reconciliation even though recent outer maintenance tasks were marked completed with zero task errors
  - emitted no key material

## Tracking

- [OHE-3260 — Add budget reconciliation observability and support-bundle diagnostics](https://linear.app/all-hands-ai/issue/OHE-3260/add-budget-reconciliation-observability-and-support-bundle-diagnostics)

### DIFF
--- a/charts/openhands/Chart.yaml
+++ b/charts/openhands/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 description: OpenHands is an AI-driven autonomous software engineer
 name: openhands
 appVersion: 1.59.1
-version: 0.65.1
+version: 0.65.0
 dependencies:
   - name: keycloak
     version: 24.7.5
--- a/charts/openhands/Chart.yaml
+++ b/charts/openhands/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 description: OpenHands is an AI-driven autonomous software engineer
 name: openhands
 appVersion: 1.59.1
-version: 0.65.0
+version: 0.65.1
 dependencies:
   - name: keycloak
     version: 24.7.5
--- a/charts/openhands/templates/troubleshoot/support-bundle.yaml
+++ b/charts/openhands/templates/troubleshoot/support-bundle.yaml
@@ -576,11 +576,13 @@ spec:
             from psycopg2.extras import RealDictCursor
 
             output = {
+                "collector_version": 1,
                 "schema_version": None,
                 "org_budget_settings": [],
                 "org_user_budget_overrides": [],
                 "recent_budget_maintenance_tasks": [],
                 "litellm_applied_state": [],
+                "summary": {},
                 "collector_errors": [],
             }
 
@@ -718,7 +720,7 @@ spec:
                 except (TypeError, ValueError):
                     return False
 
-            def desired_member_budget(budget, user_id):
+            def desired_member_budget(budget, user_id, live_spend):
                 if not budget.get("enabled"):
                     return {"status": "resolved", "max_budget_in_team": None}
                 override = overrides_by_member.get((str(budget["org_id"]), user_id))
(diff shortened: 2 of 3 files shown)


charts/openhands/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 description: OpenHands is an AI-driven autonomous software engineer
 name: openhands
 appVersion: 1.59.1
-version: 0.65.0
+version: 0.65.1
```
aivong-openhands:
```suggestion
```

Can remove the chart version bump as it is now handled by release please

aivong-openhands (commented):
Taste Rating: 🟡 **Acceptable** — the collector is isolated and read-only, but one core comparison conflates numeric caps with LiteLLM budget source.

**[CRITICAL ISSUES]**
- [charts/openhands/templates/troubleshoot/support-bundle.yaml, Line 893] **Correctness**: `budget_matches` treats `desired_max_budget_in_team == None` as “the applied max budget must also be `None`”. That is not the reconciliation contract. In the Enterprise budget sync code, an expected member budget of `None` means “this user should be on the shared team budget”; readback only fails if the user is not using the shared budget. With this collector, an org with a team monthly limit but no per-user default limit will report every shared-budget member as mismatched because `applied` is `team.max_budget`, not `None`. The inverse can also slip through: a private membership with `max_budget: null` can match even though it is not sharing the team budget. Please make the match source-aware: when `requires_private_budget` is false, healthy should mean shared/no private budget; when it is true, healthy should require a private budget and compare the numeric private cap.

**[RISK ASSESSMENT]**
  This is an isolated, read-only support-bundle collector, so it does not affect runtime behavior. The risk is elevated because it handles budget diagnostics and LiteLLM financial state; a false degraded result is likely to send Support down the wrong path for the exact incident class this collector is meant to diagnose.

**VERDICT:**
❌ **Needs rework**: the collector should preserve the reconciliation distinction between shared team budgets and private member budgets before this diagnostic is reliable.

**KEY INSIGHT:**
The diagnostic can only mirror budget reconciliation if “budget source” and “budget value” stay separate in the match logic.

_This review comment was created by an AI agent (OpenHands) on behalf of the user._

---

>
> 1. Add a `.agents/skills/custom-codereview-guide.md` file to your branch (or edit it if one already exists) with the `/codereview` trigger and the context the reviewer is missing (e.g., “Security concerns about X do not apply here because Y”). See the [customization docs](https://docs.openhands.dev/openhands/usage/use-cases/code-review#customization) for the required frontmatter format.
>
>

dylan-openhands (approved):
Seems like a great add to support bundles. Approved and my review focused mainly on the avoidance of including any sensitive information in the resulting output (already noted in the comments/PR description, but was double checking)

Edit: I see the review above and commentary around it but dont feel equipped to confirm the accuracy of that claim. Any way we can get a sample support bundle? Maybe via the unstable deployment?

charts/openhands/Chart.yaml
ak684: Removed the manual chart version bump in 38d7279; release-please now owns the chart version.

aivong-openhands (approved):
Taste Rating: 🟢 **Good taste** — the collector is isolated, read-only, and follows the existing support-bundle exec pattern without adding chart configuration surface area.

**[RISK ASSESSMENT]**
  This support-bundle collector reads budget and LiteLLM financial state from customer environments, so the review focus is data exposure and diagnostic accuracy. I verified the collector avoids managed/custom LLM key columns, does not emit the LiteLLM API key, uses a read-only DB session, and the latest `member_budget_matches` logic preserves the shared-team-budget vs private-member-budget distinction used by Enterprise reconciliation. GitHub checks, including Helm Unit Tests and chart lint/test, are passing.

**VERDICT:**
✅ **Worth merging**: the diagnostic is scoped to observability, matches the relevant reconciliation semantics, and has coverage for rendering plus key-column regressions.

**KEY INSIGHT:**
The change adds useful budget reconciliation visibility without attempting repair or expanding the self-hosted chart interface.

PR #1229, opened 2026-09-10
feat(agent-canvas): route cloud-proxy to Canvas service

juanmichelini:
## Summary

Adds first-class chart support for routing Agent Canvas cloud-runtime proxy requests to the Agent Canvas service.

## Why

Agent Canvas feature previews mount the frontend at `/canvas`, while the rest of `/api/*` normally routes to the OpenHands app. The SDK 405 fix lives on the SDK agent-server route `POST /api/cloud-proxy`, not on the enterprise app. Without chart-level routing, a feature preview can still send `/api/cloud-proxy` to the wrong backend and return `405 Method Not Allowed` even when the SDK image contains the fix.

## Changes

- Add `agent-canvas.cloudProxy` values:
  - `enabled`
  - `path` (default `/api/cloud-proxy`)
  - `agentServerPort` (default `18000`)
- When enabled, render `AGENT_CANVAS_ENABLE_SDK_CLOUD_PROXY=1` and `AGENT_CANVAS_SDK_CLOUD_PROXY_PORT=<port>` into the Agent Canvas pod.
- When enabled, add an umbrella ingress path routing `/api/cloud-proxy` to the `agent-canvas` service while leaving all other `/api` paths on the OpenHands app.
- Add Helm unittest coverage for the ingress path and deployment env.
- Document the new chart option and expected resource bump when the Canvas pod starts an SDK agent-server helper.

## Validation

```text
helm dependency build charts/openhands
helm template openhands charts/openhands -n openhands \
  --set agent-canvas.enabled=true \
  --set agent-canvas.ingress.enabled=true \
  --set agent-canvas.ingress.host=app.example.com
# confirmed default render has no /api/cloud-proxy path

helm template openhands charts/openhands -n openhands \
  --set agent-canvas.enabled=true \
  --set agent-canvas.ingress.enabled=true \
  --set agent-canvas.ingress.host=app.example.com \
  --set agent-canvas.cloudProxy.enabled=true \
  --set agent-canvas.cloudProxy.agentServerPort=18080
# confirmed enabled render has /api/cloud-proxy ingress path and AGENT_CANVAS_* env

helm lint charts/openhands
# 1 chart linted, 0 failed
```

Local `helm unittest` was not available in my environment; CI should run the added chart tests.

## Related

- OpenHands/software-agent-sdk#4878
- OpenHands/software-agent-sdk#4914
- OpenHands/OpenHands#17131
- OpenHands/saas-deploy#970

### DIFF
--- a/charts/openhands/charts/agent-canvas/values.yaml
+++ b/charts/openhands/charts/agent-canvas/values.yaml
@@ -31,6 +31,15 @@ deployment:
 service:
   port: 8000
 
+# Optional cloud runtime proxy. When enabled, the static server starts the
+# bundled SDK agent-server and routes only cloud-proxy envelopes to it; the
+# umbrella chart can then expose /api/cloud-proxy to this service while keeping
+# the rest of /api on the OpenHands app service.
+cloudProxy:
+  enabled: false
+  path: /api/cloud-proxy
+  agentServerPort: 18000
+
 # Ingress for the agent-canvas service. The umbrella `openhands` chart renders
 # this path on the host supplied here when this chart is used as a subchart.
 ingress:
--- a/charts/openhands/charts/agent-canvas/README.md
+++ b/charts/openhands/charts/agent-canvas/README.md
@@ -55,6 +55,31 @@ agent-canvas:
       secretName: app-all-hands-dev-tls
 ```
 
+### Exposing the Cloud runtime proxy
+
+Set `cloudProxy.enabled=true` when the same host also needs Canvas browser
+requests to reach the SDK agent-server's `POST /api/cloud-proxy` endpoint.
+The static server starts the bundled SDK agent-server on `cloudProxy.agentServerPort`
+and the umbrella chart adds a same-host ingress path (default `/api/cloud-proxy`)
+to this service. All other `/api` paths remain owned by the OpenHands app.
+
+```yaml
+agent-canvas:
+  enabled: true
+  image:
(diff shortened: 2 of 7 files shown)


charts/openhands/charts/agent-canvas/values.yaml
```diff
@@ -31,6 +31,15 @@ deployment:
 service:
   port: 8000
 
+# Optional cloud runtime proxy. When enabled, the static server starts the
+# bundled SDK agent-server and routes only cloud-proxy envelopes to it; the
+# umbrella chart can then expose /api/cloud-proxy to this service while keeping
+# the rest of /api on the OpenHands app service.
+cloudProxy:
+  enabled: false
+  path: /api/cloud-proxy
```
aivong-openhands, line 40: This value makes `cloudProxy.path` look configurable, but the deployment only passes the helper an enable flag and port. The SDK cloud-proxy helper currently registers `/api/cloud-proxy`, and no template passes this custom path into the pod, so a consumer setting the path to something like `/internal/cloud-proxy` (as the new umbrella test documents) would get an ingress route to agent-canvas for a path the static server does not handle. Could we either keep this path fixed/remove the override, or pass the configured path through to the static server/helper and test the deployment side too?

aivong-openhands (commented):
I found one issue that should be addressed before this is ready.

This review was created by an AI agent (OpenHands) on behalf of the user.

PR #1231, opened 2026-09-11
feat(openhands): add budget preflight and reconciliation gate hook jobs

hieptl:
## Description

Adds a budget upgrade preflight and a post-upgrade reconciliation gate to the `openhands` chart, as two Helm hook Jobs on the enterprise image (OHE-3256):

- `<release>-budget-preflight` (`pre-upgrade`, weight 5, default mode `acknowledge`): read-only, runs the new image against the not-yet-migrated database and reports each enabled organization's budget state (missing baselines, members missing from LiteLLM, unmapped identities, cap drift, over-cap state, snapshot age, last sync) before enforcement changes.
- `<release>-budget-reconcile-gate` (`post-upgrade`, weight 5, default mode `strict`): reconciles every enabled organization's LiteLLM caps in-process and verifies them by readback. In `strict` a blocking finding exits non-zero, fails `helm upgrade`, and marks the Replicated version failed; `acknowledge` records the findings and lets the release proceed.

Both Jobs are gated on `.Release.IsUpgrade`, so they never render on install or under ArgoCD (which renders with `helm template`), use `backoffLimit: 0` with `activeDeadlineSeconds` under Helm's per-hook wait (300 s / 540 s against the 600 s Replicated timeout), and keep their last run with `hook-delete-policy: before-hook-creation` so the JSON artifact in the pod log survives until the next upgrade. Each Job prints one `org_budget_preflight` artifact line.

Also included: `budgetPreflight` / `budgetReconcileGate` values blocks, a render-time guard for invalid modes, two support-bundle `logs` collectors, a Replicated Config group "Upgrade Checks" (`budget_preflight_mode`, `budget_reconcile_gate_mode`) mapped into the chart values, helm-unittest suites for both Jobs and the guard, and an upgrade-rollback runbook section covering artifact capture, strict-gate recovery, manual runs, and restoring LiteLLM-side caps after a rollback.

**Depends on** OpenHands/enterprise#362, which adds `run_budget_preflight.py` to the image. This PR must not merge before a release containing that script is the chart's default `image.tag`; with an older image the pre-upgrade Job would fail on the missing module and block upgrades. Kept as a draft until then.

Validation:

- `helm template t charts/openhands --is-upgrade --show-only templates/budget-preflight-job.yaml --show-only templates/budget-reconcile-gate-job.yaml` renders both Jobs with the expected hooks and env; without `--is-upgrade` nothing renders.
- `helm unittest charts/openhands`: 150 passed in 31 suites.
- `helm lint charts/openhands`: passed.
- `make lint` (Replicated): exit 0; only pre-existing warnings on `troubleshoot/secrets.yaml`.

## Helm Chart Checklist

New keys ship with defaults and `values.schema.json` does not restrict additional top-level keys, so existing values files are unaffected. The README lists no job values, so no README change was needed. A live upgrade on a Replicated test instance has not been run yet.

## Additional Notes

- The strict default on the post-upgrade gate is deliberate: it is what makes "failed reconciliation blocks release health" true out of the box. Customers with a known-dirty organization can select Acknowledge in the "Upgrade Checks" Config group before upgrading.
- Mutual exclusion between the hook and the 15-minute `budget-maintenance` CronJob is tracked separately (OHE-3259); both are idempotent.
- SaaS (ArgoCD) deployments get neither Job; the runbook documents running the preflight by hand there.

### DIFF
--- a/.agents/skills/upgrade-rollback-runbook.md
+++ b/.agents/skills/upgrade-rollback-runbook.md
@@ -186,6 +186,97 @@ kubectl rollout status deploy/openhands -n <namespace>
 Verify: images back at X, `alembic_version` back at the X value, pods healthy, no
 `migrate-db` crash-loop.
 
+## Budget policy hooks and LiteLLM external state
+
+Charts that carry `budget-preflight-job.yaml` run two Helm hook Jobs on every
+`helm upgrade` of the `openhands` release (never on install, and never under
+ArgoCD, which renders with `helm template`):
+
+| Job | Hook | Default mode | What it does |
+|---|---|---|---|
+| `<release-name>-budget-preflight` | pre-upgrade | `acknowledge` | Read-only. Runs the new image against the not-yet-migrated database and reports each enabled organization's budget state as found. |
+| `<release-name>-budget-reconcile-gate` | post-upgrade | `strict` | Reconciles every enabled organization's LiteLLM caps and verifies them by readback. |
+
+Modes come from `budgetPreflight.mode` / `budgetReconcileGate.mode` (Replicated
+Config group "Upgrade Checks"). `strict` fails `helm upgrade` on a blocking
+finding; `acknowledge` records the same findings and exits 0, which is the
+explicit availability-recovery mode. Expect both Jobs as substantive lines in
+the step-3 diff the first time a customer upgrades onto such a chart.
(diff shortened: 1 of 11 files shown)


tofarr (approved):
🍰

PR #1249, opened 2026-09-14
chore: PLTF-3548 raise preflight memory recommendation to 32Gi

aivong-openhands:
## Why

OHE Replicated preflight recommended at least 16GB of memory per node. The practical target is higher, so this raises the node-resources recommendation to warn when a node has less than 32GB rather than 16GB.

It stays a warning, not a hard fail. Clusters with 16GB run fine today, so failing the preflight would wrongly block them from installing or upgrading. The warning surfaces the higher recommendation without stopping anyone.


dylan-openhands (approved):
Approved, but note the PR title check for the conventional commit format 👍

PR #1253, opened 2026-09-15
chore: PLTF-3561 Add laminar app-server and consumer pod logs to support bundle

aivong-openhands:
## Why

Support bundles collected logs only for the `laminar-clickhouse` pod, leaving the Laminar `app-server` and `app-server-consumer` pods — where request handling and queue consumption run — invisible when diagnosing Laminar problems. This adds two `logs` collectors in the existing `.Values.laminar.enabled` block, for `app=laminar-app-server` and `app=laminar-app-server-consumer`, reusing the `maxAge: 336h` of the neighbouring collectors. No values or defaults change, so any install with Laminar enabled starts capturing these logs with no configuration.

## Validation

- **Live support bundle, Laminar enabled on KinD** — installed the chart with `laminar.enabled=true`, then ran `support-bundle --load-cluster-specs -n openhands`. The archive contained dedicated `app/laminar-app-server/logs/…` and `app/laminar-app-server-consumer/logs/…` collector trees holding the running pods' container logs (`app-server.log`, `app-server-consumer.log`), alongside the existing `app/laminar-clickhouse`. Both directories are named after the added collectors and are absent without the change; reproduced across independent bundle runs.


juanmichelini (approved):
LGTM

PR #1256, opened 2026-09-16
fix(replicated): PLTF-3561 drop dead laminar-quickwit status informers

aivong-openhands:
## Why

The `statusInformers` list tracks the `laminar-quickwit-*` deployments and services (control-plane, janitor, metastore, indexer, searcher, headless). Quickwit is laminar's full-text search engine: it stores its indexes in an S3 bucket and keeps its metastore in the postgres it shares with the rest of laminar. The chart treats the S3 index-root URI as the opt-in gate for the whole stack and renders no Quickwit workloads when it is empty; search degrades gracefully when Quickwit is off. The Replicated install never sets `quickwit.s3.defaultIndexRootUri` and exposes no config option to set it, so these objects never exist.

Each informer, though, is gated only on `analytics_enabled`, so enabling analytics registers informers KOTS can never satisfy. KOTS rolls the worst informer state into the app status, so these always-missing targets force the whole application to `missing` even though every laminar pod is healthy. Removing them lets the status match reality; if a Quickwit config option is added later, the informers should return gated on that option rather than on `analytics_enabled`.

## Validation

- **Support bundle from an affected install** — app status was `missing` while every laminar pod was Running, and the kotsadm status payload listed all `laminar-quickwit-*` informers as `missing` with the deployed laminar objects `ready`; no `laminar-quickwit-*` objects appear in the collected cluster resources.
- **Effective config** — `replicated/openhands.yaml` leaves `laminar.quickwit.s3.defaultIndexRootUri` empty and `replicated/config.yaml` exposes no option to set it, so the Quickwit workloads are never rendered.


aivong-openhands:
@openhands create a follow up issue to create a config option for laminar quickwit

aivong-openhands:
Follow-up issue created: #1260 — **feat(replicated): expose a ConfigOption to enable laminar Quickwit full-text search** (https://github.com/OpenHands/OpenHands-Cloud/issues/1260). It covers the missing ConfigOption wiring through `replicated/openhands.yaml` into `laminar.quickwit.s3.defaultIndexRootUri`, and notes that the informers removed here should be re-introduced gated on the new option rather than `analytics_enabled`.

This comment was created by an AI agent (OpenHands) on behalf of the user.

aivong-openhands:
@openhands resolve conflicts

aivong-openhands:
Conflicts resolved in 3e8bf7e. Merged main (which removed the adjacent laminar-query-engine informersin #1255) into this branch; the resolution keeps both removals, so the final laminar statusInformers list contains no laminar-query-engine or laminar-quickwit-* entries (both are dead workloads that KOTS can never satisfy).n
This comment was created by an AI agent (OpenHands) on behalf of the user.

mamoodi (approved):

PR #1258, opened 2026-09-18
fix(replicated): PLTF-3560 cover all LiteLLM credentials in restart checksum

aivong-openhands:
HUMAN:

AGENT:

---

## Why

LiteLLM reads its provider credentials from the `litellm-env-secrets` Secret via `envFrom`. Kubernetes never re-reads an `envFrom` Secret and never rolls the pod when it changes, so the only thing that restarts the LiteLLM pod after a credential edit is the `checksum/litellm-credentials` pod annotation in `replicated/openhands.yaml`.

That annotation only hashed 3 values (`litellm_api_key`, `litellm_admin_password`, `litellm_salt_key`). Changing any *other* provider secret in the admin console (an Anthropic key, an AWS/Bedrock credential, an Azure secret, etc.) did not change the checksum, so the pod kept running with the old credentials until someone manually restarted it. That is PLTF-3560: "LiteLLM pod does not restart when provider API key is changed."

## Summary

- Expand `checksum/litellm-credentials` to hash all 26 credential inputs that feed `charts/openhands-secrets/templates/litellm-env-secrets.yaml`, so any provider-credential edit rolls the pod.
- Add a 12-line comment in `replicated/openhands.yaml` documenting the invariant and the two special cases:
  - `litellm_api_key` is not in `litellm-env-secrets` at all. It reaches the pod as `PROXY_MASTER_KEY` via a `secretKeyRef` to the separate `lite-llm-api-key` Secret, which has the same no-auto-restart behavior, so it is hashed here explicitly.
  - `bedrock_model_ids` is the textarea ConfigOption; `secrets.yaml` derives `config.bedrock_model_id` from its first non-empty line, so the textarea (not `bedrock_model_id`) is hashed. `google_vertex_credentials` is read with `ConfigOptionData`.
- Add `scripts/check_litellm_checksum.py`, a CI guard that parses `litellm-env-secrets.yaml`, enumerates the config options it consumes, and fails if the annotation does not hash exactly that set (plus the allow-listed `litellm_api_key`).
- Add `scripts/test_litellm_credentials_checksum.py` covering the guard and its wiring assertions.
- Wire both into CI: run the guard in `check-secret-checksum.yml`, and expand the `paths` triggers for both `check-secret-checksum.yml` and `test-scripts.yml` so edits to the template, chart, guard, or workflow re-run the checks.

## Issue Number

PLTF-3560

## How to Test

```
python3 scripts/check_litellm_checksum.py
python3 -m pytest scripts/test_litellm_credentials_checksum.py -q
```

The guard passes against the updated annotation. Removing any `ConfigOption`/`ConfigOptionData` entry from `checksum/litellm-credentials` (or adding a new `.Values.config.*` reference to `litellm-env-secrets.yaml` without hashing it) makes `check_litellm_checksum.py` fail.

## Video/Screenshots

N/A - no UI change.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

The checksum only ever adds restart triggers, which is the correct direction; the prior under-coverage was the bug. No chart template logic, storage, or env-var wiring changes.

`EXPECTED_DEPENDENCY_COUNT` in the guard and the 26-entry `printf` in the annotation must move together when a provider is added, so a new provider now touches the template, `secrets.yaml`, the annotation, and this constant. The guard's total-coverage accounting is the real check; the count is a secondary backstop.

---

### Update: mutation-review follow-up

Addressed the mutation review (10 hand-written mutants). Added 7 targeted tests to `scripts/test_litellm_credentials_checksum.py` and hardened the CI-wiring assertion. Suite goes 21 → 31 passing; each new test was verified to fail with its mutant applied and pass on this branch:

- C2 — the guard could be silently unwired from CI. Replaced the substring text-match (satisfied by the `paths:` comment even if the run step is deleted) with `test_checker_is_actually_run_by_a_workflow_step`, which walks the parsed workflow steps.
- M6 — `test_printf_format_is_pipe_separated` pins the `|` separators so adjacent credential values can't trade a character and leave the hash unchanged.
- M1 / M2 — exercise the `EXPECTED_DEPENDENCY_COUNT` backstop (stage 6) and the `ORIGINAL_INPUTS` preservation loop (stage 5), which had no test that made them fire.
- M4 / M3 — reject every `_STRUCTURE_CONSTRUCTS` entry (`include`/`define`/`with`/`range`) and the generic "unsupported `.Values` reference" branch.
- M8 — assert the diagnostic line number is exact (catches an off-by-one in `_line_of`).

Also reconciled this description with the diff: the `laminar-query-engine` statusInformers change is not part of this PR (it shipped separately in #1255), so that bullet was removed.


aivong-openhands (commented):
> Note: posting as a COMMENT because the GitHub token used is the PR author account, which cannot formally APPROVE its own PR. The intended decision is **APPROVE**.

**Review decision: APPROVE** (with comments)

Reviewed against the OpenHands-Cloud code-review guidelines. This is a CI/workflow + KOTS-config bug fix; the core change is correct, complete, and verified, so it meets the bar to approve.

### What the change does
The `checksum/litellm-credentials` pod annotation previously hashed only 3 config values (`litellm_api_key`, `litellm_admin_password`, `litellm_salt_key`). Because LiteLLM reads provider secrets from `litellm-env-secrets` via `envFrom` — which Kubernetes never re-reads and never rolls the pod for — changing any *other* provider key in the admin console silently had no effect until a manual pod restart. The fix expands the checksum to all 26 credential inputs and adds a CI guard (`check_litellm_checksum.py`) plus tests to keep the annotation in sync with the Secret template.

### Verification performed
- **Coverage is exact.** `charts/openhands-secrets/templates/litellm-env-secrets.yaml` references exactly 25 distinct `.Values.config.*` keys. The annotation lists those 25 (with the `bedrock_model_id`→`bedrock_model_ids` indirection) plus the allow-listed `litellm_api_key` = 26 `ConfigOption` calls. Matches `EXPECTED_DEPENDENCY_COUNT = 25`.
- **Both special cases check out.** `secrets.yaml` derives `bedrock_model_id` from `ConfigOption "bedrock_model_ids"` (so hashing the textarea is correct), and `google_vertex_credentials` is read via `ConfigOptionData` — the annotation correctly uses `ConfigOptionData` for it.
- **No false-positive risk in the coverage checker.** The template's only `.Values` references are `.Values.config.*` (all covered); `.Release.Namespace` isn't matched, and there are no `include`/`define`/`with`/`range` constructs.
- **CI is green** on the relevant checks: `check-secret-checksum`, `test-scripts`, and all chart `lint-and-test` jobs. (`last-deploy` failed, but that's the preview-deploy job, unrelated to this change.)

### Comments (non-blocking)

1. **The PR description does not match this PR.** The body describes an entirely different change — org budget threshold alerting (`replace_thresholds`, `_maybe_send_alerts`, `tests/unit/test_org_budget_store.py`), none of which is in this diff. The title is correct, but the description looks copy-pasted from another draft PR and will mislead anyone reading the history. Please replace it with the actual why/summary for the LiteLLM checksum fix (the 12-line comment added to `openhands.yaml` is a great basis).

2. **The guard is heavy for what it enforces** (discussion, not a mandate). `check_litellm_checksum.py` is ~348 lines implementing a formal access-form grammar, total-coverage accounting, structure-construct rejection, a dependency-count backstop, and `KNOWN_INDIRECTIONS`/`ALLOWED_EXTRA`/`ORIGINAL_INPUTS` sets — to enforce "the annotation must list the same config options the template consumes." It's well-tested and fully self-contained (no production chart logic touched), so the risk is low and I'm not asking you to cut it. Just worth confirming the team is comfortable maintaining the grammar: if the template ever grows a construct the parser rejects, the intended failure mode is a hard CI failure that a future contributor must extend the grammar to satisfy.

3. **Minor:** `EXPECTED_DEPENDENCY_COUNT = 25` and the 26-entry printf must move together on every provider addition, so a new provider now touches three coordinated spots (template, `secrets.yaml`, annotation) plus this constant. The total-coverage accounting already catches an add+remove that nets to zero, so the count backstop is genuinely secondary — fine as-is, just flagging the coordination cost.

### Risk assessment
**Low.** Impact is confined to a KOTS pod annotation that only *adds* restart triggers (the correct direction — the prior under-coverage was the bug). No chart template logic, storage, or env-var wiring changes. Worst case of the new CI guard is a false CI failure, not a bad deploy.

---
_This review was generated by an AI agent (OpenHands) on behalf of the repository maintainer._

aivong-openhands:
Thanks for the review!

**Comment 1 (description mismatch):** Fixed. The description was copy-pasted from a sibling draft PR. I've replaced it with the actual why/summary for the LiteLLM checksum fix, based on the comment added to `openhands.yaml` as you suggested.

**Comment 2 (guard is heavy):** Acknowledged, leaving as-is. The guard is self-contained (no production chart logic touched) and fully tested, so the maintenance surface is isolated to the script itself. The intended failure mode is exactly what you described: if the template ever grows a construct the parser doesn't understand, CI fails hard and a contributor extends the grammar rather than silently under-hashing. That's the safe direction for this class of bug, which is why I preferred a strict parser over a loose heuristic.

**Comment 3 (EXPECTED_DEPENDENCY_COUNT coordination cost):** Acknowledged, keeping as-is. The total-coverage accounting is the real check; the count is a deliberate secondary backstop for the add+remove-nets-to-zero case. The coordination cost is real but bounded: a new provider already touches the template, `secrets.yaml`, and the annotation, so the constant is one more line in a change that's already cross-file by nature.

_This reply was generated by an AI agent (OpenHands) on behalf of the repository maintainer._

aivong-openhands:
## Mutation review of this PR's tests

I hand-wrote 10 mutants against the diff and ran them through `scripts/test_litellm_credentials_checksum.py` (baseline: **21 passed in 1.0s**). Three were controls that revert the PR's actual fix and must die; seven were candidates aimed at the machinery around it. Scope is test strength only, not correctness.

### Controls

| Mutant | Result |
|---|---|
| C1 — revert the annotation to the original 3-input `printf` | ❌ caught |
| C3 — read `google_vertex_credentials` with `ConfigOption` instead of `ConfigOptionData` | ❌ caught |
| C2 — delete the `run: python3 scripts/check_litellm_checksum.py` step from `check-secret-checksum.yml` | ✅ **survived** |

C1 dies in six different tests, which is the right shape for a central claim — `test_repo_checksum_covers_all_dependencies` catches the coverage loss, `test_missing_mapping` and `test_cli_failure` catch it through the fixture and the CLI independently, and `test_annotation_shape` catches it on the surface. C3 dies purely because `test_annotation_shape` asserts the `ConfigOptionData` spelling by name, which is a genuinely good assertion: the file-type/scalar distinction is invisible in the rendered YAML and only shows up as an empty hash input in production. `_make_fixture_root` copying the real three files rather than inventing synthetic ones is what lets the negative cases stay precise.

### Survivors

| Mutant | Result |
|---|---|
| C2 — delete the CI step that runs the guard | ✅ 21 passed |
| M1 — disable the `EXPECTED_DEPENDENCY_COUNT` backstop (stage 6) | ✅ 21 passed |
| M2 — disable the `ORIGINAL_INPUTS` preservation check (stage 5) | ✅ 21 passed |
| M6 — strip the `\|` separators out of the `printf` format | ✅ 21 passed |
| M4 — drop `include`/`define`/`with`/`range` from `_STRUCTURE_CONSTRUCTS` | ✅ 21 passed |
| M3 — silence the generic "unsupported `.Values` reference" branch | ✅ 21 passed |
| M8 — off-by-one in `_line_of` | ✅ 21 passed |

---

#### C2 — the guard can be unwired from CI and nothing notices

This is the headline. The whole PR rests on the guard running on every relevant change, and `test_ci_workflow_wires_new_checker_and_paths` does check that — but with `assert "scripts/check_litellm_checksum.py" in steps_text`, a substring match against the raw file text. The step name on the line above (`Every LiteLLM credential must be in checksum/litellm-credentials`) does not contain the path, but the comment block in the `paths:` filter does: `scripts/check_litellm_checksum.py` is listed there four times. So deleting the `run:` step entirely leaves the assertion satisfied. The guard stops executing, the `paths` triggers still fire, the job still goes green, and the next provider added without a checksum entry ships silently — exactly PLTF-3560 again.

The rest of that test parses YAML properly; this one assertion reaches for the text instead. Walking the steps closes it:

```python
def test_checker_is_actually_run_by_a_workflow_step() -> None:
    wf_path = REPO_ROOT / ".github" / "workflows" / "check-secret-checksum.yml"
    wf = yaml.safe_load(wf_path.read_text(encoding="utf-8"))
    runs = [
        step.get("run", "")
        for job in wf["jobs"].values()
        for step in job.get("steps", [])
    ]
    assert any("scripts/check_litellm_checksum.py" in r for r in runs), (
        "no workflow step runs scripts/check_litellm_checksum.py"
    )
```

Verified: passes on this branch, fails with the step deleted.

#### M6 — the `|` separators in the `printf` are unasserted

`test_printf_arity_matches_config_option_count` counts `%s` against `ConfigOption` calls, and `test_annotation_shape` checks the prefix and suffix, but nothing looks at what sits *between* the placeholders. Replacing `"%s|%s|%s|…"` with `"%s%s%s…"` keeps the arity correct, keeps the shape valid, and passes all 21 tests.

Unseparated concatenation means adjacent values can trade a character without changing the hash: an operator moving a trailing digit from `azure_client_id` to `azure_client_secret` produces the same digest, so the pod does not roll and LiteLLM keeps the stale credentials. That is precisely the failure mode this annotation exists to prevent, and the separator is the only thing standing between it and the fix. It is a narrow window, but it costs one assertion to close:

```python
def test_printf_format_is_pipe_separated() -> None:
    value = _annotation_value()
    fmt = re.search(r'printf\s+"([^"]*)"', value).group(1)
    n = len(re.findall(r"ConfigOption(?:Data)?\s+\"", value))
    assert fmt == "|".join(["%s"] * n), (
        f"format must be {n} '%s' joined by '|'; without separators two adjacent "
        f"values can trade a character and leave the hash unchanged, so the pod "
        f"never rolls. Got: {fmt!r}"
    )
```

Verified: passes on this branch, fails with the separators removed.

#### M1 and M2 — two of the six `check()` stages are never exercised

The suite drives stages 2 through 5 through fixtures (`test_missing_mapping`, `test_changed_indirection`, `test_over_broad`, `test_unresolvable_key`, the access-form tests). Stage 6 (`EXPECTED_DEPENDENCY_COUNT`) and the `ORIGINAL_INPUTS` preservation loop in stage 5 have no test that makes them fire. `EXPECTED_DEPENDENCY_COUNT` is imported at the top of the test file but only ever read, never triggered.

I checked whether these are equivalent mutants — whether another stage already covers the same ground — and they are not:

- Remove `mistral_api_key` from both the template and the annotation (a coherent "we dropped this provider" edit): coverage accounting is satisfied and **only the count backstop fails**. With M1 applied, `check()` returns clean.
- Remove `litellm_admin_password` from the template and the annotation and update the count to match: **only the preservation check fails**. With M2 applied, `check()` returns clean.

Both are the sole guard in their scenario, so silently losing either is a real regression. Note that the two `assert not (len(errors) == 1 and "dependency-count backstop" in errors[0])` lines in `test_new_dependency` / `test_unresolvable_key` assert the backstop is *not* the only message — the inverse of what is missing.

```python
def test_dependency_count_backstop_fires(tmp_path: Path, monkeypatch) -> None:
    root = _make_fixture_root(tmp_path)
    monkeypatch.setattr(clc, "EXPECTED_DEPENDENCY_COUNT", EXPECTED_DEPENDENCY_COUNT + 1)
    errors = check(root)
    assert any("dependency-count backstop" in e for e in errors), errors


def test_preservation_failure_for_original_input(tmp_path: Path) -> None:
    root = _make_fixture_root(tmp_path)
    text = _read(root, OPENHANDS_REL)
    _write(root, OPENHANDS_REL, _remove_checksum_option(text, "litellm_salt_key"))
    errors = check(root)
    assert any(
        "preservation failure" in e and "litellm_salt_key" in e for e in errors
    ), errors
```

`clc` is already imported as a module, so `monkeypatch.setattr` works without extra plumbing. Both verified: pass on this branch, fail with their mutant applied.

#### M4 and M3 — the refusal grammar is only partly pinned

`test_aliased_context` covers the `$cfg := .Values.config` binding, and `test_indirect_map_access` / `test_dynamic_key_access` cover the two named-diagnostic branches. Nothing covers the other four `_STRUCTURE_CONSTRUCTS` entries (`include`, `define`, `with`, `range`) or the generic `else` branch in `_check_coverage`. Deleting all four regex entries, or silencing the generic branch, passes the suite.

The docstring's claim is that discovery is "provably complete within the structure it parses", and these refusals are what makes that true — an `include` moves dependencies into a helper file the checker never reads, and the generic branch is what catches a `.Values` reference that matches no recognised form at all. Losing them turns a hard failure into a silent under-count.

```python
@pytest.mark.parametrize(
    "snippet,label",
    [
        ('{{- include "openhands.helper" . }}', "include"),
        ('{{- define "openhands.helper" }}{{- end }}', "define"),
        ("{{- with .Values.config }}{{- end }}", "with"),
        ("{{- range $k := list }}{{- end }}", "range"),
    ],
)
def test_structure_constructs_are_rejected(tmp_path: Path, snippet: str, label: str) -> None:
    root = _make_fixture_root(tmp_path)
    tmpl = _read(root, TEMPLATE_REL)
    _write(root, TEMPLATE_REL, tmpl + "\n" + snippet + "\n")
    errors = check(root)
    assert any(
        "template-structure violation" in e and label in e for e in errors
    ), errors


def test_unsupported_values_reference(tmp_path: Path) -> None:
    root = _make_fixture_root(tmp_path)
    tmpl = _read(root, TEMPLATE_REL)
    _write(root, TEMPLATE_REL, tmpl + '\n  EXTRA: {{ .Values.global.imageTag | quote }}\n')
    errors = check(root)
    assert any("unsupported .Values reference" in e for e in errors), errors
```

Both verified: pass on this branch, fail with their mutant applied.

#### M8 — line numbers in diagnostics are unasserted

`_line_of` returning `count("\n")` instead of `count("\n") + 1` passes the suite. `test_indirect_map_access` asserts `re.search(r":\d+:", e)` — that *a* number is present, not that it is right. Low production consequence (the checker still fails correctly), but the whole value of the line number is pointing an operator at the offending line, and off-by-one line numbers in a 60-line template are a small credibility tax on a tool whose job is to be trusted.

```python
def test_diagnostic_line_number_is_correct(tmp_path: Path) -> None:
    root = _make_fixture_root(tmp_path)
    tmpl = _read(root, TEMPLATE_REL)
    bad = '  EXTRA: {{ index .Values "config" "mistral_api_key" }}'
    new = tmpl + "\n" + bad + "\n"
    _write(root, TEMPLATE_REL, new)
    expected_line = new.splitlines().index(bad) + 1
    errors = check(root)
    assert any(
        f"{TEMPLATE_REL}:{expected_line}:" in e and "indirect map access" in e
        for e in errors
    ), (expected_line, errors)
```

Verified: passes on this branch, fails with the off-by-one applied.

---

### Not a test gap

- Adding all seven tests takes the suite from 21 to 31 passing in 1.5s, and every one of the ten mutants above then dies. I confirmed each proposed test individually passes unmodified and fails with its own mutant, not just that the suite as a whole turns red.
- `test_optional_values_litellm_blocks_are_recursive_and_dont_override` is the strongest test in the file and I could not get past it — it pins the `recursiveMerge`/override failure mode structurally rather than by example, which is the right call for a KOTS merge semantic that has no local reproduction.
- The PR description mentions dropping the `laminar-query-engine` statusInformers from `replicated/application.yaml`, but that file is not in the diff. Worth reconciling the description with the change before merge.

This comment was generated by an AI assistant on behalf of the user.

aivong-openhands:
Thanks for the mutation review — this was exactly the right level of scrutiny for a guard whose whole job is to be trusted. I've addressed all of it in `913d48381`.

**All 7 survivors now die** (suite 21 → 31 passing). I applied each of your proposed mutants locally and confirmed each new test fails with the mutant and passes on the branch:

| Mutant | New test | Result |
|---|---|---|
| C2 — delete the guard's `run:` step | `test_checker_is_actually_run_by_a_workflow_step` (walks parsed steps) | ❌ dies |
| M6 — strip `\|` separators | `test_printf_format_is_pipe_separated` | ❌ dies |
| M1 — disable count backstop | `test_dependency_count_backstop_fires` | ❌ dies |
| M2 — disable preservation loop | `test_preservation_failure_for_original_input` | ❌ dies |
| M4 — drop structure constructs | `test_structure_constructs_are_rejected` (parametrized ×4) | ❌ dies |
| M3 — silence generic branch | `test_unsupported_values_reference` | ❌ dies |
| M8 — `_line_of` off-by-one | `test_diagnostic_line_number_is_correct` | ❌ dies |

On **C2** specifically (the headline): I took your point that the substring match gave false confidence, so beyond adding the step-walking test I also removed the weak `assert "scripts/check_litellm_checksum.py" in steps_text` line from `test_ci_workflow_wires_new_checker_and_paths` so nothing is left leaning on the `paths:` comment.

I also **reconciled the description** — you were right that the `laminar-query-engine` statusInformers change isn't in this PR's diff; it shipped separately in #1255, so I dropped that bullet and noted the correction.

aivong-openhands:
## Live verification on a Replicated VM ✅

Verified on a self-hosted Replicated embedded-cluster VM (KOTS v1.130.4, app `openhands`,
LiteLLM deploy `openhands-litellm`) that after this PR, editing **any** LiteLLM provider
credential rolls the LiteLLM pod — including credentials that did not trigger a restart before.

**Builds**
- **Control (bug):** Unstable release 1632 — `checksum/litellm-credentials` hashes only 3 inputs.
- **Fix:** this PR's release 1634 — checksum hashes **26** inputs.

**Phase 0 — code guard (head `913d4838`)**
- `scripts/check_litellm_checksum.py` → OK (all 25 template deps + allow-listed `litellm_api_key`).
- `pytest scripts/test_litellm_credentials_checksum.py` → **31 passed**.

**Control — bug reproduced.** On the current build, `kots set config anthropic_api_key=… --deploy`
created a new version but the checksum annotation stayed `b75b1b5d…` and the LiteLLM pod did **not**
roll (same pod UID). The pod kept the old credential.

**After deploying this PR**, the checksum annotation went `b75b1b5d… (3-input) → 6f1caebe… (26-input)`
and rolled LiteLLM. Then every credential edit changed the annotation and rolled the pod:

| credential | note | annotation changed | pod rolled |
|---|---|---|---|
| anthropic_api_key | previously unhashed | yes | yes |
| openai_api_key | previously unhashed | yes | yes |
| aws_access_key_id | previously unhashed | yes | yes |
| azure_api_key | previously unhashed | yes | yes |
| **bedrock_model_ids** | textarea special case | yes | yes |
| **google_vertex_credentials** | `ConfigOptionData` file special case | yes | yes |
| litellm_api_key | core-3 regression (PROXY_MASTER_KEY) | yes | yes |
| litellm_admin_password | core-3 regression | yes | yes |
| litellm_salt_key | core-3 regression | yes | yes |

**Credential takes effect.** After a rolled `anthropic_api_key` edit, the new LiteLLM pod's
`printenv ANTHROPIC_API_KEY` matched the edited value (via `envFrom litellm-env-secrets`).

**Cleanup.** All credentials restored to originals (checksum returned to the exact baseline
`b75b1b5d…`); instance redeployed to its baseline release; LiteLLM + app healthy.

**Verdict:** the pre-fix bug reproduces on the 3-input checksum, and this PR's 26-input checksum
rolls the LiteLLM pod for every provider credential — including the `bedrock_model_ids` textarea
and `google_vertex_credentials` file special cases, and the core-3 regression set.

<sub>Drafted by an AI agent (OpenHands) on behalf of the tester.</sub>

jlav (approved):
Nit before merging: The AI-comments are excessive and refer to tickets that aren't accessible by consumers. Suggest trimming them out before merging.

aivong-openhands:
Thanks for the review! Addressed the nit in 4c6c32bbb:

- Condensed the AI-authored comment in `replicated/openhands.yaml` (the consumer-facing artifact) from 12 lines down to 6, keeping only the essential invariant.
- Removed the internal ticket references (PLTF-3560) from the checksum guard scripts, since those tickets aren't accessible to consumers.

The guard (`scripts/check_litellm_checksum.py`) and the 31-test suite still pass locally.

_This comment was created by an AI agent (OpenHands) on behalf of the PR author._

PR #1265, opened 2026-09-21
test(e2e): catch member financial listing pagination past its first page

aivong-openhands:
## Why

The budget certification suite could never catch a pagination regression in the member financial listing. Its only reader asked for `limit=100` and read the first page of items, without ever sending `page_id` back or following `next_page_id`. That left a server bug unguarded. When an org had more members than one page, `next_page_id` was built by comparing an offset against a `has_more` boolean, so it always came back empty and every member past the first page silently dropped out of the listing.

This adds a regression test that walks the listing with a two-row page size and follows `next_page_id` to the end. It asserts that all three seeded members are reached, that every page still carrying a cursor is exactly full, and that no member is returned twice, so an oversized page or a repeating cursor fails the test rather than slipping past the seeded-member check. To keep it independent of Keycloak and user provisioning, the test inserts throwaway `user` and `org_member` rows straight into the dedicated budget org and deletes them in a `finally` block, so the org is restored even when an assertion fails. Each seeded row keys its email and llm api key off a per-row UUID so parallel runs cannot collide, and the inserts set the `org_member` NOT NULL columns whose server defaults were dropped by later migrations. A new `getMemberFinancialPage` helper passes `page_id` back and decodes `next_page_id`, and the certification workflow now runs the current budgets spec in place of a deleted one so the new test runs on schedule.

## Validation

- **Lint and unit tests** — `npm run lint` (typecheck, ESLint, Prettier) and `npm run test:unit` pass, 20 of 20.
- **Test discovery** — the new spec lists under the returning-user Chromium project, which runs it as an authenticated `APIRequestContext` walk of the members listing on the scheduled certification workflow.


aivong-openhands (commented):
_Note: this review's conclusion is **APPROVE**, but GitHub blocks approving your own pull request, so it is posted as a COMMENT._

Taste: good.

Scope is test-only plus a CI workflow fix, and the change reads cleanly as one flow.

**What I verified on the current head (`a071c47`):**
- `budget-e2e.yml` now runs `tests/009-budgets.spec.ts`. The old target `tests/007-budgets.spec.ts` no longer exists in the tree, so the scheduled workflow was pointing at a deleted file before this PR — this is a correct and worthwhile fix, not just a cosmetic rename.
- The new pagination test is robust against ordering: with the returning org's existing members plus 3 seeded members and a page size of 2, a broken `next_page_id` (the enterprise#410 offset-vs-`has_more` bug) necessarily drops at least one member, so `seen.has(seededId)` fails; working pagination reaches all of them. The `pagesFetched <= 10` guard prevents an infinite walk, and cleanup runs in a `finally` so the dedicated org is restored before `afterAll`.
- `seedMemberFinancialListingMembers` inserts the `org_member` NOT NULL columns (`agent_settings_diff`, `conversation_settings_diff`, `has_custom_llm_a- `seedMemberFinancialListingMembers` inserts the `org_member` NOT NULrations 108/161. Since `withClient` opens a plain connection with no surrounding transaction, the manual delete-on-partial-failure path is the right way to avoid orphaned seed rows.
- `getMemberFinancial` is refactored to reuse the new `getMemberFinancialPage` helper — backward compatible.
- CI is green on this head, including **E2E Static Checks** (lint, `tsc --noEmit`, unit tests, and Chromium test discovery). That also validates the bare `crypto.randomUUID()` global under `@types/node@22` / Node 20.

Non-blocking (no action needed): `current_page` and `per_page` were added to the `MemberFinancialPage` interface but aren't used by the test; harmless.

No material findings — approving.

**[RISK ASSESSMENT]**
LOW. This touches only E2E test utilities and a CI workflow file; nothing ships in the Helm charts or self-hosted defaults. The actual server-side behavior is exercised by the scheduled `budget-e2e.yml` run against staging, not by PR CI, which is expected for this suite.

Key architectural insight: the test asserts reachability across a real `next_page_id` walk rather than a fixed member count, so it stays correct regardless of the org's baseline membership and pins the exact coKey architectural insight: the test asserts reachability across a real `next_page_id` walk rather than a fixer, using the repository's `/codereview` skill._

aivong-openhands:
On the latest commit (`1a67e84`) I checked the change against two testing skills and want to record how each one was applied.

**webapp-testing (Anthropic): assessed, not applied.** That skill drives a locally served web app with Python Playwright scripts, using reconnaissance-then-action, screenshots, and `networkidle` waits. This test is a TypeScript Playwright regression in CI that exercises the members financial endpoint over authenticated HTTP (Playwright `APIRequestContext`) plus direct SQL seeding. There is no local server, no browser page, and no DOM, so nothing from that skill transfers. Its implied move, opening a browser and driving the UI, would only duplicate the contract check at a slower layer, so I did not add it.

**front-end-testing (citypaul): applied.** Its governing rule is "one claim, one harness: prove a claim with the lightest harness that can fail when the claim is false." The claim here is that the endpoint paginates past the first page via `next_page_id`, which is an HTTP contract, so the direct authenticated walk is the right harness and a browser journey was intentionally left out. From the skill's "catch real bugs, test the contract" guidance I hardened the assertions in this commit:

- Assert that no `user_id` repeats across the walk. The old `Set`-only accumulation hid a repeating cursor, because every seeded member still showed up.
- Assert that every page still carrying a cursor is exactly `per_page`. This catches an oversized page that still hands back `next_page_id`, which is the mirror image of the original bug.
- Key each seeded row's email and llm api key off a per-row UUID instead of a shared millisecond timestamp, following the skill's idempotency rule that timestamps alone can collide in parallel runs.

I did not add an explicit "cursor is null on the last page" assertion. The loop already exits only on a null cursor and is bounded by the non-empty-page guard and the page-count cap, so an extra assertion would be a tautology.

Evidence boundary, per the skill's "state where the evidence stops" rule: this proves the HTTP pagination contract as the returning user. It does not prove that the budget members screen renders or paginates that list, because no browser initiator touches that screen. That gap is deliberate, not an oversight.

This comment was drafted by an AI agent on behalf of the user.

ak684 (approved):
Reviewed the cursor walk, seeded-member assertions, and workflow filename repair. Local lint/typecheck and all 20 unit tests pass; the new pagination case is discovered for `chromium:returning`. No blocking findings for the pagination coverage.

Nonblocking fixture hardening: if the `org_member` insert fails after its `user` insert, the current user ID has not yet been added to `users`, so the catch block leaves that user row behind. `withClient` does not open a transaction. Track the ID immediately after the user insert or wrap seeding in a transaction so partial setup is rolled back.

Validation scope: I did not run the destructive suite against a deployed environment. Run certification against a server containing Enterprise #410. Coordinate the shared `getMemberFinancialPage` helper/workflow edits with #1267/#1268 when merging.

PR #1266, opened 2026-09-21
PLTF-3562: page the budgets member listing past its first page

aivong-openhands:
## Why
The budgets member table (`GET /organizations/{id}/members/financial`) is offset-paginated: each response carries a `next_page_id` and a client is meant to follow it until it comes back null. The budget E2E suite only ever read that listing with `limit=100`, so it saw a single page and never followed the cursor — the break shipped in enterprise, where the server derived the cursor from a boolean and returned `next_page_id: null` for every page, was invisible to it. This adds a read-only test that walks the real listing one member per page and asserts the traversal reaches exactly the members the single-page view reports; it runs under the existing returning-admin budget contract and shares that context, so it changes no product behaviour and touches no other test.

The walk only reaches a second page when the org has at least two members; with one it records a `pagination-not-exercised` annotation rather than a pass, and the README asks the dedicated test org to carry a second inert member — no LiteLLM key, no spend — so the guard is active without disturbing the maintenance and unmapped-spend assertions.

## Validation
- **Endpoint contract** — Read the enterprise route and confirmed the listing returns `next_page_id` and that the suite's only reader (`getMemberFinancial`) requested `limit=100`, so the single-page path could not observe a null cursor.


aivong-openhands:
Closing as a duplicate of #1265, which covers the same budgets E2E pagination regression more completely (it seeds members directly into the test org for deterministic multi-page coverage and wires the certification workflow). This PR was also mistakenly branched off an unrelated feature branch, so its diff carried unrelated PLTF-3560 commits. Consolidating on #1265.

This comment was created by an AI agent (OpenHands) on behalf of the user.

PR #1267, opened 2026-09-21
test(e2e): catch a member spend missing from LiteLLM being reported as zero

aivong-openhands:
## Description

Adds an E2E regression test to the budget certification suite (`tests/009-budgets.spec.ts`) that catches the PLTF-3562 defect fixed in `OpenHands/enterprise` PR #412.

### The bug

`GET /orgs/{org_id}/members/financial` reported `lifetime_spend: 0` / `current_budget: 0` for any member that had no entry in the LiteLLM spend read — whether the read failed outright or simply omitted that member (a just-invited member, a key not yet provisioned, mapping drift, an outage). Those fabricated zeros were presented as observed fact, indistinguishable from a genuine zero. The fix reports those fields as `null` and adds a per-page `spend_status` describing the read.

### Why the existing suite couldn't catch it

The suite's only reader (`BudgetApi.getMemberFinancial`) asked for `limit=100`, found the member, and compared `lifetime_spend` numerically. Every read happened against a provisioned, present member, so the endpoint always returned a number. No test ever exercised the member-absent state.

### What this adds

The suite already removes the returning admin from the LiteLLM team mid-run to verify missing-membership repair (`removeLiteLLMTeamMember` → `memberAfterRemoval` is `null`). At that exact moment the member is still an OpenHands org member but has no entry in a *successful* LiteLLM read — precisely the state the old code mis-reported as `0`. The new test captures a member-financial page there and asserts:

- `spend_status` is `'live'` (the read succeeded),
- `lifetime_spend` is `null` (not `0`),
- `current_budget` is `null` (not `0`).

On the pre-fix server the assertion fails with `lifetime_spend === 0`, so the test genuinely catches the regression. It runs inside the existing serial `beforeAll`-seeded certification flow with no new services or config, and the evidence is attached for the report.

### Supporting type widening

The endpoint contract widened `lifetime_spend` / `current_budget` from `number` to `number | null`, so the E2E readers must follow:

- `e2e_tests/utils/budgets.ts` — `MemberFinancial.lifetime_spend` / `current_budget` are now `number | null`; `MemberFinancialPage` surfaces `spend_status`; `BudgetApi.getMemberFinancialPage` is added so callers can distinguish the read status from a row value.
- `e2e_tests/tests/009-budgets.spec.ts` — all existing numeric consumers guard the null case with explicit throws (they run against a provisioned, live member, so the null path is a hard setup failure).
- `e2e_tests/tests/008-managed-key-ownership.spec.ts` — types widened; the nullish coalescing keeps its original numeric-comparison behavior for provisioned members.

The certification workflow and README also now point at `tests/009-budgets.spec.ts` (the 007 file was renamed to 009 in #1263 and the workflow was still referencing the deleted path).

## Validation

- `npm run lint` (typecheck + ESLint + Prettier) passes.
- `npm run test:unit` — budget unit tests pass (8/8); the 6 failures are pre-existing OAuth/two-factor tests that require a local Playwright browser and are unrelated to this change.
- New test is discovered under `chromium:returning` and runs as an authenticated `APIRequestContext` read in the scheduled certification workflow.

The full destructive certification suite runs only against a dedicated test org with `BUDGET_E2E_MUTATION_CONFIRMED=true`, so it is exercised by the existing `budget-e2e` workflow on schedule.

## Additional Notes

This PR was created by an AI agent (OpenHands) on behalf of the requesting user, in support of `OpenHands/enterprise` PR #412.

@aivong-openhands can click here to [continue refining the PR](https://app.all-hands.dev/conversations/71663664-4182-4857-a69c-da6ec4d340e5)


ak684 (approved):
Reviewed the successful-read/missing-member precondition, null assertions before maintenance repair, nullable helper types, and numeric-consumer guards. Local lint/typecheck and all 20 unit tests pass; the missing-spend case is discovered for `chromium:returning`. No blocking findings.

Validation scope: I did not run the destructive suite against a deployed environment. Run certification against an image containing Enterprise #412. When combining with #1265, preserve a single nullable `getMemberFinancialPage` helper and both regressions.

PR #1269, opened 2026-09-22
feat: configure Keycloak admin login through Replicated

ak684:
Customers currently need Kubernetes access to retrieve the generated Keycloak admin password. Add a visible **Keycloak Administration → Keycloak Admin Password** setting in Replicated with the admin console URL and fixed username `admin`. Saving and deploying applies the password to the existing account; leaving the field blank preserves the current password.

The existing hidden bootstrap credential remains unchanged. Before the first password change, provisioning creates an `openhands-provisioner` service account in the master realm and verifies its admin access. Subsequent deployments and token refreshes use that account, so changing the human login does not break realm provisioning. The client has an explicit admin role scope and disables browser/password flows. The console password is supplied only to the provisioning init container, and changes trigger rollout through the existing secrets checksum.

Tests cover fresh and existing installations, preservation of the admin user ID, repeated rotations, special characters, password-history idempotency, blank settings, interrupted client setup, and rejecting a conflicting client without changing credentials. Docker integration tests are enabled in the existing scripts CI job. Documentation covers setup, recovery, and rollback: reverting to older charts after a password change requires restoring the original human admin password first.

Validation:
- Real Keycloak 26.3.3 integration tests plus targeted Python suites: 62 passed.
- OpenHands and secrets Helm unit tests: 179 passed.
- Helm render/lint and `make lint` (packaged Replicated manifests): passed; Replicated emitted troubleshooting-manifest warnings.
- Ruff and `git diff --check`: passed.

Live validation on existing R03 (test customer alonacorp3), release 1646 / chart 0.69.0: upgraded through the Replicated browser UI, set the first password, rotated it through a second config deployment, and reached Ready at KOTS sequences 2 and 3. Keycloak browser/API login succeeded with both configured passwords; after rotation the old password returned 401. The original admin ID and application user count were preserved. No customer installation was modified.

Full OpenHands smoke verification remains incomplete: the prescribed app verification helper exited 1 at a browser Privacy error after clicking GitLab login. Stopped per the test skill without bypassing the login issue. A successful application login and conversation still need verification.


aivong-openhands (approved):
## Code review — `feat: configure Keycloak admin login through Replicated`

Reviewed against the `OpenHands-Cloud` code-review guidelines (Helm chart focus). Overall this is a solid, well-tested change. **Approving.**

### What's good
- **Secret is scoped correctly.** `KEYCLOAK_ADMIN_UI_PASSWORD` is injected *only* into the `keycloak-config` init container via an `optional: true` `secretKeyRef`, and `test_console_password_is_only_passed_to_provisioning_container` explicitly asserts it never reaches the main app containers. Minimal blast radius for a sensitive value.
- **Backward compatible.** `optional: true` on the secret key means existing installs that don't have `admin-ui-password` still render and run unchanged; the whole console-password path is gated behind a non-empty `KEYCLOAK_ADMIN_UI_PASSWORD`.
- **Provisioning identity is stable.** The `openhands-provisioner` service account (explicit `admin` realm role + matching scope-mapping, `fullScopeAllowed: false`, browser/password flows disabled) decouples realm provisioning from the human admin login, so rotating the console password won't break provisioning. The idempotent "try to log in first, only reset if that fails" step also avoids `passwordHistory` breakage.
- **Real, thorough tests.** Integration tests run against actual Keycloak 26.3.3 and cover fresh install, rotation, special characters, password-history idempotency, blank settings, interrupted setup, client collision, and admin identity (`id`) preservation. `secretsChecksum`, Replicated config/secrets, and docs are all wired consistently.

### Minor suggestions (non-blocking)
1. **Document the two sibling secret keys.** `charts/openhands-secrets/values.yaml` adds `keycloak_admin_ui_password: ""` right next to `keycloak_admin_password: ""`. The names are nearly identical but their roles differ sharply (hidden bootstrap credential vs. the human console login), and a direct-Helm consumer can't tell them apart from the names alone. A short comment on each key would satisfy the "document non-obvious values" guideline.
2. **Credential reuse.** The `openhands-provisioner` confidential client's secret is set to `KEYCLOAK_ADMIN_PASSWORD` (the bootstrap credential). This is documented and is what enables clean rollback, but it couples two credentials — a one-line inline comment near client creation would help a future reader know the coupling is intentional.
3. **Pre-existing, out of scope:** the `secretsChecksum` `printf` format string has 45 `%s` but 47 `ConfigOption` args (the +2 delta predates this PR; POSIX `printf` cycles the format so all values still contribute deterministically). Your change is balanced (+1 placeholder / +1 arg), so nothing to fix here — just flagging in case the pre-existing delta is worth a separate cleanup.

None of these affect correctness or security.

---
_This review was generated by an AI agent (OpenHands) on behalf of the requester, following the repository's `.agents/skills/code-review.md` guidelines._

PR #1273, opened 2026-09-22
test(e2e): update page objects and specs for Agent Canvas home UI

lilagrc:
## Why

The staging Home page is now the **Agent Canvas** SPA (mounted at `/canvas`), not the enterprise home. Every UI spec in [ReportPortal launch 142](https://reportportal.beta.staging.all-hands-testing.dev/) failed on the same shape: `HomePage.waitForHomeScreen()` timing out on `launch-new-conversation-button`, which no longer exists — the launcher is now the `home-chat-launcher` composer, and account controls no longer hang off a hover-dropdown avatar.

I confirmed by scraping `data-testid` occurrences out of both the Canvas bundle at `https://app.beta.staging.all-hands-testing.dev/canvas/` and the enterprise root bundle at `/`:

- The Canvas landing exposes `home-screen`, `home-chat-launcher`, `chat-input`, `submit-button`, `conversation-panel`, `conversation-panel-new-thread-picker`, `conversation-card`, `open-repository-button` / `open-repository-dialog-body`, `home-git-control-bar-preview`.
- `user-avatar`, `user-context-menu`, `launch-new-conversation-button`, `recent-conversations`, `repo-launch-button` are all gone from the Canvas landing.
- The **enterprise Settings shell** (`/settings`, `/settings/api-keys`, `/settings/billing`, `/settings/budgets`, `/manage-org`) still exists at basename `/` with `settings-screen`, `settings-nav-user-menu`, `user-balance`, `create-api-key-modal`, etc. Logout was moved into that Settings drawer.

### Page objects

- **`pages/HomePage.ts`** — full rewrite around Canvas locators.
  - `waitForHomeScreen()` asserts on `home-screen` + `home-chat-launcher` (does not require `submit-button` to be enabled — it's disabled until an LLM profile is configured).
  - `startNewConversation(prompt)` — Canvas' launcher creates the conversation **and** sends the first user message atomically, so this now takes the prompt directly (required — no default). Drives the contentEditable composer with `chatInput.click()` + `chatInput.pressSequentially(prompt)` so real per-key `keydown`/`keyup`/`input` events reach React's `onInput`. This deliberately avoids the `page.evaluate` + hand-fired `InputEvent` shim the SDK's `setChatInput` uses: that shim keeps tests green even when the composer's real key handling / Enter binding / focus breaks, which is exactly the class of regression these specs exist to catch.
  - `selectRepository(url)` — drives the new `open-repository-button` → `open-repository-dialog-body` flow, then asserts the `home-git-control-bar-preview` pill appears.
  - `openUserMenu()` / `logout()` — navigate to `/settings` (where the enterprise Settings shell renders), then click the Logout button. Legacy `userAvatar` / `accountSettingsMenu` are kept as deprecated aliases so out-of-tree callers still compile.
  - `clickFirstConversation()` — clicks the first `conversation-card` in the Canvas sidebar.

- **`pages/ConversationPage.ts`** — trimmed to what the running specs actually use.
  - `stopButton` = `stop-button` — icon-only button in Canvas with no accessible name, so testId is the only stable handle (comment on the locator cites `chat-stop-button.tsx`).
  - `statusIndicator` and the ready/complete waits use `.or()` to accept both the old "Waiting for task" / "Agent has finished the task" text and Canvas' new `chat-status-indicator` / `conversation-status-active` / `conversation-status-working` / `conversation-status-check` testIds — so the helpers work through the migration.
  - Removed the `executePrompt` / `sendMessage` / `waitForAgentReady` / `waitForAgentComplete` / `waitForAgentProcessing` / `isAgentProcessing` / `isChatInputEnabled` / `verifyNoErrors` chain. Nothing in the migrated specs calls it — the atomic `HomePage.startNewConversation` replaced it — and keeping the dead chain around under a load-bearing docstring gave a false sense of coverage. If a future spec needs to send a follow-up message inside an already-open conversation we'll re-add `sendMessage` then, validated against a live Canvas run.

### Specs

- **`tests/001-home.spec.ts`** — asserts on Canvas landing (`home-chat-launcher`) and on `/settings` opening with a visible Logout button, instead of on `user-avatar` + `user-context-menu`.
- **`tests/002-billing.spec.ts`** — replace `openUserMenu()` + billing/api-keys link-clicks with direct navigation to `/settings/billing` and `/settings/api-keys`. Those URLs still exist in the enterprise Settings shell; going through the (Canvas) home just to reach them isn't meaningful anymore.
- **`tests/003-conversations-legacy.spec.ts`** — retarget the "reverse a string" and "Tavily search" tests to Canvas' atomic launcher (pass the prompt to `startNewConversation()`; drop the separate `executePrompt()` on the conversation page). The "VSCode integration" test is skipped with a TODO — the `conversation-tab-vscode` / `file-diff-viewer-outer` / `changes-refresh-button` surface has no Canvas equivalent yet.
- **`tests/004-api-keys.spec.ts`** — direct navigation to `/settings/api-keys`.

## Not addressed here (deliberately)

- `setup/setup-new-user.ts` failing on the GitHub OAuth "authorize" step never redirecting back. Per direction from the requester, that is out of scope for this PR — it will show up as a real (non-UI-migration) failure once the Canvas UI specs go green.
- `tests/006-org-management.spec.ts`'s `user-context-menu org-selector` assertion. That test is `new-user`-only and is gated by the currently-broken new-user setup, so it won't run until OAuth is fixed. The `org-selector` testId isn't in the deployed enterprise root shell; when the new-user flow comes back we'll need to figure out the Canvas equivalent (likely the workspace picker) and reshape that assertion.

## Verification

Run locally in `e2e_tests/`:

- `npm run typecheck` → clean.
- `npm run lint` → clean (typecheck + eslint + prettier).
- `npm run test:unit` → 20 passed (all unit tests, including the OAuth-authorization and 2FA-reminder helpers; unrelated to the UI changes).

Full E2E requires a live GitHub OAuth session against staging, so I can't run it from this container. Marking as **draft** so a reviewer with staging credentials can kick a ReportPortal run before merging.

---

_This PR was created by an AI agent (OpenHands) on behalf of a user._

@lilagrc can click here to [continue refining the PR](https://app.beta.staging.all-hands-testing.dev/conversations/5638b5a4-1a0e-47e2-b865-6f1af96ad92a)


e2e_tests/pages/HomePage.ts
```diff
@@ -83,119 +116,168 @@ export class HomePage extends BasePage {
...
+   * @param prompt - First user message. Falls back to a benign default.
    */
-  async startNewConversation(
-    buttonId: string = "launch-new-conversation-button",
-  ): Promise<void> {
-    const startButton = this.page.getByTestId(buttonId);
-    if (await startButton.isVisible({ timeout: 5_000 }).catch(() => false)) {
-      await startButton.click();
-    }
+  async startNewConversation(prompt: string = "hello"): Promise<void> {
```
aivong-openhands: The docstring above (lines 177-181) says the legacy positional `buttonId` is retained for source-compatibility and "used only as a hint" — but the implementation just types the argument verbatim into `chat-input` and submits it. There is no branch that treats a test-id string differently from a real prompt, so a legacy caller like `startNewConversation("repo-launch-button")` would send the literal string `repo-launch-button` to the agent rather than acting as a hint.

Recommend making the docstring match the code (drop the source-compat claim; the param is simply the first message) or making the code match the docstring (detect known test-id strings such as `launch-new-conversation-button` / `repo-launch-button` and fall back to the default prompt).

aivong-openhands (commented):
## Code review — Agent Canvas e2e page-object/spec migration

Reviewed against the repo's `/codereview` guidelines. This is a test-only PR (Playwright page objects + specs); none of the Helm-chart-specific rules apply. Overall the migration is well-reasoned and unusually well-documented — the transitional `.or()` matchers on `ConversationPage` status locators and the deprecated `userAvatar`/`accountSettingsMenu` aliases are sensible ways to keep the harness green across both shells.

One design issue is worth attention before this lands (see inline comments):

**`startNewConversation` silently repurposed its positional argument.** The parameter changed from `buttonId` semantics to `prompt` semantics while keeping the same positional slot, but the implementation now types whatever string it receives straight into the chat composer. The docstring claims the legacy `buttonId` is "used only as a hint" for source-compatibility — the code does no such thing. The skipped VSCode test still calls `startNewConversation("repo-launch-button")`, which would send the literal string `repo-launch-button` as the agent's first message if that test were ever un-skipped. That's a latent footgun for the un-skip and for any out-of-tree caller the docstring is trying to reassure.

Suggestion: either drop the source-compat claim from the docstring and rename/retype the param cleanly, or make the code actually honor the documented contract (detect a known test-id string and substitute a default prompt). As-is the docstring and behavior disagree.

Everything else looks good and follows existing patterns. Not blocking — flagging as a COMMENT for the author's call.

e2e_tests/pages/ConversationPage.ts
```diff
@@ -57,17 +57,18 @@ export class ConversationPage extends BasePage {
...
-      .locator(
-        'button[type="submit"], button:has-text("Send"), [data-testid*="send"]',
-      )
-      .first();
-    this.stopButton = page
-      .locator('button:has-text("Stop"), [data-testid*="stop"]')
-      .first();
+    // Canvas ships stable data-testids for both actions; prefer them over the
+    // old fuzzy button-text selectors so a UI copy tweak doesn't break tests.
+    this.sendButton = page.getByTestId("submit-button");
```
aivong-openhands, line 67: **Query priority (medium):** both skills put `getByTestId` *last* and `getByRole(name)` *first*. Container elements without a natural role/name (`home-screen`, `conversation-panel`, `conversation-card`) legitimately need testids, but the send/stop **buttons** almost certainly have accessible names, so:

```ts
this.sendButton = page.getByRole("button", { name: /send|submit/i });
this.stopButton = page.getByRole("button", { name: /stop/i });
```

This mirrors how a screen-reader user finds the control and makes the test fail if the button loses its accessible name — a regression a `data-testid` lookup silently tolerates. (Credit: `logout()` already does this correctly.)

e2e_tests/pages/ConversationPage.ts
```diff
@@ -219,22 +232,39 @@ export class ConversationPage extends BasePage {
...
 
-    // Clear any existing content and type the message
-    await this.chatInput.click();
-    await this.chatInput.fill("");
-    await this.page.keyboard.type(message);
-
-    // Submit the message
-    await this.page.keyboard.press("Enter");
+    await this.page.evaluate((text) => {
+      const el = document.querySelector('[data-testid="chat-input"]');
```
aivong-openhands: **Core anti-pattern (high):** setting `textContent` and hand-firing a single `InputEvent` inside `page.evaluate` is exactly what both skills flag — front-end-testing/`playwright-e2e.md`: *"`dispatchEvent` or DOM evaluation for ordinary user actions when a real accessible locator action exists"*; webapp-testing: drive the UI with real locator actions.

This is the load-bearing action of a "user sends a message" test, so it **stays green even if the composer's real key handling, Enter binding, onChange wiring, or focus management breaks** — the exact false-positive these specs exist to catch. It also re-`querySelector`s the element inside `evaluate`, skipping Playwright's actionability checks.

Prefer a real browser gesture that still drives React's `onInput`:

```ts
const input = this.page.getByTestId("chat-input");
await input.click();
await input.pressSequentially(message); // per-key browser events
```

If a controlled contentEditable genuinely still needs the `dispatchEvent` shim, isolate it and add a JSDoc note stating the boundary — "this seeds text; it does not prove a human could type into the composer" — so the limitation is visible rather than implied away.

e2e_tests/pages/HomePage.ts
```diff
@@ -2,51 +2,86 @@ import { Page, Locator, expect } from "@playwright/test";
...
 
     this.homeScreen = page.getByTestId("home-screen");
-    this.newConversationSection = page.getByTestId(
-      "home-screen-new-conversation-section",
-    );
-    this.recentConversationsSection = page.getByTestId(
-      "home-screen-recent-conversations-section",
+    this.chatLauncher = page.getByTestId("home-chat-launcher");
+    this.chatInput = page.getByTestId("chat-input");
+    this.submitButton = page.getByTestId("submit-button");
```
aivong-openhands, line 67: **Query priority (medium):** same as the ConversationPage send/stop buttons — `submit-button` likely has an accessible name, so `page.getByRole("button", { name: /send|submit/i })` is preferred over `getByTestId` per both skills' query-priority ordering, and it catches accessibility regressions the testid lookup ignores.

e2e_tests/pages/HomePage.ts
```diff
@@ -83,119 +116,168 @@ export class HomePage extends BasePage {
...
+    const fallback = dialog.getByText(repoName, { exact: false }).first();
+    const target = (await option.count()) > 0 ? option : fallback;
+    await expect(target).toBeVisible({ timeout: 10_000 });
+    await target.click();
+
+    // Confirm the selection so it pins into the launcher's git-control bar.
+    const confirm = dialog.getByRole("button", {
+      name: /^(open|select|confirm)$/i,
+    });
+    if (await confirm.isVisible({ timeout: 2_000 }).catch(() => false)) {
```
aivong-openhands: **Determinism (medium):** clicking the confirm button *only if it happens to be visible* makes the journey nondeterministic. If this button should always render as part of the repo-select flow, gating it behind `isVisible({ timeout: 2_000 }).catch(() => false)` will silently skip a regression where it stops appearing — the test still "passes." Prefer resolving the actual dialog flow and making the step unconditional. The final `home-git-control-bar-preview` assertion is a good observable outcome to keep as proof the selection stuck.

e2e_tests/pages/HomePage.ts
aivong-openhands: **Misleading doc + hazardous default (high):** the signature changed from `buttonId: string` to `prompt: string`, but the JSDoc still says the first positional arg is the legacy `buttonId` "retained for source-compatibility." That is no longer true — the first arg is now **sent as the chat prompt**. An un-migrated caller doing `startNewConversation("launch-new-conversation-button")` would silently submit that string as the user's first message and still pass. The `= "hello"` default compounds it: a caller who forgets the prompt gets a silent benign message instead of a failure.

Recommend: delete the "buttonId retained" paragraph (factually wrong and dangerous) and make `prompt` **required** so a missing prompt is a compile-time error.

e2e_tests/pages/HomePage.ts
```diff
@@ -83,119 +116,168 @@ export class HomePage extends BasePage {
...
+    await expect(this.chatLauncher).toBeVisible({ timeout: 10_000 });
+    await expect(this.chatInput).toBeVisible({ timeout: 10_000 });
 
-    // Wait for conversation/chat interface to load
+    // The chat input is a contentEditable div, not a real input. `.fill()`
+    // is unreliable against contentEditables, so set text programmatically
+    // and dispatch an InputEvent — same pattern the SDK's own e2e helpers
+    // use (see `setChatInput` in OpenHands/OpenHands mock-llm helpers).
+    await this.page.evaluate((text) => {
+      const el = document.querySelector('[data-testid="chat-input"]');
```
aivong-openhands: **Core anti-pattern (high):** same DOM-evaluation typing as `ConversationPage.sendMessage` — see that comment. Since this is the home launcher's primary "user starts a conversation" gesture, replacing it with `getByTestId("chat-input").click()` + `pressSequentially(prompt)` matters even more here: it's the entry point for most of the suite.

e2e_tests/pages/HomePage.ts
```diff
@@ -83,119 +116,168 @@ export class HomePage extends BasePage {
...
+    }, prompt);
+
+    await expect(this.submitButton).toBeEnabled({ timeout: 15_000 });
+    await this.submitButton.click();
+
+    // The launcher navigates to `/conversations/{id}` once the create-conversation
+    // mutation resolves. Match on the SDK path shape.
     await this.page
-      .waitForURL(/conversation|chat|app/, { timeout: 30_000 })
+      .waitForURL(/\/conversations\/[^/]+/, { timeout: 60_000 })
```
aivong-openhands: **Swallowed navigation failure (high):** `waitForURL(...).catch(() => {})` makes this method return "successfully" even when the launcher never navigates; the failure then surfaces later in `waitForConversationReady()` with a confusing, misattributed error. `playwright-e2e.md` flags exactly this — lifecycle races hidden instead of asserted. Let it throw so the failure points at the real cause:

```ts
await this.page.waitForURL(/\/conversations\/[^/]+/, { timeout: 60_000 });
```

e2e_tests/pages/HomePage.ts
```diff
@@ -83,119 +116,168 @@ export class HomePage extends BasePage {
...
-   * Note: The menu is conditionally rendered based on async state (config loaded,
-   * user authenticated, etc.). We need to wait for the menu element to be attached
-   * to the DOM before we can interact with it. The menu appears on hover over the
-   * user-actions container, or when clicking the avatar toggles state.
+   * The Canvas home has no user avatar or hover-triggered account dropdown —
+   * account controls (org selector, API keys, billing, logout) live in the
+   * enterprise Settings shell at `/settings`. Navigating there is the
+   * closest equivalent to the old "open user menu" gesture.
    */
   async openUserMenu(): Promise<void> {
```
aivong-openhands: **Naming (low):** `openUserMenu()` now performs a full `page.goto('/settings')` navigation rather than opening a menu, so the name is a misnomer. Consider renaming to `openAccountSettings()` (keeping the old name as a thin deprecated alias if out-of-tree callers need it). Same note for the `userAvatar`/`accountSettingsMenu` aliases now pointing at unrelated Settings locators — pragmatic, but a mild readability trap worth a comment.

aivong-openhands (commented):
## E2E review against two testing skills

Reviewed against Anthropic's [`webapp-testing`](https://github.com/anthropics/skills/blob/main/skills/webapp-testing/SKILL.md) skill and citypaul's [`front-end-testing`](https://github.com/citypaul/.dotfiles/blob/main/claude/.claude/skills/front-end-testing/SKILL.md) skill (plus its `playwright-e2e.md` evidence-boundary resource).

**Overall:** a solid, well-documented migration. Removing the `waitForTimeout(2000)` sleeps and the `networkidle` wait, and tightening the logout matcher to `getByRole("button", { name: /^logout$/i })`, all move in the direction both skills recommend. The PR body is an exemplary "hand-back": it names what was verified (typecheck/lint/unit) and what wasn't (full E2E needs live OAuth), which is exactly the *evidence-boundary* discipline the front-end-testing skill asks for. Keeping it draft until a staging ReportPortal run is the right call.

The inline comments below flag one core anti-pattern both skills call out (DOM-evaluation typing instead of a real browser gesture), two correctness issues, and a few query-priority / determinism improvements.

### High priority
1. **Typing via `page.evaluate` + `dispatchEvent`** (ConversationPage `sendMessage`, HomePage `startNewConversation`) bypasses the real user gesture — the load-bearing action of a "user sends a message" test. Prefer `locator.click()` + `pressSequentially()`; if the shim is truly required, isolate it and state the evidence boundary in the JSDoc.
2. **Misleading JSDoc + hazardous default** on `startNewConversation(prompt = "hello")` — the doc claims the first positional arg is still `buttonId`, but it is now sent as the chat prompt.
3. **Swallowed `waitForURL(...).catch(() => {})`** hides navigation failure and misattributes the later error.

### Medium / low
4. Prefer `getByRole` over `getByTestId` for buttons that have accessible names (submit/stop).
5. Conditional confirm-click in `selectRepository` can silently skip a regression.
6. `openUserMenu()` is now a full navigation, not a menu open — naming smell.

Details inline. _This review was posted by an AI agent (OpenHands) on behalf of @aivong._

lilagrc:
## Round 1 review — responses

Pushed [5351671](https://github.com/OpenHands/OpenHands-Cloud/pull/1273/commits/5351671) addressing the review. Summary per point:

### High priority — all addressed

**#1 Typing via `page.evaluate` + `dispatchEvent`** — Fixed on both `ConversationPage.sendMessage` and `HomePage.startNewConversation`. Both now do `chatInput.click()` + `chatInput.pressSequentially(text)` so per-key `keydown`/`keyup`/`beforeinput`/`input` events actually reach the composer. The JSDoc on `sendMessage` calls out explicitly why the SDK's `setChatInput` shim is *not* the right pattern for these specs (it masks broken key handling / Enter binding / onChange wiring / focus — exactly the regression the specs exist to catch).

**#2 Misleading JSDoc + hazardous default on `startNewConversation`** — Fixed. Dropped the "buttonId retained for source-compatibility" paragraph (which was factually wrong — the code never treated the arg as a hint) and dropped the `= "hello"` default so `prompt` is now required. The compiler now catches the case flagged in-review: the skipped VSCode test was calling `startNewConversation("repo-launch-button")`, which would have submitted the literal string `repo-launch-button` to the agent. Replaced with the actual README-edit prompt so an un-skip does the right thing.

**#3 Swallowed `waitForURL(...).catch(() => {})`** — Fixed. Removed the `.catch`; if the launcher fails to navigate the timeout now throws at the launcher call site instead of surfacing later as a confusing "conversation never became ready" inside `waitForConversationReady`.

### Medium

**#4 Prefer `getByRole` over `getByTestId` for submit/stop buttons** — **Not applied — pushing back with evidence.** I traced both components in `OpenHands/OpenHands`:
- [`src/components/features/chat/chat-send-button.tsx`](https://github.com/OpenHands/OpenHands/blob/main/src/components/features/chat/chat-send-button.tsx) renders `<button data-testid="submit-button">` with only an `<ArrowUp />` icon inside — no `aria-label`, no `title`, no text child.
- [`src/components/features/chat/chat-stop-button.tsx`](https://github.com/OpenHands/OpenHands/blob/main/src/components/features/chat/chat-stop-button.tsx) is the same shape with a `<PauseIcon />`.

`page.getByRole("button", { name: /send|submit|stop/i })` returns zero matches against those. Switching would break the tests without catching any regression — and the "regression the testid lookup silently tolerates" is *already* live: those buttons genuinely have no accessible name. That's an app-side fix (add `aria-label` to `ChatSendButton` / `ChatStopButton`) rather than a test-side one. I added a code comment on both locators documenting the finding so the next person doesn't spend cycles retrying; happy to file the frontend follow-up as a separate issue.

**#5 Conditional confirm-click in `selectRepository`** — Fixed, and while I was in there I rewrote the whole helper against the real dialog shape. Previous version was speculative (`cloud-repo-search-input` isn't a real testId). Traced [`open-repository-dialog.tsx`](https://github.com/OpenHands/OpenHands/blob/main/src/components/features/home/open-repository-dialog.tsx) → [`repo-selection-form.tsx`](https://github.com/OpenHands/OpenHands/blob/main/src/components/features/home/repo-selection-form.tsx) → [`git-repo-dropdown.tsx`](https://github.com/OpenHands/OpenHands/blob/main/src/components/features/home/git-repo-dropdown/git-repo-dropdown.tsx). The real flow is `git-repo-dropdown` (autocomplete input) → `git-repo-dropdown-menu` (option list) → `repo-launch-button` (always-rendered Confirm; disabled until repo+branch are chosen). New code asserts `toBeEnabled` on the confirm — a missing button now fails as "button never enabled" instead of as a silent skip. Since this helper is only reachable from the currently-skipped VSCode test I can't validate it end-to-end without OAuth, but at least it's now grounded in the actual component tree.

### Low

**#6 `openUserMenu` naming smell** — Fixed. Renamed to `openAccountSettings` (matches what the method actually does — `page.goto("/settings")` navigation, not a menu open). `openUserMenu` kept as a thin `@deprecated` alias with a docstring flagging the misnomer, so 002/004/006 still compile without a churn commit.

### Verification
- `npm run typecheck && npm run lint`: clean
- `npm run test:unit`: 20/20 pass
- Full E2E still needs live OAuth against staging — same evidence boundary as the original PR; still draft.

_This response was generated by an AI agent (OpenHands) on behalf of the user._

aivong-openhands (commented):
## Code review

**Decision: COMMENT** (no blocking / "important"-level issues)

This is a test-only change migrating the Playwright page objects and specs from the old enterprise home to the Agent Canvas SPA. The repo's code-review guidelines are Helm-chart-focused, so the chart-specific checks don't apply here — I reviewed it on the general framework (default-to-approve, comment on questions/design) plus test quality and doc accuracy.

Overall this is careful, well-documented work: the locator rewrites, the `.or()` transition-tolerant selectors, and the reasoning for driving the contentEditable composer with `pressSequentially` rather than a hand-fired `InputEvent` shim are all sound. The items below are questions/nitpicks, none blocking.

### Worth addressing

1. **PR description contradicts the code (contentEditable driving).** The "What" section says `startNewConversation` "uses the contentEditable-safe pattern (set `textContent`, dispatch `InputEvent`…)". The code and its in-line comments do the opposite — they use `pressSequentially` and *explicitly reject* the `evaluate`+`InputEvent` shim to avoid false positives. Worth updating the description to match.

2. **`sendMessage` / `executePrompt` are now effectively dead code.** After this PR no spec calls `executePrompt`, and `sendMessage` is only called by `executePrompt`. Yet `sendMessage`'s new doc comment calls it "the load-bearing 'user sends a message' action of every conversation spec" — no longer true now that the atomic launcher replaced it. Consider trimming the claim or removing the unused helpers.

3. **`repo-launch-button` testId reuse.** The description says `repo-launch-button` is "gone from the Canvas landing," but `selectRepository()` reuses that exact testId as the `OpenRepositoryDialog` confirm button. Can you confirm it's genuinely the dialog's confirm control (a different element sharing the id)? This path can't be E2E-verified from the PR container, so it's an assumption worth calling out.

### Minor / FYI

4. **`006-org-management` semantics shifted silently.** `openUserMenu()` now navigates to `/settings` and `accountSettingsMenu` now aliases the whole `settings-screen`, so `accountSettingsMenu.getByTestId("org-selector")` in 006 means something different than before. The PR already notes 006 is gated by the broken new-user OAuth setup so it won't run — just flagging it will need reshaping when re-enabled.

5. **`selectRepository()` is unverified.** Its only caller is the VSCode test, which is `test.skip(true, …)`, so the new dialog-driven repo flow isn't exercised by any running spec.

---
_This review was created by an AI agent (OpenHands) on behalf of a user._

aivong-openhands:
@lilagrc can you validate the latest code review?

lilagrc:
## Round 2 review — responses

Pushed [670d7e1](https://github.com/OpenHands/OpenHands-Cloud/pull/1273/commits/670d7e1) and rewrote the PR body to match the code.

### Worth addressing

**#1 PR description contradicts the code (contentEditable driving).** Fixed. The "What → HomePage" and "What → ConversationPage" sections now describe the `click()` + `pressSequentially()` approach that's actually in the code, and explicitly note we *reject* the `evaluate` + `InputEvent` shim (with reasoning). The stale "set `textContent`, dispatch `InputEvent`" claim is gone.

**#2 `sendMessage` / `executePrompt` are effectively dead code.** Fixed by removing rather than trimming — I chose "no unverified helpers claiming coverage" over "keep the primitives around." The whole dead chain is gone from `ConversationPage`:

- `executePrompt` (uncalled)
- `sendMessage` (only called by `executePrompt`)
- `verifyNoErrors`, `waitForAgentReady`, `waitForAgentComplete`, `waitForAgentProcessing`, `isAgentProcessing`, `isChatInputEnabled` (transitively only reachable through `executePrompt`)
- `sendButton` locator (only used by the deleted `sendMessage`)

Remaining vocabulary is `HomePage.startNewConversation` (atomic composer) → `ConversationPage.waitForConversationReady` / `waitForTaskCompleteMessage`, which is what the running specs actually call. If a future spec needs a follow-up message inside an already-open conversation we'll re-add `sendMessage` at that point, validated against a live Canvas run rather than leaving unverified code in tree carrying a "load-bearing" docstring.

**#3 `repo-launch-button` testId reuse.** Confirmed — annotated at the call site. The same testId names two different elements: the old enterprise landing's inline "Launch" button (gone from Canvas) and the `<BrandButton>` inside `RepositorySelectionForm`, which the modal still renders. See [`repo-selection-form.tsx` L202](https://github.com/OpenHands/OpenHands/blob/main/src/components/features/home/repo-selection-form.tsx#L202) for the second occurrence. The lookup was already scoped to the `dialog` locator so there's no ambiguity at runtime; the new comment cites the source line so the next reader doesn't have to trace it. Full E2E still can't be validated from this container (needs live OAuth); grounded in the actual component tree is the best I can do here.

### Minor / FYI

**#4 `006-org-management` semantics shifted silently.** Acknowledged — deferring per the requester's direction. The 006 test is gated by the broken new-user OAuth setup and won't run until that lands; the `openUserMenu` → `openAccountSettings` alias keeps it compiling in the meantime. The `accountSettingsMenu.getByTestId("org-selector")` assertion will need reshaping (likely to the Canvas workspace picker) in a follow-up PR after new-user OAuth is fixed.

**#5 `selectRepository()` is unverified.** Also acknowledged. Its only caller is the currently-skipped VSCode integration test, so it isn't exercised by any running spec in this PR. The rewrite is grounded in the real component tree (`OpenRepositoryDialog` → `RepositorySelectionForm` → `GitRepoDropdown` → `repo-launch-button` confirm), but I can't E2E-validate it without live OAuth. Un-skipping the VSCode test — which also needs a Canvas equivalent for `conversation-tab-vscode` / `file-diff-viewer-outer` / `changes-refresh-button` — is the natural point to prove it.

### Verification
- `npm run lint` (typecheck + eslint + prettier): clean
- `npm run test:unit`: 20/20 pass
- Full E2E still requires live OAuth against staging; still draft.

_This response was generated by an AI agent (OpenHands) on behalf of the user._

aivong-openhands (approved):

PR #1281, opened 2026-09-23
test: replace new-user GitHub OAuth with synthetic Keycloak-native login

lilagrc:
## Why

The new-user setup drives a real GitHub OAuth round-trip against a shared test account. That flow keeps breaking on GitHub-side product changes we don't own — 2FA reminder-page variants, verified-device screens, first-time consent form races — and is currently failing on ReportPortal because GitHub removed the **"skip 2FA verification"** button entirely. The `dismissTwoFactorReminder` `waitFor` times out, the flow parks on the reminder page, and the setup throws `OAuth authorization did not complete`. That fails every downstream `chromium:new-user`/`firefox:new-user`/`webkit:new-user` spec in each run.

The value of the new-user role is exercising the **app's** first-login flow — TOS acceptance, the onboarding form, org bootstrap, first landing on the Canvas home — not GitHub's OAuth screens. There's no reason for those specs to depend on GitHub's UI.

## What

Replace the new-user role's authentication with a synthetic, Keycloak-native account:

- **`keycloak-cleanup`** deletes any user matching `KEYCLOAK_NEW_USER_EMAIL` (as before).
- **`setup-new-user`** now creates the user via the Keycloak Admin API with a known password, then drives Keycloak's local Sign-In form. It bypasses the "Log in with GitHub" social button by intercepting the `/openid-connect/auth` request and stripping `kc_idp_hint` before it reaches the server (via `page.route` + `route.fallback`), so Keycloak renders its own form instead of 302-ing to github.com.
- `completeLoginAndOnboard` is unchanged — the app-side onboarding path is exactly what this role exercises.

The **returning-user** role keeps its real GitHub-federated login, so the app ↔ Keycloak ↔ GitHub round-trip is still covered on every run.

## Env var changes (breaking — Argo Secret refs need updating in the same window)

| Old | New |
|---|---|
| `NEW_GITHUB_USERNAME` | `KEYCLOAK_NEW_USER_USERNAME` |
| `NEW_GITHUB_PASSWORD` | `KEYCLOAK_NEW_USER_PASSWORD` |
| `NEW_GITHUB_TOTP_SECRET` | *removed* (no 2FA on the synthetic account) |
| `KEYCLOAK_NEW_USER_EMAIL` | *unchanged* |

## Safety guards (defense-in-depth for a script that deletes realm users)

- `KEYCLOAK_NEW_USER_EMAIL` must match `^[^@]+@([^@]+\.)?e2e\.test$`. `.test` is [RFC 2606](https://www.rfc-editor.org/rfc/rfc2606) reserved and cannot resolve to a real mailbox — any address matching the pattern is provably outside any customer or employee namespace.
- Both delete-by-email and create-user assert on the allowlist before hitting the Keycloak Admin API.
- Both setup projects call `assertNotProduction()` and refuse to run when `BASE_URL` resolves to the production host.

## Tests

Adds `unit-tests/keycloak-native-login.spec.ts` which fakes the app login page and Keycloak's authorize endpoint via `page.route`, then asserts:

- `kc_idp_hint` is stripped before it reaches Keycloak (the fixture serves a "regression" page with no `#username` input when the hint is present, so any regression is caught by the subsequent `waitFor` timing out).
- `#username` / `#password` are filled and `#kc-login` is submitted (the fake `login-actions/authenticate` handler captures the POST body).

All 21 unit tests pass; `npm run lint` (typecheck + eslint + prettier) is clean; `npx playwright test --list` resolves 159 tests when both roles are enabled and correctly filters out `keycloak-cleanup` / `*:new-user` / `setup:new-user` projects when only `RETURNING_GITHUB_USERNAME` is set (and vice versa).

## Deploy notes

Before merging, add `KEYCLOAK_NEW_USER_USERNAME` and `KEYCLOAK_NEW_USER_PASSWORD` to the e2e Secret in the relevant environments and remove the three `NEW_GITHUB_*` values. `KEYCLOAK_NEW_USER_EMAIL` needs to be set to something under `*.e2e.test` (e.g. `e2e-new-user@e2e.test`) — the guard will refuse any other value.

Scoped separately from #1273 (Canvas UI selectors) so the two can move independently.

---

_This PR was created by an AI agent (OpenHands) on behalf of the requester._

@lilagrc can click here to [continue refining the PR](https://app.beta.staging.all-hands-testing.dev/conversations/5638b5a4-1a0e-47e2-b865-6f1af96ad92a)


lilagrc:
Thanks for the review!

**Stale `playwright.config.ts` comments** — fixed in acd9233. Both the `keycloak-cleanup` and `setup:new-user` project comments now reference `KEYCLOAK_NEW_USER_USERNAME` and describe the new-user setup as Keycloak-native login (no GitHub). Also fixed `setup:returning`'s comment to spell out that it's federated through Keycloak's GitHub IdP, since that's the distinction that matters now. Confirmed `grep -rn NEW_GITHUB e2e_tests/` returns clean.

**PR title** — retitled to `test:` (only touches `e2e_tests/`, so `test:` is the accurate scope).

**Replicated deploy gate** — agreed, unrelated to this PR; `main` branch issue.

_Reply posted by an AI agent (OpenHands) on behalf of the requester._

